PR #13, opened 2021-10-27
Install Pundit as authorization system.

stevepolitodesign:
This will help us standardize our approach to authorization within the application. We'll need to make sure we're rendering a flash message within the application, otherwise users won't see a warning when the're redirected from an endpoint they do not have access to.

I deliberately did not [enable policies and scopes everywhere](https://github.com/varvet/pundit#ensuring-policies-and-scopes-are-used), but we might want to consider this if the app requires a user to be subscribed to have access.

**Other Notes**

- Pundit can help with [scopes](https://github.com/varvet/pundit#scopes)
- Pundit can help with [strong params](https://github.com/varvet/pundit#strong-parameters)

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -20,3 +20,5 @@ group :development do
 end
 
 gem "tzinfo-data", platforms: [:mingw, :mswin, :x64_mingw, :jruby]
+
+gem "pundit", "~> 2.1"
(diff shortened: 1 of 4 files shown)


joemasilotti:
Closed as these changes were included in #25.

Issue #18, opened 2021-10-30
Extract common styled elements (form inputs, buttons, error messages, etc.)


joemasilotti:

joemasilotti:
Form inputs/buttons and the Message button are the top two candidates here. @jacobdaddario, this could be a great opportunity to learn about View Components, especially for the Message button.

lsparlin:
Are you open to new contributors?  I like working with view components and have a been hoping to contribute to railsdevs since you announced the idea.

joemasilotti:
Definitely! Please do, all PRs are welcome 😀

lsparlin:
How do you envision component-izing error messages?

joemasilotti:
`app/views/shared/_flash.html.erb` is probably a good candidate, it should DRY up that repeated code for formatting errors vs. messages.

Also, if there's a way to incorporate the `yield` on the last line that would be great. Right now it's being used in `app/views/shared/_error_messages.html.erb`.

joemasilotti:
Diving in a bit more, there are three concepts:

1. Color (blue vs. red) and icon
2. Title (present or not)
3. Content (paragraph vs. list)

A single View Component might be able to handle all of these (with tests!) quite nicely.

joemasilotti:
I would also love to see avatar and cover photos get the same treatment. Currently, there are a lot of `if` statements floating around the views to handle developer's without an attached upload.

lsparlin:
👍 I aim to work on some more `ViewComponent`s for over the next few evenings

joemasilotti:
Closed via #28.

PR #19, opened 2021-10-30
feat: Show message UI element based on auth

### LINKED ISSUE #8: Only show "Message" button to signed in users

jacobdaddario:
Resolves #8

For now this just adds the UI functionality for authorizing users before showing the message button. In the future obviously we will need to authorize the actual controller actions that go into that feature.

Additionally for now, this is tested using a controller test. Given #18, we might want to install the `view_component` gem. It'll help us extract out common UI patterns, and it'll make it faster to test UI elements like this in the fast. I don't have too much experience with the gem, but I hear that's the main benefit.

### DIFF
--- a/app/views/developers/show.html.erb
+++ b/app/views/developers/show.html.erb
@@ -15,13 +15,14 @@
             <%= @developer.hero %>
           </h2>
         </div>
-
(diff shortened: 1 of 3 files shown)


jacobdaddario:
This is ready to merge unless you want to include the implementation for messages.

Issue #10, opened 2021-10-26
Developers can only edit their own profile


joemasilotti:

joemasilotti:
FYI, #7 is done with basic users and authentication.

Tonksthebear:
Awesome! I'll knock that out ASAP

PR #25, opened 2021-11-01
Authorize Developers can only edit their own profiles

Tonksthebear:
Jumping off from #13, this is intended to solve #10


Tonksthebear:
Let me know if I wrote the test the way you like! I also noticed Pundit generates its own test files, so I wasn't sure if you wanted to run authorization tests through the Pundit tests, or integrations tests. I played it safe and put the added test in the integration test. I'm also not well versed in best practices for tests, so any and all criticism is definitely welcomed

PR #29, opened 2021-11-03
Add pre-commit actions, auto Standardize

lsparlin:
Hey @joemasilotti after your revision to my last PR I had any idea 💡Feel free to decline this PR if you don't want the automatic enforcement.

## Changes
- Add `lefthook` gem in development for pre-commit hooks
- Run `standardrb` command on pre-commit
- Will require developer to resolve `standard` discrepancies before they can commit

image


jacobdaddario:
Wow, I didn't realize you could do this kind of thing. That's super cool and would be really helpful. I forget to run `standardrb` all the time.

joemasilotti:
Oh, this is awesome! But I think I'm leaning more towards running standard as a CI step, versus on the user's local machine.

What do you think about integrating this GitHub Action? https://github.com/marketplace/actions/standardrb-action

lsparlin:
Oh yeah I think that looks like a good solution also. You'd run that on pull request creation?

joemasilotti:
I'd like it to run on every push to a PR.

joemasilotti:
You know... I think I like having a commit hook. Let's do both! One request, though. Can we migrate to `overcommit` instead? The configuration file looks a bit easier to read IMO.

joemasilotti:
Unless I'm mistaken, I don't think those are `standardrb` checks. I just tried editing my Gemfile and committing and all the checks passed but it should have broken on the quotes.

```diff
diff --git a/Gemfile b/Gemfile
index f0ef5bc..09b6564 100644
--- a/Gemfile
+++ b/Gemfile
@@ -3,7 +3,7 @@ git_source(:github) { |repo| "https://github.com/#{repo}.git" }

 ruby "3.0.2"

-gem "rails", github: "rails/rails", branch: "main"
+gem 'rails', github: "rails/rails", branch: "main"

 gem "cssbundling-rails"
 gem "hotwire-rails", "~> 0.1.3"
```

lsparlin:
Crud. I'll go re-read overcommit docs and figure it out

lsparlin:
Okay I see, I actually think `lefthook` was easier to configure. `overcommit` has built-in plugins you can enable, one called `Standard`, but it isn't the testdouble one. In order to add a custom one it requires a ruby class to defined the behavior.  `lefthook` was just a `yml` file with a command

https://github.com/sds/overcommit#repo-specific-hooks

joemasilotti:
Ah, got it. Thanks for digging in. I'm happy to go with whatever is easier to read/configure then as long as it works. Sorry for the waffling!

lsparlin:
@joemasilotti no worries, I learned about a new gem from the experience.  I've changed it back to `lefthook` with the yaml only config 😄

joemasilotti:
Awesome, thanks again for this! Do you want to tackle the CI step, too? If not, I can create a new issue.

lsparlin:
Yes, I'd be happy to work the CI step too

PR #31, opened 2021-11-04
Add standardize_ruby task to github actions workflow

lsparlin:


lsparlin:
@joemasilotti I think you'll have to add a permission to this repository's `GITHUB_TOKEN` to make this Action work https://docs.github.com/en/actions/security-guides/automatic-token-authentication#permissions-for-the-github_token

joemasilotti:
Thanks for this!

OK, I looked around the docs but do not understand where/how to add anything to `GITHUB_TOKEN`. Does anyone have experience with that and could lend a hand or direct me to a tutorial/video?

lsparlin:
Is this the correct setting? (I'm not really familiar either)
image

joemasilotti:
That's what I thought, but I already have that checked.

davegudge:
@joemasilotti I'm pretty sure the error:

```
/action/lib/github_client.rb:36:in `request_http': Forbidden: {"message":"Resource not accessible by integration","documentation_url":"https://docs.github.com/rest/reference/checks#create-a-check-run"} (RuntimeError)
```

is because the check is running from a fork i.e. `lsparlin:lsparlin/standard-gh-action` using the standard [`on :pull_request`](https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#pull_request) event. 

Access is restricted to _your_ repository when checks are running in a PR outside of _your_ repository, this is to protect _your_ repository i.e. actions running on the fork only have readonly access.

You could get around this by using the [pull_request_target](https://docs.github.com/en/actions/learn-github-actions/events-that-trigger-workflows#pull_request_target) event. **However**, this does come with risks, so is probably not advisable. 

Screenshot 2021-11-06 at 20 47 04

Hopefully this makes sense, but if not I can explain further.

joemasilotti:
I ended up cherry-picking your commit directly into `main` and everything seems to be working! Thanks for the contribution :)

PR #32, opened 2021-11-05
Add ToastMessageComponent with tests

lsparlin:
`ToastMessageComponent` uses the "Slots" feature of View Components to offer flexible display options

## Screenshots
image

image

image

### DIFF
--- a/app/components/toast_message_component.html.erb
+++ b/app/components/toast_message_component.html.erb
@@ -1,15 +1,15 @@
-<div class="sm:rounded-md shadow-sm p-4 mb-8 bg-<%=- @color %>-50">
+<div class="sm:rounded-md shadow-sm p-4 mb-8 <%= background_classes %>">
   <div class="flex">
     <div class="flex-shrink-0">
-      <%= inline_svg_tag icon_path(icon: @icon), class: "h-5 w-5 text-#{@color}-400" %>
+    <%= inline_svg_tag icon_path(icon: @icon), class: "h-5 w-5 #{icon_classes}" %>
     </div>
 
     <div class="ml-3">
       <% if title.present? %>
-        <h3 class="text-sm font-medium text-<%=- @color %>-800"><%= title %></h3>
+        <h3 class="text-sm font-medium <%= title_classes %>"><%= title %></h3>
       <% end %>
 
-      <div class=" text-sm flex-1 md:flex md:justify-between text-<%=- @color %>-700">
+      <div class=" text-sm flex-1 md:flex md:justify-between">
         <%= content_tag(@messages_tag, role: "list", class: "list-disc space-y-1 #{messages_classes}") do %>
           <% messages.each do |message| %><%= message %><% end %>
         <% end %>
--- a/app/components/toast_message_component.html.erb
+++ b/app/components/toast_message_component.html.erb
@@ -1,0 +1,19 @@
+<div class="sm:rounded-md shadow-sm p-4 mb-8 bg-<%=- @color %>-50">
+  <div class="flex">
+    <div class="flex-shrink-0">
+      <%= inline_svg_tag icon_path(icon: @icon), class: "h-5 w-5 text-#{@color}-400" %>
+    </div>
+
+    <div class="ml-3">
+      <% if title.present? %>
+        <h3 class="text-sm font-medium text-<%=- @color %>-800"><%= title %></h3>
+      <% end %>
+
+      <div class=" text-sm flex-1 md:flex md:justify-between text-<%=- @color %>-700">
+        <%= content_tag(@messages_tag, role: "list", class: "list-disc space-y-1 #{messages_classes}") do %>
+          <% messages.each do |message| %><%= message %><% end %>
+        <% end %>
+      </div>
+    </div>
+  </div>
+</div>
--- a/app/components/toast_message_component.html.erb
+++ b/app/components/toast_message_component.html.erb
@@ -1,7 +1,7 @@
 <div class="sm:rounded-md shadow-sm p-4 mb-8 <%= background_classes %>">
   <div class="flex">
     <div class="flex-shrink-0">
-    <%= inline_svg_tag icon_path(icon: @icon), class: "h-5 w-5 #{icon_classes}" %>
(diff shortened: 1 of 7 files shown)


thubamamba:
LGTM. 👍🏽

Maybe we can also think of use cases where a form has multiple form fields.

lsparlin:
Sadly ViewComponent has some known issues with rails form_with https://viewcomponent.org/known_issues.html

jacobdaddario:
> Sadly ViewComponent has some known issues with rails form_with https://viewcomponent.org/known_issues.html

Yeah, I recently looked into this for a personal project. Since Rails' form builder is essentially a component, it creates some problems. The design of `FormBuilder` being a wrapper around `*_tag` methods creates a lot of weird code interactions in general 😵

app/components/toast_message_component.html.erb
```diff
@@ -0,0 +1,19 @@
+<div class="sm:rounded-md shadow-sm p-4 mb-8 bg-<%=- @color %>-50">
+  <div class="flex">
+    <div class="flex-shrink-0">
+      <%= inline_svg_tag icon_path(icon: @icon), class: "h-5 w-5 text-#{@color}-400" %>
```
Tonksthebear, line 4: Just a heads up, the tailwind purger won't correctly purge classes that are spliced together like this. It's annoying, but you'd have to clearly declare text-red-400, text-green-400, etc. for the purger to pick up on it, rather than concatenating strings together

Tonksthebear: I noticed there was at least one other spot where the tailwind class was concatenated like this, so just thought I'd mention it!

lsparlin: Oh thanks for pointing this out. I'm not super familiar with tailwind yet. Do you have any suggestions?

joemasilotti: Great catch, @Tonksthebear.

@lsparlin, instead of passing in only the `color`, we should pass in the entire class name. For example, `text-red-400` or `text-blue-400`. Then we can make sure that the Tailwind config is checking `app/components/` for class names to purge.

Tonksthebear: I'm not sure if it's proper, but I normally just define a list of constant class colors in the ruby file, such as:

`STYLE = {
  blue: 'text-blue-400',
  green: 'text-green-400'
}`

Then I just pass in the style I want to the component like you already have set up. Just instead of setting @color the parameter, I'd do
`@color = STYLES[color]`

Then swap `text-#{@color}-400` with `@color`

Hopefully that makes sense! I wrote this up pretty quickly... there might be a better way, just wanted to get some ideas out there in case they're useful.

lsparlin: There are several such classes in the component. You think there should be a parameter for each of them?

joemasilotti (changes requested):
This looks great outside of the Tailwind classes comment from @Tonksthebear. Once that is resolved we can merge this in.

app/components/toast_message_component.html.erb
joemasilotti: Yeah, this could get ugly quickly, @lsparlin. I wonder what other apps are doing in this space.

Tonksthebear: I definitely understand how it becomes unwieldy. I just couldn't find another way because Tailwind requires such a verbose syntax. To accommodate multiple parameters, I just make the hash larger and define getters. So some quick sudo code for what I do

```
STYLES = {
  blue: {
    header: 'text-blue-200',
    message: 'text-blue-400',
    footer: 'text-blue-100'
  }
...
def header_classes
  STYLES[@color][:header]
end

def message_classes
  STYLES[@color][:message]
end

def footer_classes
  STYLES[@color][:message]
end
```

Then I would just call the corresponding getters from the html.

Again, I completely get how verbose this solution may sound but this is what worked best for my mental model when working with Tailwind

lsparlin: So as long as the class (full literal string) shows up anywhere in `html` or `rb` files tailwind will be happy?

Thank you @Tonksthebear

Tonksthebear: Not to overload this thread with more information, but another avenue I've pursued is creating a helper that pulls from a config file for the corresponding classes.

For instance, I'd have a config/styles.yml file that contains something like

```
toast:
  blue:
    header: 'text-blue-200 ...'
    message: 'text-blue-400 ...'
    footer: 'text-blue-100 ...'
```

Then in the html I'd just call `tailwind(toast: { @color => :header })`

I've enjoyed this pattern due to no boilerplate in the .rb files, but it does feel dangerously close to just simulating css classes. I suppose another avenue would be using ViewComponent's sidecar assets in some way (if possible). 

In either case, it's a problem I've been personally wrestling with so I'd love to hear other solutions

Tonksthebear: > So as long as the class (full literal string) shows up anywhere in `html` or `rb` files tailwind will be happy?
> 
> Thank you @Tonksthebear

Yup! Just have to update the tailwind.config.js to point it to the correct files

joemasilotti: My vote is for a big hash and helper methods for now. But I do love a YAML-powered configuration for the different types of colors! Maybe if we need them in the future that will be a good route to take.

lsparlin:
Finally got around to this. Hope I wasn't holding anything up. This should work better with Tailwind's purging

joemasilotti:
Amazing work, team! This code is looking awesome. I pulled in `main` and made two style changes.

Before I merge, can you fix the display issue of the component? It somehow lost its bottom margin, or something.

image

lsparlin:
@joemasilotti Weird, it still has the `mb-8` class

joemasilotti:
I know... I'm stumped too. But I don't think this was an issue before the latest round of commits.

lsparlin:
Got it. Needed to make tailwind are of `components/**/*.html.erb`

joemasilotti:
Amazing! Will merge when CI is green. Thanks again for this!

PR #34, opened 2021-11-05
feat: Validate that user doesn't already have dev

### LINKED ISSUE #33: Users who already have a developer profile can create a new one
Users who already have a developer profile can create a new one despite already having a profile. This can create a confusing UX when the following happens:

1. A user mistakenly clicks "Create a profile" thinking they're logged in.
2. Realizing they are not logged in, the user logs in.
3. The log in controller attempts to re-direct the user to their previous location.
4. The user is now on the screen to create a new profile even though they already have one.

jacobdaddario:
Resolves #33 

Also, I noticed that `DeveloperController` doesn't use the Rails association methods to create records. I'm not sure if you care one way or another about using them, but I figured I'd mention it just in case.


joemasilotti:
This looks great! Two change requests:

1. If I'm signed out and click "Add my profile" but I already have a profile... let's redirect to editing my profile instead of the "You are already signed in" message
2. Please update `developerprofile` to `developer profile`

And two questions:

1. Do we need to add `policy_class: DeveloperPolicy` in the controller? It seems to work without it, but maybe I'm missing something.
2. What do you mean by "the Rails association methods?"

joemasilotti (changes requested):
This looks great! Two change requests:

1. If I'm signed out and click "Add my profile" but I already have a profile... let's redirect to editing my profile instead of the "You are already signed in" message
2. Please update `developerprofile` to `developer profile`

And two questions:

1. Do we need to add `policy_class: DeveloperPolicy` in the controller? It seems to work without it, but maybe I'm missing something.
2. What do you mean by "the Rails association methods?"

jacobdaddario:
Hey! I just got off a long day of a work and have a date night with the girlfriend, so I may not get to the revision items. That said, I can answer the questions super fast.

> Do we need to add policy_class: DeveloperPolicy in the controller? It seems to work without it, but maybe I'm missing something.

Yeah, that's the case. The policy is essentially checking if the user already has a developer associated. That means that the object sent to `authorize` can be `Nil`. Since `authorize` infers the policy when it's not explicitly given, when it is given `Nil`, it looks for `NilPolicy` or something like that. Therefore, we have to pass the policy name, since we intend to occasionally pass in `Nil`. There might be a smarter way to do this, but I couldn't think of anything.

> What do you mean by "the Rails association methods?"

The way I learned Rails was through Michael Hartl's _The Ruby on Rails Tutorial_. In it he talks a bit about the semantic way to build associations. This involves using the methods added to a model from `has_many`, `has_one`, and `belongs_to`. As a side note, that's what those methods do; they use metaprogramming to add methods to the object they're on

Particularly Michael talks about using something like `@user.build_developer(developer_params)` to build associations rather than something like is done in the app, `Developer.new(developer_params.merge(user: @user))`. The first uses one of the association methods added by `has_one` which is considered idiomatic Rails according to him.

Now, with all of that said, I have never written an ounce of professional Rails. I don't know how often people really adhere to these kinds of things, and I absolutely wouldn't want to make code style changes on another person's repo. I just wanted to mention it so that you were aware.

joemasilotti:
First off, there's zero pressure to get to this tonight! So no worries there; enjoy the weekend :)

Ah, that makes sense regarding passing in the policy name. Thanks for the clarification! That was super helpful.

Finally, building the association record via the method you listed sounds great. I like it better than what we are currently doing. Feel free to make those changes to the controller as well.

joemasilotti:
I took this over the finish line – thanks for the PR @jacobdaddario! I decided to punt on the one outstanding request and created a [new issue](#50) for it.

jacobdaddario:
Thank you! I'm in College Station today for the football game, and I'm super jealous of all the Rs I see coming through 😆

PR #41, opened 2021-11-05
Paginate /developers

A7madXatab:
- installed faker for seeding test data.
- installed and configured pagy.
- adds a new `pagination_controller` that will help paginate the elements.
- stop at the last page to prevent the client from sending extra requests by using 
https://github.com/A7madXatab/railsdevs.io/blob/1157dcb1f797cfdab66b3a21f191f13527e82a18/app/views/developers/index.turbo_stream.erb#L4
- closes https://github.com/joemasilotti/railsdevs.io/issues/38.
- for the pagination controller i took inspiration from one i built(and use in my work). Check it out [here](https://dev.to/rockwell/infinite-scroll-with-hotwire-part-1-configuration-4n32)

https://user-images.githubusercontent.com/38360603/140589509-a7ff365a-15b9-4ac6-8290-169c4aaa5ed0.mov


app/javascript/controllers/pagination_controller.js
```diff
@@ -0,0 +1,69 @@
...
+export default class extends Controller {
+    static targets = ['lastPage', 'loadMoreContainer', 'skeleton']
+
+    static values = {
+        url: String,
+        page: Number,
+    };
+
+    initialize() {
+        this.bodyScrollListener = this.bodyScrollListener.bind(this);
```
joemasilotti, line 13: I strongly prefer using `IntersectionObserver` over a scroll listener. @excide3 has a great [GoRails example with some sample code](https://github.com/gorails-screencasts/infinite-scroll-stimulus-js/blob/master/app/javascript/controllers/infinite_scroll_controller.js) showing how to do this.

Also, we should be able to use the next page `rel` links that `pagy` autogenerates (and we hide) to tell if there are more pages. We shouldn't need a custom HTML tag.

app/views/developers/index.turbo_stream.erb
```diff
@@ -0,0 +1,5 @@
+<%= turbo_stream.append "developers" do %>
+  <%= render @developers %>
+
+  <%= last_page_target if @pagy.last == @pagy.page %>
```
joemasilotti, line 4: > Also, we should be able to use the next page rel links that pagy autogenerates (and we hide) to tell if there are more pages. We shouldn't need a custom HTML tag.

joemasilotti (changes requested):
Thanks for this PR! The demo video looks awesome.

I requested a fair amount of changes, but they boil down to two things.

1. Code formatting. Along with all the nitpicky requests I made, please run `bundle exec standardrb --fix` to fix all Ruby code style issues.
2. I think the overall approach to the JavaScript could be simplified a bit – there's more context in the comments.

Tonksthebear:
I'm not sure if it is desirable or necessary to be this "lean", but it is actually decently straightforward to leverage Turbo for 0 custom code infinite scroll. I can give examples if that's a route you all want to take, but I understand just sticking to a stimulus controller since it's already written

joemasilotti:
My gut reaction is the less JavaScript we have to manage the better.

Tonksthebear:
@A7madXatab  if you want any suggestions, let me know. I'd be happy to help. I will have a decently busy weekend though, so I'll try to respond when I can if you want any help

app/javascript/controllers/pagination_controller.js
A7madXatab: Alright, so i've ended up using the controller provided by Chris. The first time the pagy links are intersected it works fine, but after that it does not work. Any ideas why?


https://user-images.githubusercontent.com/38360603/140611916-be5a75c8-5ab1-4c42-b9c4-828c0b76e993.mov

joemasilotti: Can you set a `debugger` breakpoint or log out in the `loadMore()` method to make sure: 

1. It get's called
1. `next_page` isn't `null`
2. `next_page.href` isn't `null`
3. Add an error callback and see if the XHR is failing

A7madXatab: I've found what was causing the issue. When i was replacing the pagy links 
```erb
<%= turbo_stream.replace "pagination" do %>
  <div id="pagination" data-pagination-target="pagination">
    <%== pagy_nav(@pagy) %>
  </div>
<% end %>
```

the observer was getting disconnected. So, i'm listening to connection event and simply reattaching the observer.

```js
    paginationTargetConnected() {
        this.intersectionObserver.observe(this.paginationTarget)
    }
```

db/seeds.rb
```diff
@@ -42,3 +42,19 @@ def create_user!(name)
   github: "ada",
   twitter: "lovelace"
 )
+
+require 'faker'
+
+100.times do |index|
```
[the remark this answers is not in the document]
A7madXatab: removed faker!.

app/helpers/application_helper.rb
```diff
@@ -1,2 +1,5 @@
 module ApplicationHelper
+  def last_page_target
```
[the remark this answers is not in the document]
A7madXatab: placed it inside `PagyHelper`. see https://github.com/joemasilotti/railsdevs.io/pull/41/commits/eec3d05116938c457ef54ac51472ca4c5c17da54

A7madXatab:
@joemasilotti all the issues you pointed out were resolved. Lemme know if i need to make other changes!. Thanks for the `IntersectionObserver` shoutout. I didn't know how to use it before.

joemasilotti:
Looking good! I left the default pagination visible in case the user doesn't have JavaScript enabled. Then, when the Stimulus controller connects, the pagination is hidden. I'll merge this when CI is green. Thanks for the PR!

A7madXatab:
> Looking good! I left the default pagination visible in case the user doesn't have JavaScript enabled. Then, when the Stimulus controller connects, the pagination is hidden. I'll merge this when CI is green. Thanks for the PR!

Thanks for merging this!. Can't wait to contribute more to this repo!.

Issue #36, opened 2021-11-05
Website links on developer profiles don't link externally


joemasilotti:

maful:
@joemasilotti  I'm happy to take this one

joemasilotti:
Sure thing! Open up a PR and I'll make sure to review yours first for this issue.

SmithWebDev:
PR input to resolve external link error

PR #42, opened 2021-11-06
Updated developer show view with link to external developer website

SmithWebDev:
Wrapped the ActiveRecord call with https request to allow User input website to open as an external link


SmithWebDev:
Closes #36

maful:
This will not work if the data from the DB seed 

https://github.com/joemasilotti/railsdevs.io/blob/ca8db02190a1c807871fa9af76cf8e40c7505da5/db/seeds.rb#L17

I believe we should prevent the developer from adding protocol to the website field and update the DB seed as well. what do you think @joemasilotti ?

joemasilotti:
Good call, @maful. Here's what I propose:

1. Remove the `https://` from the seed data
2. Write a migration (or one-off script) that removes `http://` and `https://` from the beginning of `developers.website`
3. Add a `before_save` hook on `Developer` that drops the scheme from their URL

jacobdaddario:
Could also be worth just adding a few methods to detect and strip the leading `https://` without altering the existing data in the database.

joemasilotti:
Great call, @jacobdaddario!

I created a new component that naively checks if the website starts with `http://` or `https://`, if not, it adds `https://` to the `href` param. Should work fine for now and we don't have to mess with any data or validations moving forward.

Issue #50, opened 2021-11-06
Redirect to profile when signing in via "Add my profile" link


joemasilotti:
Actual flow:

1. Sign out
2. Click "Add my profile"
3. Sign in
4. See flash error "You are already signed in."

Expected flow:

1. Sign out
2. Click "Add my profile"
3. Sign in
4. Get redirected to editing your profile

joemasilotti:
No longer needed after the redesign #73.

PR #52, opened 2021-11-06
redirect to profile when user signs in successfully

A7madXatab:
closes https://github.com/joemasilotti/railsdevs.io/issues/50.

- fixes issue when the logout button is pressed the redirect to the root path will be a `turbo_stream` so the `index.turbo_stream` will be rendered instead of `index.html`.

Prod: 

https://user-images.githubusercontent.com/38360603/140626242-1de6cd90-abfd-432e-8ae1-e8b051c221e7.mov

Dev: via https://github.com/joemasilotti/railsdevs.io/pull/52/commits/65a08196a44807da9912964db21c619a31b21569

https://user-images.githubusercontent.com/38360603/140626248-0928a057-d365-486e-9bf4-4eb55aaf9018.mov

### DIFF
--- a/app/controllers/users/sessions_controller.rb
+++ b/app/controllers/users/sessions_controller.rb
@@ -1,0 +1,5 @@
+class Users::SessionsController < Devise::SessionsController
+  def after_sign_in_path_for(resource)
+    developer_path resource
+  end
+end
--- a/app/controllers/users/sessions_controller.rb
+++ b/app/controllers/users/sessions_controller.rb
@@ -2,4 +2,8 @@ class Users::SessionsController < Devise::SessionsController
   def after_sign_in_path_for(resource)
     developer_path resource
   end
+
+  def after_sign_out_path_for(resource)
+    request.referer || root_path
+  end
 end
--- a/app/controllers/users/sessions_controller.rb
+++ b/app/controllers/users/sessions_controller.rb
@@ -1,6 +1,10 @@
 class Users::SessionsController < Devise::SessionsController
   def after_sign_in_path_for(resource)
-    developer_path resource
+    if resource.developer.present?
+      developer_path resource.developer
+    else
+      root_path
+    end
   end
 
   def after_sign_out_path_for(resource)
--- a/app/controllers/users/sessions_controller.rb
+++ b/app/controllers/users/sessions_controller.rb
@@ -1,10 +1,8 @@
 class Users::SessionsController < Devise::SessionsController
   def after_sign_in_path_for(resource)
-    if resource.developer.present?
-      developer_path resource.developer
-    else
-      root_path
-    end
+    return root_path unless resource.developer
+
+    developer_path resource.developer
   end
 
   def after_sign_out_path_for(resource)
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,5 +1,5 @@
 Rails.application.routes.draw do
-  devise_for :users
+  devise_for :users, controllers: {sessions: "users/sessions"}
 
   resources :developers, except: %i[destroy]
 
--- a/test/fixtures/users.yml
+++ b/test/fixtures/users.yml
@@ -1,6 +1,7 @@
 with_profile_one:
   email: one@example.com
   confirmed_at: <%= DateTime.current %>
+  encrypted_password: <%= Devise::Encryptor.digest(User, 'password') %>
 
 with_profile_two:
   email: two@example.com
@@ -9,6 +10,7 @@ with_profile_two:
 without_profile:
   email: three@example.com
   confirmed_at: <%= DateTime.current %>
+  encrypted_password: <%= Devise::Encryptor.digest(User, 'password') %>
 
 admin:
   email: admin@example.com
--- a/test/integration/users/sessions_test.rb
+++ b/test/integration/users/sessions_test.rb
@@ -2,12 +2,12 @@ require "test_helper"
 
 class Users::SessionsTest < ActionDispatch::IntegrationTest
   test "it redirects to developer profile if profile is present" do
-    post user_session_path user: { email: users(:with_profile_one).email, password: 'password'}
+    post user_session_path user: {email: users(:with_profile_one).email, password: "password"}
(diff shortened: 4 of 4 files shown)


app/controllers/users/sessions_controller.rb
```diff
@@ -0,0 +1,9 @@
+class Users::SessionsController < Devise::SessionsController
+  def after_sign_in_path_for(resource)
+    developer_path resource
```
joemasilotti, line 3: If the user doesn't have a developer profile yet then `resource` is nil and this raises an exception.

joemasilotti (changes requested):
Along with the comment, can you please add tests for the three code paths you added?

1. Signing in with an existing developer profile
2. Signing in to an account without an existing developer profile
3. Signing out

app/controllers/users/sessions_controller.rb
```diff
@@ -0,0 +1,13 @@
+class Users::SessionsController < Devise::SessionsController
+  def after_sign_in_path_for(resource)
+    if resource.developer.present?
+      developer_path resource.developer
+    else
+      root_path
+    end
```
davegudge, line 7: You could do something like this if exit clauses are preferred over if/else's:

```suggestion
    return root_path unless resource.developer

    developer_path resource.developer

Comment (A7madXatab) [reply app/controllers/users/sessions_controller.rb]: This is actually nicer than mine. Committing it. Thanks for the suggestion.

File: app/controllers/users/sessions_controller.rb (hunk 1 above)
Comment (A7madXatab) [reply app/controllers/users/sessions_controller.rb]: Solved by checking if the developer is present

```ruby
  def after_sign_in_path_for(resource)
    if resource.developer.present?
      developer_path resource.developer
    else
      root_path
    end
  end
```

A7madXatab:
@joemasilotti All is done, let me know if there is anymore i need to do.

A7madXatab:
@joemasilotti any news about this?

app/controllers/users/sessions_controller.rb
```diff
@@ -0,0 +1,11 @@
+class Users::SessionsController < Devise::SessionsController
+  def after_sign_in_path_for(resource)
+    return root_path unless resource.developer
```
joemasilotti, line 3: This breaks existing behavior. If you checkout `main`, clicking "Add my profile" then signing in will redirect to editing your profile. Is there a way to call `super` or something and only override the behavior if the resource is a `developer` or something? Or maybe we need to rethink this override?

joemasilotti (changes requested):

app/controllers/users/sessions_controller.rb
davegudge: When a user signs in, if they have a developer [profile], they'll be redirected to the edit view. That's the expectation from #50 ?

The 'resource' relates to the the 'user' record, so this line says: 'return the root path unless the user has a developer profile' e.g. any users who have confirmed their account but not yet created a developer [profile] are redirected to the root path. 

Do you think a better approach would be: upon confirming a user account, users should be redirected to create a developer profile i.e. all requests are intercepted to ensure a developer profile exists, if not the user is redirected to the new developer form?

joemasilotti: Good point! Here's what I think the flows should be:

1. Confirm account -> redirect to creating new developer profile
2. Click "Add your profile" -> sign in with existing with account with developer profile -> redirect to edit your profile
3. Sign in (by manually navigating to `/users/sign_in`) -> redirect to referrer or `root_path`

davegudge: > Confirm account -> redirect to creating new developer profile

What is your expectation if the user does navigates away from the [new developer] form without submitting it?

Would you like to prevent an authenticated user from navigating elsewhere in the application until they have created a profile? If so, overriding the `after_sign_in_path_for` is not the correct place to achieve this, as this method is only catering for a redirect post authentication.

If you do want to enforce developer profile creation (as a kind of next step after confirming the user account), I've implemented this before, so I can always open a PR if that's the preferred approach.

A7madXatab: > Would you like to prevent an authenticated user from navigating elsewhere in the application until they have created a profile? If so, overriding the after_sign_in_path_for is not the correct place to achieve this, as this method is only catering for a redirect post authentication.

I agree with @davegudge my approach does not work for if we want to disallow a user to visit pages until they have a profile created.

If this is desired behavior, we can have a `before_action :developer!` that will check if the user is developer, if not they will be redirect to create profile page?

davegudge: > we can have a before_action :developer!

Yes, that's what I was thinking. 

With regard to addressing #50, I don't think any redirects are required. The unauthenticated user is clicking 'Add my profile', presented with the sign in screen, and upon success authentication, using devise's standard implementation of 'redirecting to the referrer' should suffice? 

I believe the issue outlined in #50 is that the flash message 'You are already signed in.' is being displayed?

A7madXatab: > I believe the issue outlined in #50 is that the flash message 'You are already signed in.' is being displayed?

I believe so, yes. So, if we could remove the message I think this PR will also become obsolete. Then removing the message would be fine, and perhaps even desirable?. 
CC @joemasilotti

davegudge: One solution is to blank out the [`devise.failure.already_authenticated`](https://github.com/joemasilotti/railsdevs.io/blob/main/config/locales/devise.en.yml#L10) i18n entry. 

Another i18n entry I tend to blank out is [`devise.sessions.signed_in`](https://github.com/joemasilotti/railsdevs.io/blob/main/config/locales/devise.en.yml#L49) as it's stating the obvious.

joemasilotti:
Closed via #73. The copy/hrefs on the links are now dynamic in the header.

I had to set `data: {turbo: false}` on the sign out link. I couldn't get the redirection you mentioned to work on my machine (Safari).

If you can reproduce and fix it again with changes to the `SessionsController` please open a new PR.

PR #54, opened 2021-11-07
Make available_on optional

maful:
Issue https://github.com/joemasilotti/railsdevs.io/issues/46#issuecomment-962522548

image

### DIFF
--- a/db/migrate/20211107024057_change_developers_available_on.rb
+++ b/db/migrate/20211107024057_change_developers_available_on.rb
@@ -1,0 +1,5 @@
+class ChangeDevelopersAvailableOn < ActiveRecord::Migration[7.0]
+  def change
+    change_column_null :developers, :available_on, true
+  end
+end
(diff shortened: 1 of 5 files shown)


db/migrate/20211107024057_change_developers_available_on.rb
```diff
@@ -0,0 +1,5 @@
+class ChangeDevelopersAvailableOn < ActiveRecord::Migration[7.0]
+  def change
+    change_column_null :developers, :available_on, true
```
joemasilotti, line 3: Oh! I didn't know this was a thing. TIL.

maful: I just found out too 

API https://edgeapi.rubyonrails.org/classes/ActiveRecord/ConnectionAdapters/SchemaStatements.html#method-i-change_column_null

joemasilotti (approved):
Looks great! Thanks for the PR :)

db/migrate/20211107024057_change_developers_available_on.rb
joemasilotti: I've always done `change_column :developers, :available_on, null: false` but you have to remember to copy over any other params, like `default:`. Yours is so much better!

Issue #46, opened 2021-11-06
Add filter developers by Availability


maful:
This will be easier to find the developer by their availability

Here's the time range I propose:
- All -> this is the default one
- Available Now
- In a week
- In two weeks
- In a month
- More than a month

lsparlin:
Is anyone working in this? I cannot not

joemasilotti:
I'm wondering if we need a filter here. Like, would sorting by availability work even better? I imagine employers would be looking for devs ASAP, but I could be wrong. 

Maybe we could add a sorting option (default to recently added) for next available date. Then, group everyone by how far out they are. For example, Available Now, available in less than 1 month, less than 3 months, etc. 

What do folks think about that?

A7madXatab:
Shouldn't we allow turning availability off?. Like it's not required?. Currently, we don't allow to create an account unless the availability is set. How about we allow nil values and when nil, we count them as unavailable ?

joemasilotti:
I wanted availability to be mandatory, but it seems like folks are kind of abusing the date by selecting something unreleasticly in the future. So I've changed my mind and would be happy to make `available_on` not required.

maful:
@joemasilotti I will create a PR to make `available_on` optional

joemasilotti:
Sounds great, thanks, @maful!

joemasilotti:
The site has changed a bit since this issue was opened. I'm going to close this for now as I focus on Messaging and Payments.

PR #58, opened 2021-11-07
`Availability` concern

davegudge:
Introduce an `Availability` concern to encapsulate the logic to determine if a developer is available by utilising an `enum`.

This PR [originally](88cd5b242fdb991044cd41776c00e0461907130a) DRY'ed up the view logic by utilising a partial, however, the logic in #59 has been cherry-picked and adapted to work with the `Availability` concern.

*N.B:* This PR retains the previous styling / [wording](https://github.com/joemasilotti/railsdevs.io/pull/58/files#diff-44438ce218f5287c58d0017f965d888715635d94280669896f75841fbd7b4cd7R17-R20). i.e there are no changes to the UI from a user perspective.

---
Screenshot 2021-11-07 at 13 59 34

Screenshot 2021-11-07 at 13 59 27

### DIFF
--- a/app/components/availability_component.rb
+++ b/app/components/availability_component.rb
@@ -1,5 +1,3 @@
-# frozen_string_literal: true
-
 class AvailabilityComponent < ViewComponent::Base
   def initialize(developer:)
     @developer = developer
--- a/app/components/availability_component.rb
+++ b/app/components/availability_component.rb
@@ -1,0 +1,17 @@
+# frozen_string_literal: true
+
+class AvailabilityComponent < ViewComponent::Base
+  def initialize(developer:)
+    @developer = developer
+  end
+
+  def date
+    @developer.available_on&.to_s(:db)
+  end
+
+  def date_in_words
+    return unless @developer.available?
+
+    time_ago_in_words(@developer.available_on)
+  end
+end
--- a/app/components/availability_component.rb
+++ b/app/components/availability_component.rb
@@ -10,7 +10,7 @@ class AvailabilityComponent < ViewComponent::Base
   end
 
   def date_in_words
-    return unless @developer.available?
+    return unless @developer.available_in_future?
 
     time_ago_in_words(@developer.available_on)
   end
--- a/app/models/concerns/availability.rb
+++ b/app/models/concerns/availability.rb
@@ -1,0 +1,28 @@
+# frozen_string_literal: true
+
+module Availability
+  extend ActiveSupport::Concern
+
+  included do
+    enum availability_status: [:unspecified, :now, :in_future], _default: :unspecified, _prefix: :available
+  end
+
+  def available_on=(date)
+    super(date)
+    derive_availability_status
+  end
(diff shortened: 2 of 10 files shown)


app/views/developers/_availability.html.erb
```diff
@@ -0,0 +1,5 @@
+<span class="text-gray-900 text-sm <%= "font-medium" if developer.available_now? %>">
```
joemasilotti, line 1: This file can be removed, right?

app/components/availability_component.rb
```diff
@@ -0,0 +1,17 @@
+# frozen_string_literal: true
```
joemasilotti, line 1: Please remove this comment.

app/models/concerns/availability.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
```
joemasilotti, line 1: Please remove this comment.

test/models/developer_test.rb
```diff
@@ -0,0 +1,37 @@
+# frozen_string_literal: true
```
joemasilotti, line 1: Please remove this comment.

joemasilotti (changes requested):
This looks amazing! I love the architecture of the code.

Unfortunately, I don't think the concern is working as expected. `#available_on=` doesn't seem to be getting called when the record is loaded from the database. So if you visit the index view you can see that all developer's are "Currently unavailable" even though the seeds data has 2 available devs.

app/views/developers/_availability.html.erb
davegudge: Good spot! I must have forgotten to delete this file when cherry-picking the component commit. I guess this is why we code review 😉

app/components/availability_component.rb
davegudge: No problem. 

I was going to suggest that this magic comment improves performance (https://www.mikeperham.com/2018/02/28/ruby-optimization-with-one-magic-comment/), but that does not always appear to be the case according to this post: https://accidentaltechnologist.com/ruby-on-rails/frozen_string_literal-the-not-so-magical-comment/

joemasilotti: I seem to remember that it's not needed with Ruby 3.0 either.

app/models/concerns/availability.rb
```diff
@@ -0,0 +1,28 @@
...
+
+module Availability
+  extend ActiveSupport::Concern
+
+  included do
+    enum availability_status: [:unspecified, :now, :in_future], _default: :unspecified, _prefix: :available
```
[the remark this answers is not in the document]
davegudge: This `enum` is not backed by a db column at the moment, but that may change when #46 is actioned.

davegudge:
New commits pushed onto this PR.

joemasilotti (approved):
Looks great, thanks for the fixes!

Issue #37, opened 2021-11-05
Open Graph meta tags


joemasilotti:
* title and `twitter:title`
* description and `twitter:description`
* image and `twitter:image` – hoping to integrate [Mugshot Bot](https://mugshotbot.com) for dynamic images

pingortle:
@joemasilotti I'm on it! Should be able to get to it in the next couple days. Sound good?

joemasilotti:
Awesome! Looking forward to the PR.

PR #62, opened 2021-11-08
Add basic twitter summary card.

pingortle:
This was a very quick take just to get something out there. Happy to make any changes desired.

Please let me know if you want any tests in particular.

#### No avatar
image

#### With avatar
image

Note: Twitter inherits og:title and og:description so there's no need to redefine them just for twitter.

/cc #37

### DIFF
--- a/app/components/developer_social_meta_tags_component.html.erb
+++ b/app/components/developer_social_meta_tags_component.html.erb
@@ -1,0 +1,13 @@
+<meta property="og:type" content="website">
+<meta property="og:title" content="<%= @developer.hero %>">
+<meta property="og:description" content="<%= @developer.bio %>">
+<meta property="og:url" content="<%= developer_url(@developer) %>">
+
+<% if @developer.avatar.present? %>
+  <meta property="og:image" content="<%= rails_blob_url(@developer.avatar) %>">
+<% end %>
+
+<meta name="twitter:card" content="summary">
+<% if @developer.twitter.present? %>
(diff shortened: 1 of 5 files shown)


pingortle:
Hmmm… A developer's name should be in there somewhere… Maybe it's better to make the tw title their name and the tw description their hero?

@joemasilotti your call! Wdyt?

joemasilotti:
No, this looks perfect (visually)! I'm purposely "hiding" names for now and want to expose the hero/description. Note how on the site you can't see the name, either. 

Will do a PR review today. Thanks for the contribution!

joemasilotti (changes requested):
Would love to see this moved to a tested view component just for that little bit of logic for the avatar. 

Also, some devs might not have a Twitter so that shouldn't be rendered if blank.

joemasilotti:
Please remove the string literal comment and standardize and I'll merge!

`bundle exec standardrb --fix`

pingortle:
> Please remove the string literal comment and standardize and I'll merge!
> 
> `bundle exec standardrb --fix`

Hmmm… I might've been using an old global standard (via ale) instead of the bundled one.

Thanks for handling that! 🚀

joemasilotti:
No problem! I also realized when I dove into the code that standard won't remove that comment. Nor "fix" the quotes in the tests because they use both quote types. TIL!

joemasilotti:
FYI, I refactored this after realizing we need Open Graph tags for different pages.

https://github.com/joemasilotti/railsdevs.io/pull/68

PR #64, opened 2021-11-09
chore: add user admin test

JuanVqz:
Hi! I'm not pretty much familiar with **minitest** but I'm doing my best, feedback is welcomed, thank you!

### DIFF
--- a/test/models/user_test.rb
+++ b/test/models/user_test.rb
@@ -1,0 +1,15 @@
+require "test_helper"
+
+class UserTest < ActiveSupport::TestCase
+  test "when the user is admin" do
+    user = users(:admin)
+
+    assert user.admin?
+  end
+
+  test "when the user is not admin" do
+    user = users(:with_profile_one)
+
+    assert_equal false, user.admin?
+  end
+end


test/models/user_test.rb
```diff
@@ -0,0 +1,15 @@
...
+  test "when the user is admin" do
+    user = users(:admin)
+
+    assert user.admin?
+  end
+
+  test "when the user is not admin" do
+    user = users(:with_profile_one)
+
+    assert_equal false, user.admin?
```
joemasilotti, line 13: You can use `refute` here. So, `refute user.admin?`.

joemasilotti:
Thanks for the PR! I added a comment to help you on your minitest journey :)

But I'm not going to merge this PR. The method you are testing, `#admin?`, is defined by Rails. And we don't need to test their implementation (nor our fixtures).

test/models/user_test.rb
JuanVqz: I didn't see it, I'll fix it next PR, thank you!

PR #65, opened 2021-11-09
Route developer links /developers/:id/links/:field

joemasilotti:
## Goals:

1. Only signed in users can message a developer and visit their website, Twitter, GitHub, and LinkedIn
2. Track clicks on these links

## Which enables:

1. Better messaging when trying to message a developer when not signed in (vs. the button not showing up at all)
2. Upselling a company to pay to message a developer

I know that moving links to the social profiles doesn't make a ton of sense right now. I mean, you can _see_ the developer's account name. But I'm hoping a redesign to `Developers#show` will only show those links as buttons and generic labels (e.g. "GitHub", not "@joemasilotti").

### To-do

- [x] Route links through app
- [x] Require authentication
- [ ] Analytics
- [ ] Better messaging when not signed in


app/views/developers/show.html.erb
```diff
@@ -87,7 +85,7 @@
             LinkedIn
           </dt>
           <dd class="mt-1 text-sm text-gray-900">
-            <%= link_to @developer.linkedin , "https://www.linkedin.com/in/#{@developer.linkedin}/", target: "_blank", class: "hover:text-gray-600" %>
+            <%= link_to @developer.linkedin , developer_link_path(@developer, :linkedin), target: "_blank", class: "hover:text-gray-600" %>
```
lsparlin, line 88: I like this approach! I was thinking about this issue when you first added it and this is about how I pictured it working

joemasilotti: Thanks! It gives us fine grain access control without having to change the markup of the `#show` page at all. Ideally, we can start doing analytics server side, too!

joemasilotti:
Putting this on via #72.

PR #66, opened 2021-11-09
chore: add developer validation test

JuanVqz:


JuanVqz:
I'm wondering if we have something like RSpec's `context`?

joemasilotti:
This looks great! I like to keep my test files "flat", so I rarely use context or anything like that.

PR #68, opened 2021-11-09
Refactor open graph tags

joemasilotti:
This refactor extracts a base `OpenGraphTagsComponent` that sets default values for generic pages (like the homepage). `DeveloperOpenGraphTagsComponent` then inherits from this, overriding the different values when needed.

Note how there is no longer an HTML partial for the developer component. It uses the parent component to keep the actual rendering of the tags in one place.

The PR also adds a small helper to render the Open Graph tags from the application layout.

---

Areas for improvement:

1. It's a little awkward that you have to remember to type `content_for :open_graph_tags` when rendering these. I'd love to move that into the view component partial. But I couldn't figure out how to test that.
2. It would be nice if the static strings came from the i18n file.


app/components/open_graph_tags_component.rb
```diff
@@ -0,0 +1,21 @@
+class OpenGraphTagsComponent < ViewComponent::Base
+  def title
+    "railsdevs.io"
+  end
+
+  def description
+    "Find Ruby on Rails developers looking for freelance and full-time work."
```
davegudge, line 7: Not a biggy, but this tag line differs slightly from the [README](https://github.com/joemasilotti/railsdevs.io/blame/main/README.md#L3).

> Find Rails developers looking for freelance and contract work.

On a related note: If the site is also intended for developers who are currently employed full-time, it may be worth having an availability duration option (as well as the current `available_on` date). 

An availability duration would provider developers, who are currently employed full-time, the ability to represent their notice period without continually having to update the `available_on` date. 

Perhaps the availability option is determined by if a developers current employment is either full-time vs freelance or contractor. That said, I believe contractors tend to have an end date, thus availability duration may be more suitable.

PR #76, opened 2021-11-11
Replace pagination with turbo

Tonksthebear:
Previously in #41 we discussed potentially removing unnecessary javascript, such as the pagination controller.

If this is still the direction you want to go, this PR swaps the pagination controller with turbo frame.

I am not completely happy with how the index is rendering @developers twice. It would obviously be possible to reduce the rendering by adding one more if statement, but I wasn't sure if even more logic would outweigh that benefit.

I wouldn't be offended if this PR is deemed pointless, but I figured I'd throw it out there 👍


app/views/developers/index.html.erb
```diff
@@ -12,14 +12,25 @@
...
+  <%=turbo_frame_tag "developers-#{@pagy.page}" do %>
+    <%=turbo_stream.append "developers" do %>
+      <%= render @developers %>
+      <% if @pagy.next %>
+        <%=turbo_frame_tag "developers-#{@pagy.page + 1}", src: [:developers, page: @pagy.next], loading: :lazy %>
+      <% end %>
+    <% end %>
+    <%=turbo_stream.remove "developers-#{@pagy.page}" %>
+  <% end %>
+
```
joemasilotti, line 35: Please delete this line.

app/views/developers/index.html.erb
```diff
@@ -12,14 +12,25 @@
...
   </ul>
 
-  <div class="bg-gray-100 pt-8">
+  <div class="mt-8">
     <%= pagy_links do %>
       <%== pagy_nav @pagy %>
     <% end %>
   </div>
+
+  <%=turbo_frame_tag "developers-#{@pagy.page}" do %>
```
joemasilotti, line 26: Please add a space after the `=` (3 more times in this file).

joemasilotti (changes requested):
Bravo on this! I feel like I didn't have a _really_ solid understanding of how Turbo Streams can be used until I read through this diff. Also, nice catch on the `ToggleController.js` refactor.

Please make the few code style changes I requested and we can merge this in.

app/views/developers/index.html.erb
Tonksthebear: This is actually necessary to preserve proper back button usage. If you don't remove the previous lazy-loaded frame, pressing the back button after clicking on a developer will cause duplicate loads

app/views/developers/index.html.erb
Tonksthebear: Ah darn. I thought I caught on to your desired syntax but this slipped through the cracks :(

app/views/developers/index.html.erb
joemasilotti: The blank line?

joemasilotti: The blank line?

joemasilotti (approved):

app/views/developers/index.html.erb
Tonksthebear: Oh! I mis-interpreted that. I'll get is removed now

joemasilotti:
Thanks again for this!

Next up is figuring out why I had to disable Turbo on the sign out buttons...

Tonksthebear:
Strange... I'll poke around too and let you know if I find anything next time I get a chance to hop on

joemasilotti:
Thanks! I just opened #78 which I think this PR caused. Can you please take a look? I'm going to revert the PR for now.

PR #80, opened 2021-11-11
Replace pagination with turbo

Tonksthebear:
Duplicate PR with fix for duplicating developers. Also removed blank line. This replaces #76


app/views/developers/index.html.erb
```diff
@@ -12,14 +12,24 @@
...
-  <div class="bg-gray-100 pt-8">
+  <div class="mt-8">
     <%= pagy_links do %>
       <%== pagy_nav @pagy %>
     <% end %>
   </div>
+
+  <%= turbo_frame_tag "developers-#{@pagy.page}" do %>
+    <%= turbo_stream.append "developers" do %>
+      <%= render @developers unless @pagy.page == 1%>
```
Tonksthebear, line 28: I believe this was the culprit. Turbo was adding the developers twice on the first page

joemasilotti: Ah, got it. Nice catch!

joemasilotti:
Thanks for the quick turnaround – merging!

PR #82, opened 2021-11-11
First pass at /about

joemasilotti:
Still need to tweak the copy here, but I think it's coming along nicely!


joemasilotti:
This is ready to deploy code-wise, but I'd love a second/third pass on the copy.

joemasilotti:
This PR also removes `lefthook`.

In theory, I love the idea of running `standardrb` on commit. In practice, it is too slow and gets in the way of my workflow. The alternative is to install a linter in your text editor or IDE. I'm using [ale](https://github.com/dense-analysis/ale) in my vim config.

CC @lsparlin

app/views/about/show.body.md
```diff
@@ -0,0 +1,19 @@
+### Empowering the developer
+
+I've been an independent developer for the past two years. And leads are rarely consistent. Sometimes I can barely keep up with the work and other times I struggle to find my next gig.
+
```
vamsipavanmahesh, line 4: I know you are the owner of the project, but at this point, this is a community project right, suggest you remove the wording "I". Just my opinion

vamsipavanmahesh: actually, never mind, this might not be a true community project yet like rails.

PR #83, opened 2021-11-11
Field specific form errors

Tonksthebear:
Screen Shot 2021-11-11 at 2 59 10 PM

When I was adding extra developers to check pagination for #80, I noticed I had to keep scrolling to the top in order to check the errors. I wanted to make this more intuitive.

In #32, I discussed using a yaml file to store utility classes for easy references. I extracted helpers I use privately into a super small gem [classy-yaml](https://github.com/Tonksthebear/classy-yaml) to provide this feature. I needed this feature because I wanted to avoid adding custom css classes.

I know adding gems to a project is a big risk, so I completely understand wanting to swap it out and/or pull that logic into this project directly. I made it to use with my other projects, so no lost work.

I also made sure [classy-yaml](https://github.com/Tonksthebear/classy-yaml) can work with a ViewComponent sidecar yaml file (the README covers this), so with this gem, it would be possible to clean up other components that use a hash for the class names and then custom getters.

This was another unprompted feature, so I'm good with whatever you want to do with this PR!


Tonksthebear:
Yikes, forgot to check the tests. I'll get those checks cleared up in case this PR is a direction you want to go 👍

config/application.rb
```diff
@@ -41,5 +41,15 @@ class Application < Rails::Application
 
     # Run background jobs asynchronously in an in-process thread pool.
     config.active_job.queue_adapter = :async
+
+    # Add error span to form fields with errors
+    config.action_view.field_error_proc = proc { |html_tag, instance|
```
joemasilotti, line 46: Whoa, TIL!

Is it possible to move the proc (or this entire section of the configuration) to a different file? I don't love the idea of this much "stuff" inside a configuration file.

config/utility_classes.yml
```diff
@@ -0,0 +1,10 @@
+input:
+  text_field:
+    base: "shadow-sm block w-full sm:text-sm rounded-md"
+    clean: "focus:ring-gray-500 focus:border-gray-500 border-gray-300"
+    error: "focus:ring-red-500 focus:border-red-500 border-red-300"
+  text_area:
+    base: "shadow-sm block w-full sm:text-sm border rounded-md"
+    clean: "focus:ring-gray-500 focus:border-gray-500 border-gray-300"
+    error: "focus:ring-red-500 focus:border-red-500 border-red-300"
+
```
joemasilotti, line 10: Please remove this empty line.

tailwind.config.js
```diff
@@ -11,10 +11,11 @@ module.exports = {
 
   purge: [
     "./app/views/**/*.html.erb",
-    "./app/components/**/*.rb",
```
joemasilotti, line 14: Can you explain the changes to this file, please?

joemasilotti (commented):
OK, this is super cool!

I'm a bit wary on depending on your gem because it is so new. Do you have any idea what kind of effort and amount of code it would take to bring it into this codebase?

I'm not 100% against using the gem, I just want to make sure I understand the full picture.

joemasilotti:
Follow up... is it possible to move the error label next to the actual label? So this screenshot would read "Email can't be blank" in red above the input field?

image

config/application.rb
Tonksthebear: I agree, it's not the best spot. Would this be good for an initializer then?

joemasilotti: I've never added a proc to a configuration, but an initializer does sound like a good spot to start!

tailwind.config.js
Tonksthebear: Yup! Rather than having
```
purge: [
  "./app/components/**/*html.erb",
  "./app/components/**/*.rb",
  "./app/components/**/*.html",
```

You can specify all extensions with `{extensions}

So the above just becomes
```
purge: [
  "./app/components/**/*.{html.erb,rb,html}",
]
```

I personally prefer the single line approach as it's easier to lump concepts together, but I'm completely happy separating it back out

Tonksthebear:
> OK, this is super cool!
> 
> I'm a bit wary on depending on your gem because it is so new. Do you have any idea what kind of effort and amount of code it would take to bring it into this codebase?
> 
> I'm not 100% against using the gem, I just want to make sure I understand the full picture.

Oh I can bring it on over if that's what you'd prefer, I don't take it personally. It's small in scope so it wouldn't be hard. I  do plan on still adding tests and making it more "complete" though, as it's definitely useful outside of a single project

Tonksthebear:
Updated view:
Screen Shot 2021-11-11 at 5 23 43 PM

It was more trouble than it was worth trying to retrieve the field name from the error proc, and this does look less repetitive

tailwind.config.js
joemasilotti: Ahh, got it. That makes a ton of sense. Thanks for the explanation!

joemasilotti:
Making progress! My last request is to make the label red, too, if possible. Also, did you push the design change yet? I'm not seeing it locally.

As for the gem, let me sleep on it. I love the idea of supporting something new and exciting (and the tests help!). But I'm a little wary of being left with supporting it if it becomes abandoned.

config/initializers/form_errors.rb
```diff
@@ -0,0 +1,11 @@
+# Add error span to form fields with errors
+Rails.application.config.action_view.field_error_proc = proc do |html_tag, instance|
```
joemasilotti, line 2: Does `ActionView::Base.field_error_proc` work here?

Tonksthebear: Looks like it does! Just swapped it out

Tonksthebear:
> Making progress! My last request is to make the label red, too, if possible. Also, did you push the design change yet? I'm not seeing it locally.
> 
> As for the gem, let me sleep on it. I love the idea of supporting something new and exciting (and the tests help!). But I'm a little wary of being left with supporting it if it becomes abandoned.

Made the label red! It was more of a pain than you'd think but I think the solution covers most cases. As far as why you weren't seeing the design change, I'm not sure... it should all be pushed and it has been working for me.

I'm fine with whatever you choose! I don't have enough GitHub history to plea my commitment so I understand the wariness. All I can say is it's useful to me long term so regardless of what you choose I don't view it as wasted effort at all 👍

config/initializers/form_errors.rb
```diff
@@ -0,0 +1,15 @@
+# Add error span to form fields with errors
+ActionView::Base.field_error_proc = proc do |html_tag, instance|
+  if /<label/.match?(html_tag)
+    # Remove text color if it exists so that container color takes precedence
```
Tonksthebear, line 4: I played around with a couple options. Originally, I wanted to avoid regex and just use a container. I tried putting `text-red-600!` on the container, but of course the specificity of having text color classes on the label directly outweighed this.

I also debated swapping any text coloring on the label element class, but figured there'd have to be way more logic to handle cases where there aren't any class attributes. Ultimately, clearing any text colors then adding a container seemed to be the shortest and most inclusive solution. 

Let me know if there is a better way!

Gemfile
```diff
@@ -12,6 +12,7 @@ gem "pg", "~> 1.1"
 gem "puma", "~> 5.0"
 gem "sass-rails", ">= 6"
 gem "view_component", "~> 2.42"
+gem "classy-yaml", "~> 0.5"
```
joemasilotti, line 15: Please move this gem to the bottom group and sort alphabetically. I'm trying to keep this area for Rails and core gems.

tailwind.config.js
```diff
@@ -11,10 +11,11 @@ module.exports = {
...
-    "./app/components/**/*.html.erb",
+    "./app/components/**/*.{rb,html,html.erb,yml}",
     "./app/helpers/**/*.rb",
-    "./app/javascript/**/*.js"
+    "./app/javascript/**/*.js",
+    "./config/initializers/form_errors.rb",
+    "./config/utility_classes.yml"
```
joemasilotti, line 18: Please sort this array.

config/initializers/form_errors.rb
joemasilotti: I like this approach, I just wish it wasn't so much code. It took me a few minutes to wrap my head around what is going on.

Can you please...

1. Add a comment on line 8 explaining what is going on? Similar to the comment on line 4.
2. Extract the actual algorithm to a Ruby model, then call said model from this initializer.
2. Add an integration test that verifies `text-red-600` is added to a label when trying to submit an invalid form.
3. Add a unit test for the new PORO that passes in an HTML block and verifies the result is correct.

joemasilotti (changes requested):
OK, I think this should be the last round of feedback!

I'm really loving how this is coming along. It feels much cleaner than anything I've done in the past.

I'm cool with including your gem for now, but I will be keeping a close eye on it to see if we ever want to implement in-house in the future.

joemasilotti:
And please resolve conflicts and/or merge in `main`.

Tonksthebear:
Ah I'm sorry about the ugly last push. My head isn't on right today... give me a few minutes and I'll clear it all up

app/models/tag_color_manipulator.rb
```diff
@@ -0,0 +1,15 @@
+class TagColorManipulator
```
Tonksthebear, line 1: I'm not well versed in PORO design so let me know whatever you want changed here

config/initializers/form_errors.rb
Tonksthebear: I may have gone overboard with the comments, I just wanted to be sure to save time for any one trying to read this code. I'm happy to remove any unnecessary comments

tailwind.config.js
Tonksthebear: Sorted

test/models/tag_color_manipulator_test.rb
```diff
@@ -0,0 +1,12 @@
+require "test_helper"
+
+class TagColorManipulatorTest < ActiveSupport::TestCase
```
Tonksthebear, line 3: Is this the type of test you wanted for the PORO

Gemfile
Tonksthebear: Done!

test/models/tag_color_manipulator_test.rb
joemasilotti: This is looking great! And the integration test looks perfect.

I just want to take it one step farther. I don't want _any_ logic in the initializer. Essentially, I want that entire file to read... (Feel free to mess with the PORO name here.)

```ruby
ActionView::Base.field_error_proc = proc do |html_tag, instance|
  FieldErrorTagBuilder.new(html_tag, instance).error_field
end
```

Once all of the logic is in the PORO we can test everything, including the tag index logic. It might also make it a bit easier to read through because the tests call out exactly what is going on.

joemasilotti (changes requested):
So close! Thanks for continuing to iterate on this. I know I'm being picky :)

config/application.rb
jacobdaddario: Oh, I've looked at this before, and the `field_error_proc` is something... I'd looked at trying to make a PR to Rails to allow it to be configurable in form builders, but it's defined on `ActionView::Base`. I think it just has to live at the configuration level or in an initializer because of how Rails is written right now. It's like my open-source dream to find a way to disentangle this ancient piece of code into something more useful.

Tonksthebear:
> So close! Thanks for continuing to iterate on this. I know I'm being picky :)

Oh I'm enjoying the back and forth. I'm here to learn so this is perfect.

I went ahead and made those changes. I'm still not confident in my PORO design or the tests I constructed, but I gave it an attempt. I'm happy to make more changes so don't hold back!

test/models/field_error_tag_builder_test.rb
```diff
@@ -0,0 +1,30 @@
+require "test_helper"
+
+class FieldErrorTagBuilderTest < ActiveSupport::TestCase
+  setup do
+    @spoofed_instance = Struct.new(:error_message).new(["fake error"])
```
joemasilotti, line 5: Let's rename this instance variable to `@instance`.

joemasilotti (changes requested):
Oh man, I love `FieldErrorTagBuilder`! I could even see us extracting this to a gem down the road.

Tests look great too. I made one tiny request (variable rename). After that this is ready to be merged.

Tonksthebear:
Agreed, it's the cleanest way I've handled form errors and I'm definitely going to be swapping over my code to this

joemasilotti (approved):

joemasilotti:
I love how this turned out. Thanks again for the PR!

PR #85, opened 2021-11-12
developers.looking_for_work

joemasilotti:


db/migrate/20211112035233_add_looking_for_work_to_developers.rb
```diff
@@ -0,0 +1,14 @@
+class AddLookingForWorkToDevelopers < ActiveRecord::Migration[7.0]
+  def up
+    add_column :developers, :looking_for_work, :boolean
+
+    Developer.where("available_on < ?", Date.today + 1.year).update_all(looking_for_work: true)
+    Developer.where(looking_for_work: nil).update_all(looking_for_work: false)
+
+    change_column :developers, :looking_for_work, :boolean, null: false, default: true
+  end
```
davegudge, line 9: This _should_ be equivalent, i.e. add a non-null column, with a default, then update the developers that are not _looking for work_:

```suggestion
  def up
    add_column :developers, :looking_for_work, :boolean, null: false, default: true

    Developer.where("available_on >= ?", Date.today + 1.year).update_all(looking_for_work: false)
  end
```

It would be sensible to add an index to this column too e.g:

```
add_column :developers, :looking_for_work, :boolean, null: false, default: true, index: true
```

app/controllers/developers_controller.rb
```diff
@@ -46,6 +46,10 @@ def show
 
   private
 
+  def developers
+    Developer.looking_for_work.most_recently_added.with_attached_avatar
```
vamsipavanmahesh, line 50: As this is getting duplicated and this is probably going to be used in multiple places, adding a class method on the Developer model is not a bad idea

db/migrate/20211112035233_add_looking_for_work_to_developers.rb
vamsipavanmahesh: I don't think adding an index to a column whose cardinality is less will give us any improvements, let me research more and get back on this

db/migrate/20211112035233_add_looking_for_work_to_developers.rb
```diff
@@ -0,0 +1,14 @@
+class AddLookingForWorkToDevelopers < ActiveRecord::Migration[7.0]
+  def up
+    add_column :developers, :looking_for_work, :boolean
+
+    Developer.where("available_on < ?", Date.today + 1.year).update_all(looking_for_work: true)
```
vamsipavanmahesh, line 5: Do we have any plans to communicate with all the registered profiles about this change? or it's implicit

db/migrate/20211112035233_add_looking_for_work_to_developers.rb
vamsipavanmahesh: confirming that there are no gains when you add an index, whose cardinality is less.

References:
https://stackoverflow.com/questions/1844991/indexing-boolean-fields
https://stackoverflow.com/a/10524734/1395129

davegudge: Thanks @vamsipavanmahesh that's good to know.

davegudge: I agree that an index should **not** be added, but it does look like there are certain scenarios where an index could be beneficial.

For example, 10,000 developers register on the site and only 10 developers are 'looking for work', apparently, indexing would be more efficient. There's a good explanation here: https://stackoverflow.com/a/43273820

I guess the indexing will be determined by the dataset, which cannot be predicted in advance.

vamsipavanmahesh: Yeah I hope this project won't have such data haha .. but let's see depending on the data and queries may be we can defer this

joemasilotti: Good call on the code suggestion, @davegudge! I accepted it.

I'm going to hold off on the index for now. Thanks for the research.

db/migrate/20211112035233_add_looking_for_work_to_developers.rb
joemasilotti: I was planning on it being implicit, but would love to discuss more in the Discussion forum #86.

app/controllers/developers_controller.rb
joemasilotti: Once we use it three times we can extract something. But for now I like how explicit the query is.

joemasilotti:
@davegudge and @vamsipavanmahesh, if you could take a glance (see the original comment) I would appreciate it!

app/views/developers/_details.html.erb
```diff
@@ -0,0 +1,80 @@
+<h2 class="sr-only">Details</h2>
+<div class="space-y-5">
+  <div class="flex items-center space-x-2">
+    <%= render AvailabilityComponent.new(developer: developer, show_unavailable_icon: true) %>
+  </div>
+
+  <% if developer.actively_looking? %>
```
joemasilotti, line 7: I feel like there might be a better want to do this? Probably a view component.

app/views/developers/_details.html.erb
```diff
@@ -0,0 +1,80 @@
...
+      <span class="text-gray-900 text-sm font-medium">Open, but not actively looking for work</span>
+    </div>
+  <% elsif developer.not_interested? %>
+    <div class="flex items-center space-x-2">
+      <%= inline_svg_tag "icons/solid/x_circle", class: "h-5 w-5 text-gray-400" %>
+      <span class="text-gray-900 text-sm font-medium">Not interested in work right now</span>
+    </div>
+  <% end %>
+
+  <% if developer.role_type? %>
```
joemasilotti, line 24: Same here. Another component with `#render?` implemented.

app/views/developers/_details.html.erb
```diff
@@ -0,0 +1,80 @@
...
+  <div class="mt-6 border-t border-gray-200 py-6 space-y-8">
+    <div>
+      <ul role="list" class="space-y-3">
+        <% if developer.website.present? %>
+          <li>
+            <%= render ExternalLinkComponent.new(developer.website) %>
+          </li>
+        <% end %>
+
+        <% if developer.github.present? %>
```
joemasilotti, line 51: These could also probably be extracted. Params being:

* `developer`
* `attribute`
* `icon`
* `url_prefix`

app/views/developers/_form.html.erb
```diff
@@ -83,22 +83,79 @@
...
+                    <%= form.label :search_status, value: :actively_looking, class: "font-medium text-gray-700" %>
+                    <p class="text-gray-500">Your profile can get featured on the homepage.</p>
+                  </div>
+                </div>
+                <div class="flex items-start">
+                  <div class="flex items-center h-5">
+                    <%= form.radio_button :search_status, :open, class: "focus:ring-gray-500 h-4 w-4 text-gray-600 border-gray-300" %>
+                  </div>
+                  <div class="ml-3 text-sm">
+                    <%= form.label :search_status, "Open, but not actively looking", value: :open, class: "font-medium text-gray-700" %>
```
joemasilotti, line 109: This string is duplicated. Should probably move it to i18n.

app/views/developers/_form.html.erb
```diff
@@ -83,22 +83,79 @@
...
+
+            <fieldset>
+              <div>
+                <legend class="text-base font-medium text-gray-900">Role type</legend>
+                <p class="text-sm text-gray-500">Select all roles that you would consider taking.</p>
+              </div>
+              <div class="mt-4 space-y-4">
+                <div class="flex items-center">
+                  <%= form.check_box :part_time_contract, class: "focus:ring-gray-500 h-4 w-4 text-gray-600 border-gray-300 rounded" %>
+                  <%= form.label :part_time_contract, "Part-time contract", class: "ml-3 block text-sm font-medium text-gray-700" %>
```
joemasilotti, line 133: This can definitely move to i18n.

app/views/developers/show.html.erb
```diff
@@ -2,94 +2,50 @@
...
+                </div>
+              </div>
+
+              <aside class="mt-8 lg:hidden">
+                <%= render "details", developer: @developer %>
+              </aside>
+
+              <div class="py-3 lg:pt-6 lg:pb-0">
+                <h2 class="sr-only">Description</h2>
+                <div class="prose max-w-none">
```
joemasilotti, line 37: I think this should be `max-w-prose`?

vamsipavanmahesh:
I have looked at the original comment. product-wise, it looks great.

On the tech side:
Consider using string enums, instead of integer enum .. if you keep integer enums, the table will have values that are not comprehensible outside of AR

Have you considered

Developer has_one role_type instead of keeping all the role_type columns in the developer. Role_type could be a model on its own and could have its own sort of validations etc in the future

Just like we don't put all the address fields in the user table, a similar analogy

joemasilotti:
> Consider using string enums, instead of integer enum .. if you keep integer enums, the table will have values that are not comprehensible outside of AR

That's a good point! But I'm not too worried about ever reading this data outside of an Active Record context.

> Developer has_one role_type instead of keeping all the role_type columns in the developer. Role_type could be a model on its own and could have its own sort of validations etc in the future

I hadn't considered that. My only hesitation is that every query for developers would then require a join (e.g. the home page, the index, if/when we implement search, etc.). But it might be cleaner in the long run. Let me think on that.

Thanks for the review!

davegudge:
> @davegudge and @vamsipavanmahesh, if you could take a glance (see the original comment) I would appreciate it!

👍🏼 I'll get on this today.

db/migrate/20211116041805_add_search_status_and_role_preferences_to_developers.rb
```diff
@@ -0,0 +1,9 @@
+class AddSearchStatusAndRolePreferencesToDevelopers < ActiveRecord::Migration[7.0]
+  def change
+    add_column :developers, :search_status, :integer
+
+    add_column :developers, :part_time_contract, :boolean
+    add_column :developers, :full_time_contract, :boolean
+    add_column :developers, :full_time_employment, :boolean
```
davegudge, line 7: To represent some checkboxes, I recently used an enum on the model and stored the values in one string array column:

```suggestion
    add_column :developers, :role_types, :string, array: true
```

The model had a scope similar to this: 

```ruby
  scope :with_role, ->(value) do
    where arel_table[:role_types].array_contains value
  end
```

The scope is reliant on a method I have in `lib/rails_ext/arel/predications.rb`:

```ruby
module Arel
  module Predications
    ...
    def array_contains(*rhs)
      Nodes::InfixOperation.new "@>", self, Nodes.build_quoted(to_array_literal(rhs))
    end
    ...
  end
end
```

app/components/role_type_component.html.erb
```diff
@@ -0,0 +1,15 @@
+<div class="mt-6 border-t border-gray-200 pt-6 space-y-8">
+  <div>
+    <h2 class="text-sm font-medium text-gray-500">Interested in roles</h2>
+    <ul role="list" class="mt-3 space-y-3">
+      <% role_types.each do |role| %>
+        <% if developer.send("#{role}?") %>
```
davegudge, line 6: Is this equivalent?

```suggestion
      <% role_types.select { |rt| developer.public_send "#{role}?" }.each do |role| %>
```

Alternatively, you could add a `developer_role_types` method to the `RoleTypeComponent` to DRY up the logic e.g.

```ruby
  def developer_role_types
    role_types.select do |rt| 
      developer.public_send "#{role}?"
    end
  end
```

thus simplifying the view logic to `<% developer_role_types.each do |role| %>` and the component's render logic to:

```ruby
  def render?
    developer_role_types.any?
  end
```

davegudge (approved):
Hey @joemasilotti I've run out of time reviewing this. It all looks good 👍🏼 

I've added a couple of alternative solutions, simply to share different ways of doing things. I'm not suggesting anything needs to change. 

I think an `IconComponent` would be very useful to centralise the icon logic.

fabionl:
Page looks great! Amazing design skills. 

Code also looks great. I made a couple of very minor comments. Hope they are helpful.

A couple notes on architecture/implementation:
- I do prefer string enums, but using integer isn’t a big issue, and at least making the integer values explicitly avoids future issues.
- I tend to prefer making the search status a model as suggested by others. I agree it will lead to using joins in a lot of places, but in my experience the database handles it quite well. But that being said, I don’t think it is a decision you must make now. For now, the requirements seem simple enough to keep in the same model.

app/components/availability_component.rb
```diff
@@ -1,6 +1,15 @@
 class AvailabilityComponent < ApplicationComponent
-  def initialize(developer:)
+  def initialize(developer:, show_unavailable_icon: false)
```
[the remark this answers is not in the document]
joemasilotti: Showing the unavailable icon on the index pages looks weird. But it makes sense in the detail page.

I'd rather keep the logic in Ruby and not add additional JavaScript to hide elements for something like this.

db/migrate/20211116041805_add_search_status_and_role_preferences_to_developers.rb
joemasilotti: Ah, very cool! I've shied away from array columns in the past because if/when you change the values the migrations can be a pain. I can see that helper and scope being super useful, though! I'll keep this in my back pocket if we ever migrate to an array.

I'm leaning more towards a `has_one` first-class model now.

app/components/role_type_component.html.erb
joemasilotti: Ohh, the methods added to `RoleTypeComponent` are a great call! I just applied that suggestion.

app/components/availability_component.html.erb
```diff
@@ -1,4 +1,11 @@
-<span class="text-gray-900 text-sm <%= "font-medium" if @developer.available_now? %>">
-  <%= t(".#{@developer.availability_status}_html", date: date, date_in_words: date_in_words) %>
-</span>
-<%= inline_svg_tag "icons/solid/check_circle.svg", class: "h-5 w-5 text-green-400 mx-2" if @developer.available_now? %>
+<div class="flex items-center text-gray-900">
+  <% if available? %>
+    <%= inline_svg_tag "icons/solid/check_circle.svg", class: "h-5 w-5 text-green-500" %>
+  <% elsif show_unavailable_icon? %>
+    <%= inline_svg_tag "icons/solid/x_circle.svg", class: "h-5 w-5 text-gray-400" %>
+  <% end %>
```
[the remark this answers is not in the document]
joemasilotti: Great call. I extracted `AvailabilityComponent#availability_icon`. An `IconComponent` could work, too, but I don't want that until we need to use it somewhere else.

joemasilotti:
Thanks for the review and suggestions @davegudge! I applied a few of your changes (buried in the outdated and resolved comment threads).

@fabionl, I'm leaning towards making a first-class model now. Expect something soon.

app/components/availability_component.rb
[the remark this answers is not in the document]
davegudge: Yes, it makes more sense to keep the logic within the component 👍🏼

db/migrate/20211116041805_add_search_status_and_role_preferences_to_developers.rb
davegudge: > if/when you change the values the migrations can be a pain

Yes, I know what you mean. I stuck to ruby objects for the migration I implemented recently e.g.

```ruby
    Developer.find_each do |developer|
      developer.role_types << :part_time_contract
      developer.role_types << :full_time_employment unless some condition
      developer.save(validate: false)
    end
```

app/components/role_type_component.html.erb
```diff
@@ -0,0 +1,15 @@
+<div class="mt-6 border-t border-gray-200 pt-6 space-y-8">
+  <div>
+    <h2 class="text-sm font-medium text-gray-500">Interested in roles</h2>
+    <ul role="list" class="mt-3 space-y-3">
+      <% role_types.each do |role| %>
```
davegudge, line 5: Following on from 3484888, this can now be updated to:

```suggestion
      <% developer_role_types.each do |role| %>
```

and the nested condition can be removed.

app/components/search_status_component.html.erb
```diff
@@ -0,0 +1,16 @@
...
+  <div class="flex items-center space-x-2">
+    <%= inline_svg_tag "icons/solid/search_circle", class: "h-5 w-5 text-gray-400" %>
+    <span class="text-gray-900 text-sm font-medium"><%= humanize :open %></span>
+  </div>
+<% elsif @developer.not_interested? %>
+  <div class="flex items-center space-x-2">
+    <%= inline_svg_tag "icons/solid/x_circle", class: "h-5 w-5 text-gray-400" %>
+    <span class="text-gray-900 text-sm font-medium"><%= humanize :not_interested %></span>
+  </div>
+<% end %>
```
davegudge, line 16: The html within each condition is very similar. 

You could add `icon` and `icon_attrs` methods to the `SearchStatusComponent` (similar to what was done in 
76cc310) and your erb file becomes more like:

```erb
  <div class="flex items-center space-x-2">
    <%= inline_svg_tag icon, icon_attrs %>
    <span class="text-sm font-medium text-#{@developer.actively_looking? ? "green-700" : "gray-900"} ">
      <%= humanize @developer.search_status %>
    </span>
  </div>
```

app/views/developers/_form.html.erb
```diff
@@ -83,22 +83,79 @@
     <div class="bg-white shadow mt-8 px-4 py-5 lg:rounded-lg sm:p-6">
       <div class="md:grid md:grid-cols-3 md:gap-6">
         <div class="md:col-span-1">
-          <h3 class="text-lg font-medium leading-6 text-gray-900">Availability</h3>
-          <p class="mt-1 text-sm text-gray-500">If you're currently looking for work, how soon can you start?</p>
+          <h3 class="text-lg font-medium leading-6 text-gray-900">Work preferences</h3>
```
davegudge, line 86: Obviously it's totally up to you, but if you keep all of these strings in the `i18n` files, the app becomes much easier to translate into another language later down the line. 

If you are planing on taking that approach, I would recommend enabling this config in `development.rb`

```
  # Raises error for missing translations
  config.action_view.raise_on_missing_translations = true
```

This will ensure an exception is thrown (in dev) if a translation is missing.

davegudge (approved):

app/components/availability_component.html.erb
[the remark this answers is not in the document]
davegudge: I was thinking the `IconComponent` would encapsulate the [default] icon logic, thus it would be called anywhere there is currently an `inline_svg_tag`. These calls are always specifying `class: "h-5 w-5` which could form the basis of the class on the component. This would probably fall outside the scope of this PR.

app/components/role_type_component.html.erb
joemasilotti: Moving `RoleType` to a db-backed model cleaned this up a bit!

app/views/developers/_form.html.erb
joemasilotti: No, I agree. I'm slowly moving all the strings into the translation. I added #110 to keep track of this separately.

I also enabled `raise_on_missing_translations` for both development and test environments.

app/components/search_status_component.html.erb
joemasilotti: Great call!

joemasilotti:
I really like how this ended up. Lot's of good view components and the data modeling feels solid.

Thanks for the feedback and reviews everyone! I appreciate it.

app/components/role_type_component.html.erb
davegudge: I _think_ it should still be possible to remove the nested condition by using the method you've defined on the `RoleTypeComponent`: 

```suggestion
      <% selected_role_types.each do |role| %>
```

app/components/search_status_component.rb
```diff
@@ -0,0 +1,47 @@
+class SearchStatusComponent < ApplicationComponent
+  attr_reader :developer
+
+  def initialize(developer)
+    @developer = developer
+  end
+
+  def render?
+    developer.actively_looking? || developer.open? || developer.not_interested?
```
davegudge, line 9: A few suggestions to reduce the logic on this line: 

```suggestion
    developer.search_status
```

or 

```suggestion
    developer.search_status.present?
```

or

```suggestion
    !developer.search_status.nil?
```

app/components/search_status_component.rb
```diff
@@ -0,0 +1,47 @@
...
+    inline_svg_tag "icons/solid/#{icon}.svg", class: icon_classes.join(" ")
+  end
+
+  def text_classes
+    if developer.actively_looking?
+      "text-green-700"
+    elsif developer.open? || developer.not_interested?
+      "text-gray-900"
+    end
+  end
```
davegudge, line 22: The below is an alternative that would ensure `text-gray-900` is used [as a default] if any further statuses were added: 

```suggestion
  def text_classes
    return "text-green-700" if developer.actively_looking?

    "text-gray-900" # default colour unless actively looking 
  end

app/components/search_status_component.rb
joemasilotti: Good call!

app/components/search_status_component.rb
joemasilotti: Good call again. Having defaults is definitely the way to go.

PR #87, opened 2021-11-12
chore: add gemfile production group

JuanVqz:
do we need a honeybadger gem on development?

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -28,13 +28,16 @@ group :test do
   gem "capybara"
 end
 
+group :production do
+  gem "honeybadger", "~> 4.0"
+end
+
 # Point at main until Rails 7 changes are released.
 gem "devise", github: "heartcombo/devise"
 
 gem "tzinfo-data", platforms: [:mingw, :mswin, :x64_mingw, :jruby]
 
 gem "aws-sdk-s3", "~> 1", require: false
-gem "honeybadger", "~> 4.0"
 gem "inline_svg", "~> 1.7"
 gem "mailgun-ruby", "~> 1.2"
 gem "noticed", "~> 1.4"


Gemfile
```diff
@@ -28,13 +28,16 @@ group :test do
   gem "capybara"
 end
 
+group :production do
+  gem "honeybadger", "~> 4.0"
```
vamsipavanmahesh, line 32: Why do we want to do this? IMHO except local machine all errors should be logged once the code goes on cloud, even on non production envs such as UAT, Staging, development

joemasilotti: What's the difference between development and local machine?

The site only has dev (local) and production right now. So moving this gem to the production-only group make sense to me.

joemasilotti:
> do we need a honeybadger gem on development?

Nope, I don't believe we do. It was only there to test the error reporting during setup.

Can you please confirm that when you move the gem to production, and run `bundle install`, that raising an error in development doesn't interfere at all with Honeybadger?

Gemfile
vamsipavanmahesh: local is basically developers machine, dev environment could be a hosted environment, staging/uat could be just before prod

dev environment is usually unstable, branch testing etc can happen there

staging is stable

I don't know what other environments are present before prod, but if there are any, ideally it's good if honey badger could catch them before going to prod

joemasilotti:
On second thought, and after reading @vamsipavanmahesh's comments, I'm leaning towards keeping this in all environments.

Keeping Honeybadger in all environments will make creating a staging environment one step easier. I'm going to close this PR, but thanks for sparking the discussion!

JuanVqz:
I had a long weekend, I could check today.

I think we still have another option when we need staging.

```
group :production, :staging do
 gem...
end
```

PR #88, opened 2021-11-13
Don't show add your profile button when profile is already there

vamsipavanmahesh:
- currently, there is add your profile button shown even when the profile is added

- also added an instance method has_developer_profile? which can be used across views

### DIFF
--- a/app/models/user.rb
+++ b/app/models/user.rb
@@ -10,4 +10,8 @@ class User < ApplicationRecord
   has_many :notifications, as: :recipient
 
   scope :admin, -> { where(admin: true) }
+
+  def has_developer_profile?
+    developer&.persisted?
+  end
 end
--- a/app/views/home/show.html.erb
+++ b/app/views/home/show.html.erb
@@ -17,9 +17,11 @@
     <h3 class="text-lg leading-6 font-medium text-gray-900">
       Developers available now
     </h3>
-    <div class="mt-3 sm:mt-0 sm:ml-4">
-      <%= link_to "Add your profile", new_developer_path, class: "inline-flex items-center px-4 py-2 border border-gray-300 rounded-md shadow-sm text-sm font-medium text-gray-700 bg-white hover:bg-gray-50 focus:outline-none focus:ring-2 focus:ring-offset-2 focus:ring-gray-500" %>
-    </div>
+    <% unless current_user&.has_developer_profile? %>
+      <div class="mt-3 sm:mt-0 sm:ml-4">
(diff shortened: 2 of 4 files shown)


config/environments/test.rb
```diff
@@ -44,6 +44,10 @@
   # Store test uploads on the disk in the temp folder
   config.active_storage.service = :test
 
+  config.action_mailer.delivery_method = :test
+  config.action_mailer.perform_deliveries = false
+  config.action_mailer.default_url_options = {host: "your-host-name"}
+
```
vamsipavanmahesh, line 50: This is added because my test is creating user, and creating user is sending email which is causing error

joemasilotti: We shouldn't need to create users under test just yet. You should be able to test your code with `users(:with_available_profile)` vs. `users(:without_profile)`.

joemasilotti (changes requested):
I'd prefer to handle this in a Pundit authorization file versus view logic.

Instead of _not_ rendering the button, I'd prefer that when you access `/developers/new` and you already have a developer profile, you are redirected to edit your profile.

That should make for a much smaller change set – and it will continue to work wherever we link to `new_developer_path` in the future.

vamsipavanmahesh:
I agree with handling in the pundit layer in addition to not showing the button, because there's a button on the top which already says Edit your profile, and now this button says add your profile. It does not make sense visually. Thoughts?

vamsipavanmahesh:
Screenshot 2021-11-13 at 11 54 53 PM

In this image, you can see it's already throwing error, when trying to enter /developers/new

also, you can see on the top right that there is a button Edit your profile, when this button is there, why would we need to Add your profile button?
cc @joemasilotti

vamsipavanmahesh:
>  I'd prefer that when you access /developers/new and you already have a developer profile, you are redirected to edit your profile

I am okay to implement this if you can help me understand the use case, IMO `new_developer_path` is not something we can use lightly or everywhere

joemasilotti:
I'm trying to avoid having an `if/else` statement every time we need to display this link. I would like it handled in two ways, one for the UI and one as a backup layer.

1. Extract a `ProfileLinkComponent` view component
    * `def initialize(user, options = {})`
    * Render a link, passing in `options` to `link_to`
    * If the `user` has a developer profile:
        1. Set the link's text to "Edit your profile"
        2. Link to `edit_developer_profile(user.developer)`
    *  Otherwise:
        1. Set the link's text to "Add your profile"
        2. Link to `new_developer_profile`
1. Tweak the policy to redirect to a user's existing developer profile if `new_developer_profile` visited

Happy to take the first one in one PR and the second in another since they solve two problems.

vamsipavanmahesh:
For point 2
> Tweak the policy to redirect to a user's existing developer profile if new_developer_profile visited

I have raised a PR https://github.com/joemasilotti/railsdevs.io/pull/92

Point 1 will take time for me, I am not familiar with view components rn, will check and get back

Closing this PR

PR #97, opened 2021-11-15
[bugfix] Respect new lines on the developers/show page

caseyjenks:
## Purpose

The developers/show page was not respecting new lines in the developer's bio text. I believe the intent is to keep this field relatively short, but I found myself wanting to split my bio into 2 small paragraphs. Perhaps discussing and implementing a character limit is a good next step for a separate PR.

### Before Screenshot
image

### After Screenshot
image

## Approach
Uses the `simple_format` method to output sanitized HTML that respects new lines

https://apidock.com/rails/ActionView/Helpers/TextHelper/simple_format

### Testing
I do not believe this warrants any testing since it's a one line fix that uses a pretty standard `ActionView` helper

### DIFF
--- a/app/views/developers/show.html.erb
+++ b/app/views/developers/show.html.erb
@@ -93,7 +93,7 @@
           Bio
         </dt>
         <dd class="mt-1 text-sm text-gray-900 space-y-5">
-          <p><%= @developer.bio %></p>
+          <%= simple_format @developer.bio %>
         </dd>
       </div>
     </dl>


joemasilotti:
Oh, good call. I didn't know about this helper!

I'll merge this PR, but I think we need to address long bios on the index view. Your profile, for example, takes up the whole screen on mobile. 

image

Can you add an issue to truncate or CSS-ellipse bios on index pages, please?

PR #98, opened 2021-11-15
Fix main content area on developers/show from butting up against site footer

caseyjenks:
## Purpose

The main content area on the developers/show page was butting right up against the footer. This PR fixes that.

### Screenshots
Before:
image

After(desktop):
image

After(mobile):
image

## Approach

Add `mb-16` to the outer most div on developers/show page. `mb-16` seemed to be the standard elsewhere across the site, so I rolled with that.

### DIFF
--- a/app/views/developers/show.html.erb
+++ b/app/views/developers/show.html.erb
@@ -2,7 +2,7 @@
   <%= render DeveloperOpenGraphTagsComponent.new(developer: @developer) %>
 <% end %>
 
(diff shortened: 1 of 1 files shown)


joemasilotti:
Thanks for the PR! Can you take a screenshot on mobile for documentation sake?

caseyjenks:
@joemasilotti done!

joemasilotti:
Awesome, thanks for the contribution!

PR #99, opened 2021-11-15
Fixes CTA buttons with different texts

caseyjenks:
## Purpose & Approach
The header's main CTA read "Edit your Profile" when signed in, and "Add your Profile" when not signed in.
The CTA on the developer listing views, on both homepage and developers/index pages both read "Add your Profile" even when signed in and a developer profile was already created.

I originally thought making the button dynamic as it's done in the header would be the way to go, but then thought perhaps removing the button altogether would be the way to go since it's a duplicate of the big CTA In the header. If you think this is the wrong path, let me know and we can go with keeping it + making it dynamic and extracting into some sort of component.

image

After:
image

### DIFF
--- a/app/views/developers/index.html.erb
+++ b/app/views/developers/index.html.erb
@@ -6,9 +6,6 @@
   <h3 class="text-lg leading-6 font-medium text-gray-900">
     Ruby on Rails developers
   </h3>
-  <div class="mt-3 sm:mt-0 sm:ml-4">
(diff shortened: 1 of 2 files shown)


joemasilotti:
Thanks for calling this out. There have been a few takes in this in previous PRs. 

Thinking on it some more, I'd like the approach to be:

1. If you don't have a profile yet, show the CTA like it currently does
2. If you already have a profile, then don't show the additional CTAs outside of the page header

caseyjenks:
👍

joemasilotti:
Thanks for the PR, @caseyjenks. The inline buttons are being removed in a future PR so I'm going to cherry-pick your commit into that.

caseyjenks:
Sounds good!

PR #100, opened 2021-11-15
Introduce length validation on developer.hero column

caseyjenks:
## Purpose
When setting up my profile today, I noticed the hero text I entered got cut off on the profile show page. It seems the character length before text starts getting truncated is 65. It would be better if we let the user know they're trying to input a hero that is too long before they save and get taken back to the profile show page

## Approach
Adds a maximum length validation of 65 on `hero` to the `developer` model
Adds a maxlength of 65 to the HTML `input` element

### DIFF
--- a/test/models/developer_test.rb
+++ b/test/models/developer_test.rb
@@ -71,6 +71,13 @@ class DeveloperTest < ActiveSupport::TestCase
     assert_not_nil developer.errors[:bio]
   end
 
+  test "invalid if hero is too long" do
+    @developer.hero = "This is hero text is entirely too long for this field and will be invalid"
+
+    refute @developer.valid?
+    assert_equal @developer.errors[:hero], ["is too long (maximum is 65 characters)"]
+  end
+
   test "available scope is only available developers" do
     travel_to Time.zone.local(2021, 5, 4)
 
--- a/app/models/developer.rb
+++ b/app/models/developer.rb
@@ -6,7 +6,7 @@ class Developer < ApplicationRecord
   has_one_attached :cover_image
 
   validates :name, presence: true
-  validates :hero, presence: true
+  validates :hero, presence: true, length: {maximum: 65}
   validates :bio, presence: true
   validates :avatar, content_type: ["image/png", "image/jpg", "image/jpeg"],
     max_file_size: 2.megabytes
--- a/app/views/developers/_form.html.erb
+++ b/app/views/developers/_form.html.erb
@@ -22,7 +22,7 @@
             <div class="sm:col-span-6">
               <%= form.label :hero, class: "block text-sm font-medium text-gray-700" %>
               <div class="mt-1">
-                <%= form.text_field :hero, class: field_classes(form, :hero) %>
+                <%= form.text_field :hero, class: field_classes(form, :hero), maxlength: 65 %>
               </div>
               <p class="mt-2 text-sm text-gray-500">Summarize yourself as a developer in a few words.</p>
             </div>


test/models/developer_test.rb
```diff
@@ -71,6 +71,13 @@ class DeveloperTest < ActiveSupport::TestCase
     assert_not_nil developer.errors[:bio]
   end
 
+  test "invalid if hero is too long" do
+    @developer.hero = "This is hero text is entirely too long for this field and will be invalid"
+
+    refute @developer.valid?
+    assert_not_nil @developer.errors[:hero]
```
vamsipavanmahesh, line 78: may be assert the error message itself, so if it's looking too odd by some chance, we can customize it

joemasilotti:
Oh! Thanks for bringing this up. On mobile, the truncation actually occurs much earlier.

image

I think we should actually drop the `truncate` CSS and let this wrap across multiple lines. Along with that, having a length restriction for both the hero and bio sounds like good ideas.

---

I'm not 100% sure what the cutoff should be yet. Here's a breakdown of existing developer profile lengths to help us make a decision.

#### Hero

```ruby
["0-10 => 2",
 "10-20 => 14",
 "20-30 => 19",
 "30-40 => 10",
 "40-50 => 9",
 "50-60 => 4",
 "60-70 => 2",
 "70-80 => 2",
 "90-100 => 1"]
```

#### Bio

```ruby
["0-50 => 8",
 "50-100 => 9",
 "100-150 => 14",
 "150-200 => 4",
 "200-250 => 8",
 "250-300 => 6",
 "300-350 => 4",
 "350-400 => 5",
 "550-600 => 1",
 "600-650 => 1",
 "750-800 => 2",
 "850-900 => 1"]
```

Another outstanding questions is what to do with existing profiles that violate these validations.

test/models/developer_test.rb
caseyjenks: @vamsipavanmahesh updated: https://github.com/joemasilotti/railsdevs.io/pull/100/files#diff-744524efd81cc7981472e7b3ce18a95469a168d333dd97435f19371a9d20036cR78

joemasilotti:
Yeah, let's drop `truncate` in one PR then come back to this one.

joemasilotti:
Instead of removing the `truncate`, it might be better to set up line clamping. Tailwind CSS has an [official plugin for it](https://github.com/tailwindlabs/tailwindcss-line-clamp), too. We could try 2 lines on mobile and 1 line on tablet/desktop.

caseyjenks:
> Another outstanding questions is what to do with existing profiles that violate these validations.

We could run a script in the console that truncates the hero's to the character limit. I'd like to propose `65` as the cutoff since that's how many characters you can fit at the widest width. Thoughts?

caseyjenks:
@joemasilotti Actually, I don't think truncating or line-clamping is necessary since it was handled nicely in another PR that introduced the 2 column layout. I suppose the only thing left to do on this PR is decide on a limit.

joemasilotti:
I've been thinking on this one a bit. If the text is cut off after a few lines, do we really need a character length validation?

I ask because if we end up changing designs in the future it would stink to have potentially "lost" data on a migration. Or unnecessarily restricted developers from adding more info.

caseyjenks:
Totally your call. I can see either way working fine

joemasilotti:
Thanks again for the PR @caseyjenks. I'm going to close this for now because I'm not yet sure where the design of these pages will end up in the future.

Issue #70, opened 2021-11-10
Developer profiles with name


thubamamba:
This might be a good one as there's already thought on SEO. Perhaps profile links shouldn't use developers ID but their name. 

I'd be happy to jump on this @joemasilotti.

davegudge:
I've thought about this one too. 

My initial thought was to use the GitHub username, however, it's not mandatory and there is no way to ensure people are 'claiming' their own username. 

Also, the plan is to hide the contact details, so using the GitHub username won't work. I'm sure I read that the developer name was also going to be hidden, so exposing them in the url goes against that? 

An alternative would be to generate a uuid for each developer and use the uuid in the url, but uuid's are long!

Perhaps a randomly generated string, 8 or 10 chars in length, would do the trick for now? This approach does lose the SEO aspect, but would be better than using ids in the url.

joemasilotti:
At this point I'm not planning on exposing a developer's name in the HTML nor the URL string. I agree that it would be better for SEO, but I'm happy to have numbers for now.

Are there any benefits to UUIDs besides hiding how many developers there are? Or the order someone signed up in? What's so bad about using IDs in the URL?

davegudge:
UUIDs are beneficial if you ever need to interface with an external system. 

There's no issue with using the ids in the URL. As you have mentioned, it simply exposes your developer count and the order in which people signed up.

joemasilotti:
Got it. Well, until we interface with an external system I think IDs should be fine. Thanks for the feedback!

PR #101, opened 2021-11-15
Adds slug to developers table for prettier URLs

caseyjenks:
## Purpose
This PR adds slugs to the developers table & URLs that are based on the `name` column. This makes URL's a little more human friendly, so the user sees something like `/developers/casey-jenks` instead of `/developers/231`.

## Approach
The `friendly_id` gem made this pretty trivial. 

## Open Questions
We need to populate the `slug` field for all existing developers. I added this into the migration and figure we could probably get away with it since the table is still so tiny:

```
Developer.find_each(&:save)
```

If you would rather this be pulled into a rake task, or even just run this from the console post deployment, let me know and we can adjust.

## Screenshots
_Before_:
image

_After_:
image

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -6,6 +6,7 @@ ruby "3.0.2"
 gem "rails", github: "rails/rails", branch: "main"
 
 gem "cssbundling-rails", "~> 0.2"
+gem "friendly_id", "~> 5.4.0"
 gem "hotwire-rails", "~> 0.1"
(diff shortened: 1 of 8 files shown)


joemasilotti:
Thanks for the PR!

I'm still trying to keep developer names' anonymous (see #70). But if that changes then we can definitely revisit.

Issue #21, opened 2021-11-01
Add tests for avatar and cover photo uploads


jacobdaddario:
We need to determine how we're going to go about testing the avatar uploads in #20. This breaks down into basically two items.

1. Are we okay with using shoulda matchers? This is the recommended way to test file attachments according to the `active_storage_validations` gem. I've never personally used them.
2. How do we handle file fixtures? I need to do some investigation to figure this out, but I wanted to put it out there in case anyone else knew.

jacobdaddario:
Going to go ahead and handle this now that #60 is ready to be merged.

joemasilotti:
Hey @jacobdaddario, this has been open for a month now. I'm going to close it to clear out old issues.

Feel free to open it up again or open a PR if you get around to working on it.

jacobdaddario:
😱 In the craziness of the holidays and fresh-installing my Mac I completely forgot about this. I'm re-cloning my fork and pulling down all of the changes now. Sorry about that! I had been kind of proud of staying on top of stuff for this project, and I'm embarrassed I forgot about this.

joemasilotti:
No worries at all! Enjoy the holidays and get to this whenever. No rush - it all seems to be working fine in production now.

PR #102, opened 2021-11-16
feature: Users can create conversations.

jacobdaddario:
I'm going to take care of #21 soon, but I know that how we want to handle messaging is going to require a lot of discussion. I had some ideas about the data model, so I wanted to get some of my ideas down in code so that we can start discussing things now instead of later.

I had a lot of trouble naming some of the relationships, and I'm still trying to completely determine how the UI would look. Any input here would be appreciated. Right now I'm particularly struggling with how we differentiate between `Client` users and `Developer` users.

### DIFF
--- a/app/components/conversation_component.html.erb
+++ b/app/components/conversation_component.html.erb
@@ -1,0 +1,6 @@
+<li class="py-6 flex">
+  <%= image_tag recipient_avatar, alt: "Recipient avatar", class: "h-10 w-10 rounded-full" %>
+  <div class="ml-3">
+    <%= content_tag :p, recipient_header, class: "text-sm font-medium text-gray-900" %>
+  </div>
+</li>
--- a/app/components/conversation_component.rb
+++ b/app/components/conversation_component.rb
@@ -1,0 +1,18 @@
+class ConversationComponent < ViewComponent::Base
+  def initialize(conversation:, user:)
+    @conversation = conversation
+    @recipient_user = user.company? ? conversation.developer : conversation.client
+  end
+
+  def recipient_avatar
+    begin
+      @recipient_user.avatar
+    rescue
+      "joe.jpg"
+    end
+  end
+
+  def recipient_header
+    @recipient_user.company? ? @recipient_user.email : @recipient_user.developer.hero
+  end
+end
--- a/app/controllers/conversations_controller.rb
+++ b/app/controllers/conversations_controller.rb
@@ -1,0 +1,7 @@
+class ConversationsController < ApplicationController
+  def show
+  end
+
+  def index
+  end
+end
--- a/app/controllers/conversations_controller.rb
+++ b/app/controllers/conversations_controller.rb
@@ -1,7 +1,10 @@
 class ConversationsController < ApplicationController
+  before_action :authenticate_user!
+
   def show
   end
 
   def index
+    @conversations = policy_scope(Conversation)
   end
 end
--- a/app/helpers/conversations_helper.rb
+++ b/app/helpers/conversations_helper.rb
@@ -1,0 +1,2 @@
+module ConversationsHelper
+end
--- a/app/models/conversation.rb
+++ b/app/models/conversation.rb
@@ -1,4 +1,6 @@
 class Conversation < ApplicationRecord
-  belongs_to :developer
-  belongs_to :client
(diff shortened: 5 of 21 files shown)


jacobdaddario:
Just a note for myself after doing some more thinking.

I'm considering changing `client_conversations` to be `leads` and `developer_conversations` to be `prospects`. I'm not 100% on these names, but I feel like they fit the domain and that they'll be helpful with the UI design.

joemasilotti:
Thanks for kicking this off! This is probably the next "big" feature for railsdevs, but I want to make sure #86 is live before I dive into this too deeply.

That said, for the modeling, I think you have some solid ideas. Here's how I've sketched it out before.

* A `Conversation` has many `Message`s
    * Messages belong to a conversation, developers have many messages through conversations
* A `Conversation` belongs to a `Developer` and a `Company`
    * I want a better name for `Company`, but they are essentially the "hirer"
* Developers can't message other developers (for now)
* Only companies can initiate a conversation with a developer
* That leads to a `ConversationsController` and a `MessagesController`
    * Former is where the viewing of messages occurs
    * Latter is where the POSTing of messages occurs
* "Multi-column inbox" from Tailwind UI is the (more complicated) direction I'm thinking design-wise
* This will flesh out when you create an account saying you are a developer or a company
    * This could be started with a boolean flag (radio) on the sign up form

On top of that, there's moderation. We don't need this for this PR, but I won't launch messaging until this is in place.

* `ModerationController` is nested under `MessagesController` to allow POSTing to flag a message
* This creates a `Moderation` record, belonging to a `Message`
* Moderations are only visible to admins via `Admin::ModerationsController`
    * There's a lot to do here, but just being able to see the flagged content is a great start

jacobdaddario:
Hey! I'm just poking around GitHub while finishing eating breakfast, and I need to get back to my work. That said, I just wanted to drop a super quick reply.

I feel that what you've sketched out roughly matches the conceptual model I have in my head as well. I ran into a few stumbling blocks around handling some of the domain logic. If you're available to chat today I'd be more than happy to offer some insight on what I've found so far.

joemasilotti:
Awesome! If possible, I'd love to keep as much conversation on GitHub. Then everything is kept public and anyone can chime in who is reading.

jacobdaddario:
@joemasilotti I think that I'm going to go ahead and close this PR for now, but I wanted to quickly write a summary post while eating breakfast.

So, the data model essentially played out the way you described.
* Conversation belongs to `developer` and `client`
* Clients have many `hiring_leads`
* Developers have many `work_leads`
* Conversations have index and show views
* The index and show views are combined into a single UI
Screen Shot 2021-11-19 at 9 52 41 AM

I didn't ultimately need separate controllers for hiring leads vs. work leads though. Instead I used Pundit scopes to select the appropriate conversations for each type of user.

From there the plan was the following:
* Conversation has many `messages`
* Clients could only message developers and vice versa
* Conversations could be created in the show action using `find_or_create_by`

Ultimately I want to put a pin in this for 3 major reasons.
1. In this implementation I attached conversations to users rather than the developer model. I'm not sure if this is the best approach, and I want to take some time to thoughtfully discuss with you what we consider `Developer` to be. As of now, it kind of seems to represent more of a developer profile to me, but I can also see the argument that it represents a role of `User` that's composed into the `User` class using the has_one relationship.
2. We need to figure out how we want to model companies before getting the behavior for conversations down solid. This goes hand in hand with point 1. For instance, if we want to go with the composable approach mentioned in 1, we might need a new registration flow. Additionally, I think we might have to consider moving avatar into the `User` class since that would need to be a shared behavior. Either way, in order to get `Conversation` solid, we'll need the Company API to be solid so that  we can establish the `Conversant` ducktype or role solid.
3. Even this one PR which was meant to be just a portion of the behavior, is like three or four actual feature PRs in one. We need to make a decision regarding whether we want to pull this out into a branch outside of master while working on it, or if we want to do feature flags.

I hope this was insightful. If you have any questions I'll try and answer when I'm available. Things will be pretty spotting this week with Thanksgiving coming up though.

joemasilotti:
Thanks so much for this spike @jacobdaddario! I agree with your third point – this is way more work than a single PR. It will most likely be built up slowly and perhaps a lot behind feature flags to start.

I'm hoping to break this down into PR-sized chunks over the next few days. And I'll be referencing your PR for guidance.

PR #103, opened 2021-11-16
feat: Componentize the dev card and truncate bio

jacobdaddario:
Hey! I saw the discussion about truncating developer bios in another issue, so I wanted to quickly bang something out while eating breakfast. I have to get back to work now, but I wanted to make a PR since I thought you might want to take a look at the seed data. I wanted to include something to allow me to test the pagination behavior, but I feel that what I currently have is inelegant.

This evening I will come back to write some tests, clean up the code, and make any changes you'd like to make.


joemasilotti:
After a quick glance at the code I have one request and one question.

1. Please split this out into 2 PRs. I'm not sure where the concatenation/truncation will end up and don't want that to hold up the View Component extraction.
2. Are there any performance benefits or possibilities with rendering a collection of View Components? Similar to how partials you can render collections via `render @developers`.

joemasilotti:
Instead of removing the `truncate`, it might be better to set up line clamping. Tailwind CSS has an [official plugin for it](https://github.com/tailwindlabs/tailwindcss-line-clamp), too. We could try 2 lines on mobile and 1 line on tablet/desktop.

jacobdaddario:
Okay! I want to clean this PR up today since I didn't get to it yesterday. I'll have a look at that.

jacobdaddario:
Added tests and made requested changes while eating breakfast.

app/components/dev_card_component.html.erb
```diff
@@ -0,0 +1,23 @@
+<li>
```
joemasilotti, line 1: Let's move this tag to the index. The component might not always be rendered as a list item.

app/components/dev_card_component.rb
```diff
@@ -0,0 +1,13 @@
+class DevCardComponent < ApplicationComponent
```
joemasilotti, line 1: Let's rename this to `DeveloperCardComponent`.

app/components/dev_card_component.rb
```diff
@@ -0,0 +1,13 @@
...
+  def initialize(developer:)
+    @developer = developer
+  end
+
+  def hero
```
joemasilotti, line 6: Oh, nice use of helpers! Much better than calling them directly in the view.

app/views/developers/index.html.erb
```diff
@@ -23,7 +25,11 @@
 
 <%= turbo_frame_tag "developers-#{@pagy.page}" do %>
   <%= turbo_stream.append "developers" do %>
-    <%= render @developers unless @pagy.page == 1%>
+    <% unless @pagy.page == 1 %>
```
joemasilotti, line 28: See? This one isn't getting rendered in a list.

joemasilotti (changes requested):
I requested a few small tweaks. Other than that this looks good!

joemasilotti:
Ohhh, we probably want to use the collection rendering here, too!

https://viewcomponent.org/guide/collections.html

jacobdaddario:
About to sit down and code. Thanks for pointing that out!

jacobdaddario:
The requested changes have been made and committed. Only problem I ran into was how to include the `<li>` tags with the `with_collection` method. I couldn't find any hooks that would allow us to conditionally add the list tags only if we were rendering a collectioj.

jacobdaddario:
How would you want to move forwards with handling seed data in the future? I added it in just to browser test the pagination behavior, but I had a feeling you'd prefer to not keep it. Is it generally acceptable to add seeding data? Is there a standard you'd like it to adhere to?

joemasilotti:
Excellent! `with_collection` is awesome, glad we stumbled across that.

I removed the 20 developer seeds. When someone needs to test pagination they can manually do it (or add this back in). In the future, we will probably have more robust seeds that mean things, anyway (e.g. an available developer, one that is looking for $200k, one that has the minimum filled in, etc.)

I also dropped the `<ul>`. I'm not sure how much of a list this is anymore, so I don't think we need the semantic part.

Thanks for the PR!

joemasilotti (approved):

joemasilotti:
> Is it generally acceptable to add seeding data? Is there a standard you'd like it to adhere to?

Great questions. I don't have any specific rules, but 20 (very) random developer profiles doesn't look very nice when you're staring at it for hours. :)

Maybe a rake task that's just for testing pagination could work!

jacobdaddario:
> > Is it generally acceptable to add seeding data? Is there a standard you'd like it to adhere to?
> 
> Great questions. I don't have any specific rules, but 20 (very) random developer profiles doesn't look very nice when you're staring at it for hours. :)
> 
> Maybe a rake task that's just for testing pagination could work!

Makes sense! Yeah I figured you'd be unhappy with that, but I wanted to leave it in to spark some discussion. My go-to solution would be to [add Faker](https://github.com/faker-ruby/faker). I omitted that here because I felt it'd be more appropriate for a separate PR. Would you like me install that and open one?

joemasilotti:
No, not yet. I'd rather think more critically about our seeds and explicitly create different combinations of developer profile settings. To test pagination, we could always change the page size in development to something smaller.

PR #108, opened 2021-11-18
Add expected compensation developer

peterberkenbosch:

If this PR is not ready for review, please make sure to submit it as a draft.

Updated todo list:

- [x] Change expected salary to minimum and maximum
- [x] Change expected hourly rate to minimum and maximum
- [x] Rename columns to "preferred"
- [x] Rename view components to "preferred"
- [x] Round the currency to 0 precision
- [x] Add "k" for salary, rounded
- [x] Show the salary ranges and hourly ranges
- [ ] Add the $ icon
- [x] Add the new columns to the form
- [ ]  (maybe) Shrink the form inputs for these numbers
- [x] Replace "USD" with "$" in the form


joemasilotti:
Quick glance (I haven't run the code yet, just viewing the diff on my phone) and it's looking good!

Two questions/suggestions:

1. We don't need the Money gem. Let's accept whole integers only for salary and rate and store that directly in the database. Cents is too granular for numbers like this. 
2. I'm not sure if I like "expected" - "desired" might be a better word. But am open to your thoughts on that.

peterberkenbosch:
thanks @joemasilotti I'll update the PR with removing the money gem and just using whole numbers.   

naming is hard... 😅

joemasilotti:
Hey @peterberkenbosch, is this PR ready for review? 

After a few nights sleep... lets use "preferred" – that feels like the closest to what the field actually means.

peterberkenbosch:
@joemasilotti I'm rebasing upstream, fix conflicts and add viewcomponents this morning.

peterberkenbosch:
@joemasilotti qq, I see 2 approaches for initializing components, with keyword arguments and without. What is your preference here? 

Will rebase and rewrite the commits to be more atomic if this is the approach you like. Will mark it ready for review now.

peterberkenbosch:
@joemasilotti Thanks! Makes sense, I reverted the keyword arg commit so it matches. Ready for review

joemasilotti:
Thanks for the PR! Now that I'm seeing this "live" I have a lot more thoughts on how we can handle this. :D

First, I'm backtracking my decision on ranges. Saying you want/desire/expect $150k means that an employer might skip over you if their role is for $125k (or $175k!). But if you said your range was $125k-$150k then that employer knows everyone is on the same page.

Once we have the ranges, let's round them to fewer significant figures, add an icon, and remove the labels. For example...

image

Then, let's rename everything mentioning "expected" to "preferred." I think that better captures what we are saying here.

In the form, let's make the input field much smaller. There's an example in the "Available to start on" input. Also, let's change the copy to "What range of compensation do you prefer for a role?" And we can replace "USD" with "$".

---

Here's a checklist of everything I mentioned. Feel free to punt this back to me if you don't want to take care of all of these. I know they are a little nit-picky!

- [ ] Change expected salary to minimum and maximum
- [ ] Change expected hourly rate to minimum and maximum
- [ ] Rename columns to "preferred"
- [ ] Rename view components to "preferred"
- [ ] Round the currency to 0 precision
- [ ] Add "k" for salary, rounded
- [ ] Show the salary ranges and hourly ranges
- [ ] Add the $ icon
- [ ] Add the new columns to the form
- [ ] (maybe) Shrink the form inputs for these numbers
- [ ] Replace "USD" with "$" in the form

peterberkenbosch:
@joemasilotti on it! btw not nit-picky at all

joemasilotti:
Great! Thank you.

peterberkenbosch:
pushed most of the changes, could use some help with styling on the inputs, smaller and min - max on the same line and the $ icon. // @joemasilotti

peterberkenbosch:
@joemasilotti still WIP, few little things around validations and some failing tests to be fixed. Looking at wrapping this up tomorrow.

peterberkenbosch:
getting there though: https://cln.sh/0rbNp2

app/components/compensation_amount_component.html.erb
```diff
@@ -0,0 +1,9 @@
+<div class="sm:col-span-1">
+  <dt class="text-sm font-medium text-gray-500">
+    <%= title %>
+  </dt>
+  <dd class="flex mt-1 text-sm text-gray-900">
+    $<%= number_to_human(min_amount, format: "%n%u", units: {thousand: "K"}) %> -
```
joemasilotti, line 6: You can toss that `$` inside the `format` param.

```
number_to_human(min_amount, format: "$%n%u", units: {thousand: "K"})
```

Another approach is the following. Not sure which is more readable, though.

```
number_to_currency(min_amount/1000, precision: 0, format: "%u%nk")
```

app/components/compensation_amount_component.rb
```diff
@@ -0,0 +1,14 @@
...
+
+  def initialize(min_amount:, max_amount:, suffix:, title:)
+    @min_amount = min_amount
+    @max_amount = max_amount
+    @suffix = suffix
+    @title = title
+  end
+
+  def render?
+    @min_amount.present?
```
joemasilotti, line 12: What if someone only adds the max?

app/components/preferred_compensation_component.rb
```diff
@@ -0,0 +1,11 @@
+class PreferredCompensationComponent < ApplicationComponent
+  attr_reader :developer
+
+  def initialize(developer)
+    @developer = developer
+  end
+
+  def render?
+    @developer.preferred_min_hourly_rate.present? || @developer.preferred_min_salary.present?
```
joemasilotti, line 9:
```suggestion
    developer.preferred_min_hourly_rate.present? || developer.preferred_min_salary.present?
```

app/components/compensation_amount_component.rb
joemasilotti, line 12:
```suggestion
    min_amount.present?
```

app/models/developer.rb
```diff
@@ -22,6 +22,11 @@ class Developer < ApplicationRecord
   validates :cover_image, content_type: ["image/png", "image/jpg", "image/jpeg", "image/gif"],
     max_file_size: 10.megabytes
 
+  validates_numericality_of :preferred_min_hourly_rate, less_than: ->(developer) { developer.preferred_max_hourly_rate }, if: -> { preferred_min_hourly_rate.present? }
+  validates_numericality_of :preferred_max_hourly_rate, greater_than: ->(developer) { developer.preferred_min_hourly_rate }, if: -> { preferred_max_hourly_rate.present? }
+  validates_numericality_of :preferred_min_salary, less_than: ->(developer) { developer.preferred_max_salary }, if: -> { preferred_min_salary.present? }
+  validates_numericality_of :preferred_max_salary, greater_than: ->(developer) { developer.preferred_min_salary }, if: -> { preferred_max_salary.present? }
```
joemasilotti, line 28: Let's change all of these to `_or_equal_to`, just in case someone _really_ has a specific rate/salary in mind.

app/views/developers/_form.html.erb
```diff
@@ -162,6 +162,58 @@
...
+            <div class="sm:col-span-3">
+              <%= form.label :preferred_max_hourly_rate, class: "block text-sm font-medium text-gray-700" %>
+              <div class="mt-1 flex rounded-md shadow-sm">
+                <span class="inline-flex items-center px-3 rounded-l-md border border-r-0 border-gray-300 bg-gray-50 text-gray-500 sm:text-sm">
+                  $
+                </span>
+                <%= form.number_field :preferred_max_hourly_rate, class: "flex-1 focus:ring-gray-500 focus:border-gray-500 block w-full min-w-0 rounded-none rounded-r-md sm:text-sm border-gray-300" %>
+              </div>
+            </div>
+
```
joemasilotti, line 211:
```suggestion
```

db/migrate/20211123083446_add_preferred_max_hourly_rate_to_developers.rb
```diff
@@ -0,0 +1,5 @@
+class AddPreferredMaxHourlyRateToDevelopers < ActiveRecord::Migration[7.0]
+  def change
+    add_column :developers, :preferred_max_hourly_rate, :integer
+  end
+end
```
joemasilotti, line 5: Let's combine all 4 migrations into a single one.

test/components/compensation_amount_component_test.rb
```diff
@@ -0,0 +1,14 @@
+require "test_helper"
+
+class CompensationAmountComponentTest < ViewComponent::TestCase
+  test "doesn't render with no amount" do
```
joemasilotti, line 4:
```suggestion
  test "doesn't render without an amount" do
```

test/components/preferred_compensation_component_test.rb
```diff
@@ -0,0 +1,26 @@
+require "test_helper"
+
+class PreferredCompensationComponentTest < ViewComponent::TestCase
+  setup do
+    @developer = developers(:available)
+  end
+
+  test "renders nothing whith no expected compensation" do
```
joemasilotti, line 8:
```suggestion
  test "doesn't render without an expected compensation" do
```

app/models/developer.rb
```diff
@@ -22,6 +22,11 @@ class Developer < ApplicationRecord
   validates :cover_image, content_type: ["image/png", "image/jpg", "image/jpeg", "image/gif"],
     max_file_size: 10.megabytes
 
+  validates_numericality_of :preferred_min_hourly_rate, less_than: ->(developer) { developer.preferred_max_hourly_rate }, if: -> { preferred_min_hourly_rate.present? }
```
joemasilotti, line 25: If you add a min but no max hourly or salary an exception is raised.

```
web    | TypeError (can't convert nil into Float):
web    |
web    | app/controllers/developers_controller.rb:38:in `update'
```

I _think_ this is the line, but not 100% sure.

joemasilotti (commented):
Looking good! This is getting close. Here are a couple of things to get ahead on while you're still deep in the code (see review comments).

Design-wise, let's combine the two compensation blocks a little bit more.
image

app/components/compensation_amount_component.html.erb
peterberkenbosch: Fixed this, you can actually to `number_to_currency(number_to_human` :)

peterberkenbosch:
@joemasilotti next round has been committed, removed the validation around min and max amounts, let me know if you would like those back in.

joemasilotti (approved):

joemasilotti:
Thanks again for this! I made a few tweaks but other than that it looks great. Going to merge in now.

### Changes

#### Validations

I tweaked the validations to make things a bit less busy. Now, the max values are only validated against min, if both are present. I think this cleans things up a tad.

image

image

#### Migrations

I moved the migration after the most recent on `main` so it applies correctly.

#### Tests

I added tests for the new validations on the `Developer` model.

#### Copy

I made a few minor copy updates to form and developer profile.

peterberkenbosch:
Thanks @joemasilotti, fun project.

Issue #110, opened 2021-11-18
Move all copy to translation (locales/en.yml) file


joemasilotti:
**Describe the bug**

Not really a bug, but more a chore. We should move all of the copy to the i18n file `config/locales/en.yml`. This will make our lives much easier if/when we translate to a new language. It also keeps everything in one spot vs. littered across the app.

fetijashari:
Hey @joemasilotti - are you thinking to move all "static text" that currently exist on controllers, views,  components etc - to I18n file (en.yml) ?

joemasilotti:
Yep, exactly!

fetijashari:
Ok - I'll take it :)

PR #118, opened 2021-11-23
Move all copy to translation (locales/en.yml) file #110

fetijashari:

If this PR is not ready for review, please make sure to submit it as a draft.


app/assets/stylesheets/components/buttons.css
```diff
@@ -0,0 +1,13 @@
...
+
+    &:hover {
+      @apply bg-gray-700
+    }
+
+    &:focus {
+      @apply outline-none ring-2 ring-offset-2 ring-gray-500;
+    }
+  }
+}
```
joemasilotti, line 13: Let's move this to a new PR.

bin/bundle
```diff
@@ -28,7 +28,7 @@ m = Module.new do
       if update_index && update_index.succ == i && a =~ Gem::Version::ANCHORED_VERSION_PATTERN
         bundler_version = a
       end
-      next unless a =~ /\A--bundler(?:[= ](#{Gem::Version::VERSION_PATTERN}))?\z/
+      next unless a =~ /\A--bundler(?:[= ](#{Gem::Version::VERSION_PATTERN}))?\z/o
```
joemasilotti, line 31: What are these "o" characters at the end?

bin/bundle
```diff
@@ -54,14 +54,14 @@ m = Module.new do
   def lockfile_version
     return unless File.file?(lockfile)
     lockfile_contents = File.read(lockfile)
-    return unless lockfile_contents =~ /\n\nBUNDLED WITH\n\s{2,}(#{Gem::Version::VERSION_PATTERN})\n/
+    return unless lockfile_contents =~ /\n\nBUNDLED WITH\n\s{2,}(#{Gem::Version::VERSION_PATTERN})\n/o
```
joemasilotti, line 57: Same question here.

joemasilotti:
Thanks for the PR! I gave this a quick glance and wanted to get some things straightened out before I dive in to every change.

Can you please...
- [ ] Replace all strings with double quotes
- [ ] Extract the CSS changes to a new PR
- [ ] Answer the questions about the trailing "o" character

Once those are done I'll take a deep dive into each extracted translation.

bin/bundle
fetijashari: Hey @joemasilotti - it came as a suggestion from Standardrb.

This was the message on terminal: ```bin/bundle:31:24: Performance/ConstantRegexp: Extract this regexp into a constant, memoize it, or append an `/o` option to its options.```

bin/bundle
fetijashari: @joemasilotti - same with this, a suggestion from standardrb

```bin/bundle:57:40: Performance/ConstantRegexp: Extract this regexp into a constant, memoize it, or append an `/o` option to its options.```

**Comment by joemasilotti** [comment]
This is looking great, thanks for the quick turnaround! A few more thoughts...

First, l started googling i18n naming conventions and stumbled across [this article](https://phrase.com/blog/posts/ruby-lessons-learned-naming-and-managing-rails-i18n-keys/). There's a lot of hand-wavy stuff in there, but I think it provides some solid ideas for how to structure our translations.

Installing [`i18n-tasks`](https://github.com/glebm/i18n-tasks) in dev and test got me a lot of mileage in terms of unused translations and cleaning up the locale files. I'm leaning towards installing that gem, adding the minitest test, and running the `normalize` task. Perhaps even as part of CI or, to start, added in the PR template.

I'm torn on how they recommend explicit over implicit keys. For example, they recommend using `general.create_profile` instead of `.create` (inside of `app/views/developers/form`). I like that this would remove duplication, something I've always hated with the translation files. And it helps when we refactor and rename files. But I don't like that it effectively ignores the ["lazy" lookup that Rails provides](https://guides.rubyonrails.org/i18n.html#lazy-lookup) with view translations.

Minor, but it looks like there's some `_html` translations that are only plain text.

---

I'm going to think about this a bit more and get back to you. I really appreciate the work you've done! I want to make sure this is easy to maintain in the future and very obvious for everyone contributing to the project.

**Comment by joemasilotti** [comment]
I've decided that sticking with standard Rails is the best approach here, which is exactly what you did. I'd love to merge this in.

@fetijashari, do you still have the appetite to address the feedback I left in my last comment (the gem and `_html` stuff)? It also looks like it is behind `main` a bit and there are conflicts.

**Comment by fetijashari** [comment]
Hey @joemasilotti - Sure thing !

I'll get it done later today.

File: config/i18n-tasks.yml [hunk 4]
```diff
@@ -0,0 +1,143 @@
+# i18n-tasks finds and manages missing and unused translations: https://github.com/glebm/i18n-tasks
```
joemasilotti, line 1: Let's add the `bundle exec i18n-tasks health` task to the PR checklist and CI.

joemasilotti: Also, run `bundle exec i18n-tasks normalize` to clean up the translation file and commit.

joemasilotti (changes requested):
This is looking great! Thanks for applying all of these new changes. I made a few code formatting requests, a bigger request regarding the structure of the YAML file, and a CI/PR template request.

Other than those this should be ready for merging!

test/components/availability_component_test.rb
```diff
@@ -1,6 +1,7 @@
 require "test_helper"
 
 class AvailabilityComponentTest < ViewComponent::TestCase
+  include ActionView::Helpers::DateHelper
```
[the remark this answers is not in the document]
fetijashari: @joemasilotti I have removed this include, as it is not needed (I tried somethings and forgot to remove)

config/i18n-tasks.yml
fetijashari: > Let's add the bundle exec i18n-tasks health task to the PR checklist and CI. @joemasilotti - I guess I need to move this gem to developemnt and test group then ?

Gemfile
```diff
@@ -22,6 +22,7 @@ end
 group :development do
   gem "erb_lint", require: false
   gem "listen", "~> 3.3"
+  gem "i18n-tasks", "~> 0.9.35"
```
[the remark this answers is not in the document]
fetijashari: @joemasilotti - since we need to run `bundle exec i18n-tasks health` I will move this gem to development and group and sort it :)

config/i18n-tasks.yml
joemasilotti: Ah, yeah, that's correct.

fetijashari:
Hey @joemasilotti  - all Done here. 

`i18n-tasks normalize` triggered a lot of changes - basically, it sorted all the keys on both files `en` and `devise.en`

app/components/open_graph_tags_component.rb
```diff
@@ -9,7 +9,7 @@ def title
   end
 
   def description
-    @description || t(".default_description")
+    @description || t("open_graph_tags_component.default_description")
```
joemasilotti, line 12: Is this change needed?

config/i18n-tasks.yml
```diff
@@ -15,19 +15,23 @@ data:
...
     # - config/locales/**/*.%{locale}.yml
 
   # Locale files to write new keys to, based on a list of key pattern => file rules. Matched from top to bottom:
   # `i18n-tasks normalize -p` will force move the keys according to these rules
-  write:
+  # write:
     ## For example, write devise and simple form keys to their respective files:
     # - ['{devise, simple_form}.*', 'config/locales/\1.%{locale}.yml']
     ## Catch-all default:
     # - config/locales/%{locale}.yml
```
joemasilotti: Let's get rid of these comments if we are using the `write:` key below.

config/i18n-tasks.yml
```diff
@@ -107,8 +111,12 @@ search:
 # - '{devise,simple_form}.*'
 
 ## Consider these keys used:
-# ignore_unused:
-# - 'activerecord.attributes.*'
+ignore_unused:
+- 'activerecord.models.*'
+- 'activerecord.attributes.*'
```
joemasilotti, line 116: Why are these two lines needed? What attributes are we not translating?

joemasilotti (changes requested):
Just a few more tiny changes and a question. Getting close!

config/i18n-tasks.yml
fetijashari: @joemasilotti  if you check the view part of it, I have removed the hardcoded text on labels (within form for each attribute) - for some reason they are not recognized by this gem but as per rails guides and best practices we should translate attributes on the active record and everywhere (incl. errors) they will apear the same.

It could be that I am missing some knowledge here on this gem.

app/components/open_graph_tags_component.rb
fetijashari: yes, according to 18n-tasks

config/i18n-tasks.yml
fetijashari: I left is as an example, for later reference - these comments are the defaults.

config/i18n-tasks.yml
joemasilotti: Gotcha, works for me. Thanks!

config/i18n-tasks.yml
joemasilotti: We can always regenerate the file if need be. Let's remove these lines.

fetijashari: ok - np I'll remove it :)

joemasilotti (approved):
Thanks so much for this! I appreciate the back and forth and time you spent.

PR #119, opened 2021-11-23
Lint ERB files

### LINKED ISSUE #67: Interested in linting erb with standardrb?
Love that this project uses standardrb. How about linting your erb files too?

I recently set up a new Rails app with a pretty minimal erblint config using standardrb for embedded ruby code. Happy to contribute PR if desired!

- https://github.com/Shopify/erb-lint

`cat .erb-lint.yml | pbcopy`
👇
```yaml
---
EnableDefaultLinters: true
linters:
  ErbSafety:
    enabled: true

  Rubocop:
    enabled: true
    rubocop_config:
      require: standard
      inherit_gem:
        standard: config/base.yml
      AllCops:
        DisabledByDefault: true
      Style/IdenticalConditionalBranches:
        Enabled: false
      Layout/BlockAlignment:
        Enabled: false
      Layout/HashAlignment:
        Enabled: false
      Layout/IndentationConsistency:
        Enabled: false
      Layout/InitialIndentation:
        Enabled: false
      Layout/IndentationWidth:
        Enabled: false
      Layout/LeadingEmptyLines:
        Enabled: false
      Layout/ParameterAlignment:
        Enabled: false
      Layout/TrailingEmptyLines:
        Enabled: false
      Layout/TrailingWhitespace:
        Enabled: false
      Lint/Void:
        Enabled: false
```

joemasilotti:

---

@pingortle, mind taking a look at this? This closes #67.

### DIFF
--- a/app/components/avatar_component.html.erb
+++ b/app/components/avatar_component.html.erb
@@ -1,1 +1,1 @@
-<%= image_tag @image, alt: "Developer's avatar", data: @data, class: "object-cover h-24 w-24 bg-gray-300 rounded-full ring-4 ring-white sm:h-32 sm:w-32" %>
+<%= image_tag image, alt: "Developer's avatar", data: data, class: "object-cover h-24 w-24 bg-gray-300 rounded-full ring-4 ring-white sm:h-32 sm:w-32" %>
--- a/.erb-lint.yml
+++ b/.erb-lint.yml
@@ -3,6 +3,7 @@ EnableDefaultLinters: true
 
 exclude:
 - "app/views/*_mailer/*"
+- "vendor/bundle/**/*"
 
 linters:
   ErbSafety:
--- a/.erb-lint.yml
+++ b/.erb-lint.yml
@@ -1,0 +1,22 @@
+---
+EnableDefaultLinters: true
+
+exclude:
+- "app/views/*_mailer/*"
+
+linters:
+  ErbSafety:
+    enabled: true
+  PartialInstanceVariable:
+    enabled: true
+
+  Rubocop:
+    enabled: true
+    rubocop_config:
+      require: standard
+      inherit_gem:
+        standard: config/base.yml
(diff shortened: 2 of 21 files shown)


app/components/avatar_component.html.erb
```diff
@@ -1 +1 @@
-<%= image_tag @image, alt: "Developer's avatar", data: @data, class: "object-cover h-24 w-24 bg-gray-300 rounded-full ring-4 ring-white sm:h-32 sm:w-32" %>
+<%= image_tag image, alt: "Developer's avatar", data: data, class: "object-cover h-24 w-24 bg-gray-300 rounded-full ring-4 ring-white sm:h-32 sm:w-32" %>
```
pingortle, line 1: @joemasilotti 

I'm cool with changing ivars to methods, and it looks like you are too since you put in the work! 😄  I don't remember doing this when I set it up in my app, so perhaps I was just a bit more aggressive ignoring rules.

It does keep the code that much more consistent. Net positive there in my book. 👍

pingortle (approved):
Nice! Thanks for doing this!

I'm honestly impressed that it didn't require more tweaks. ✨ 

Erb lint has more html/erb oriented features–many of which I have not delved into. I think this is a win even if just for standardizing the embedded ruby. ⚡

app/components/avatar_component.html.erb
joemasilotti: Agreed! I ran the linter with `--enable-all-linters` so see how much would needed to be done. Then, I slowly added each relevant check to the config file. One that I'd love to add is the linter that checks for missing translations! But I need to merge #110 first.

This particular fix is from the `PartialInstanceVariable` linter.

pingortle: Ah, makes sense. Thanks for sharing the `PartialInstanceVariable` tip so I don't have to search for it. 🙏

PR #129, opened 2021-12-02
Pundit for authorization only

joemasilotti:
In implementing #120, I realized that I was starting to put business logic inside of the Pundit policy classes. These classes should only contain logic related to authorization. This PR moves the business logic into controller `before_action` calls. It also adds some missing tests for `Business` and cleans up the controllers a bit.

### Pull request checklist

- [x] Your code contains tests relevant for code you modified
- [x] All new and existing tests are passing
- [x] You have linted the project with `bundle exec standardrb --fix`
- [x] You have linted the project with `bundle exec erblint --lint-all --autocorrect`

PR #130, opened 2021-12-02
Message a developer

### LINKED ISSUE #120: A business can message a developer
The ~first~ second (of many) [Messaging](https://github.com/joemasilotti/railsdevs.com/projects/3) issues!

A PR addressing this issue should...

* Add a Message button to a developer's profile
    * When clicked, if signed in as a business then open a new conversation screen
    * When clicked, if not signed in as a business then prompt to create a business profile
* `developers/:id/messages/new` with an input field
    * Create a `Conversation` that belongs to the developer and business
    * Create a `Message` that belongs to the conversation
    * Can only be access by a business (via `pundit`)
* `conversations` index of all conversations the developer or business is a part of (read only)
    * Link to conversations from header / menu
* `/developers/:id/conversation` show a specific conversation (read only)
    * `pundit` for authorization scopes
* Hide the Message button in production based on the feature flag
* Double check mobile views

joemasilotti:
This PR closes #120.

### Features from issue

- [x] Add a Message button to a developer's profile
    - [x] When clicked, if signed in as a business then open a new conversation screen
    - [x] When clicked, if not signed in as a business then prompt to create a business profile
- [x] `developers/:id/messages/new` with an input field
    - [x] Create a `Conversation` that belongs to the developer and business
    - [x] Create a `Message` that belongs to the conversation
    - [x] Can only be access by a business (via `pundit`)
- [x] `conversations` index of all conversations the developer or business is a part of (read only)
    - [x] Link to conversations from header / menu
- [x] `/developers/:id/conversation` show a specific conversation (read only)
    - [x] `pundit` for authorization scopes
- [x] Hide the Message button in production based on the feature flag
- [x] Double check mobile views

### Implementation details

I went through a few different approaches to this feature and I like where I ended up. My thought process was broken down into two categories: (ab)using pundit and routes. Here's some decisions I made and why I made them.

#### (Ab)using pundit

First off, we were kind of abusing pundit. See #129 for more details. But in summary, pundit should be used for authorization, only. We were starting to creep business logic in there.

I started to fall for this trap and was adding logic like raising errors if a business already had a profile then catching them and redirecting in the controller. Not good. Instead, I've opted for a few `before_action` methods that check and do the redirections. This could probably be cleaned up, but I think it reads quite nicely and isn't too complicated (yet).

#### Routes

I spent a lot of time figuring out the best routes for this. My first pass was to POST a `Conversation` and nest the `Message` inside. That fell apart because if there was an error you could use context and you had additional authorization checks because the `Conversation` attributes were coming from the form.

Second, I moved to POST a `Message` and have the `Conversation` `#find_or_create_by`. This worked OK, but I ended up needing to wrap actions in `begin` and rolling back which felt ugly.

[Finally](https://github.com/joemasilotti/railsdevs.com/pull/130/commits/48cf3916fba0ca18750968c15fe36ce0878d391a), I decided to POST to `/developers/:id/messages`. This keeps the `Developer` context (it's in the URL!) and grabbing the `Business` from `current_user`. Much harder to fake. I also like this because your URL is no `/developers/:id/conversation` instead of `/conversations/:id`.

Moving forward, this approach _should_ scale better when we introduce developers viewing their own conversations. The controller will need to determine who is the "context" but then things hopefully fall together nicely.

I hope that ramble helped!

### Pull request checklist

### DIFF
--- a/app/components/time_component.html.erb
+++ b/app/components/time_component.html.erb
@@ -1,0 +1,3 @@
+<%= time_tag time do %>
+  <%= time_ago_in_words time %> ago
+<% end %>
--- a/app/components/time_component.rb
+++ b/app/components/time_component.rb
@@ -1,0 +1,7 @@
+class TimeComponent < ApplicationComponent
+  attr_reader :time
+
+  def initialize(time)
+    @time = time
+  end
+end
--- a/app/components/user_menu/signed_in_component.html.erb
+++ b/app/components/user_menu/signed_in_component.html.erb
@@ -1,12 +1,21 @@
-<div data-controller="toggle" data-toggle-close-class="hidden" class="absolute inset-y-0 right-0 flex items-center pr-2 sm:static sm:inset-auto sm:ml-6 sm:pr-0">
+<div class="absolute inset-y-0 right-0 flex items-center pr-2 sm:static sm:inset-auto sm:ml-6 sm:pr-0">
   <% if business? %>
-    <%= link_to new_business_path, class: "bg-gray-800 p-1 rounded-full text-gray-400 hover:text-white focus:outline-none focus:ring-2 focus:ring-offset-2 focus:ring-offset-gray-800 focus:ring-white" do %>
-      <span class="sr-only">Edit business</span>
-      <%= inline_svg_tag "icons/outline/briefcase.svg", class: "h-6 w-6" %>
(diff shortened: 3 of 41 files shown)


joemasilotti:
Closed in favor of #132.

PR #137, opened 2021-12-08
Tidy up word wrap on developer cards

### LINKED ISSUE #134: Tidy up word wrap on developer cards
The developer bio is cut off at weird-looking lines on desktop vs. mobile. The homepage, on desktop, limits the bio to a single line. But 2 lines are visible on mobile.

Also, the word wrap of the developer hero is being wrapped on characters which makes for some heroes hard to read. This should instead break on words.

image

image

danieldpence:
👋🏻 Cool project, thought I'd give this a shot real quick.

I went with max 3 lines of text on mobile and desktop since the cards seemed to have room to allow for 3 lines. You may have plans to add additional data beneath the bio at some point, I don't know.

Closes #134 
Screen Capture on 2021-12-08 at 10-10-41

### Pull request checklist

- [x] You have linted the project with `bundle exec i18n-tasks health`

### DIFF
--- a/app/components/developer_card_component.html.erb
+++ b/app/components/developer_card_component.html.erb
@@ -4,8 +4,8 @@
   </div>
 
   <div class="self-start flex-1">
-    <div class="flex items-baseline break-all">
-      <h2 class="text-lg sm:text-xl font-medium text-gray-900">
+    <div class="flex items-baseline">
+      <h2 class="text-lg sm:text-xl font-medium text-gray-900 break-words">
         <%= hero %>
       </h2>
 
@@ -14,8 +14,8 @@
       </div>
     </div>
 
-    <div class="mt-2 text-sm line-clamp-2 sm:text-base sm:line-clamp-1 text-gray-700 space-y-4">
-      <p><%= bio %></p>
+    <div class="mt-2 text-sm sm:text-base text-gray-700 space-y-4">
+      <p class="line-clamp-3 break-words"><%= bio %></p>
     </div>
   </div>
 <% end %>
--- a/app/components/developer_card_component.html.erb
+++ b/app/components/developer_card_component.html.erb
@@ -5,7 +5,7 @@
 
   <div class="self-start flex-1">
     <div class="flex items-baseline">
-      <h2 class="text-lg sm:text-xl font-medium text-gray-900 break-words">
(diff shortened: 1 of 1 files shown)


joemasilotti:
Fancy seeing you here!

Thanks for the PR. Can you please screenshot a developer hero with a very long word?

If that looks good then we can merge!

danieldpence:
@joemasilotti That's a good point, and tricky to solve. After playing with various ideas, this seems to best I could come up with so far. How's this feel?

Clamping to 1 line + `title` attribute for an easy way to view the full hero on hover (maybe tooltips in the future?).
Screen Capture on 2021-12-08 at 13-25-21
Screen Shot 2021-12-08 at 1 34 55 PM

joemasilotti:
Good call with the title. I like this. My concern is that a single line is not enough info, though. Let's clamp to 2 lines and, assuming it all looks good, merge this in!

danieldpence:
@joemasilotti yeah, I agree with your concern, for sure, but clamping to 2 lines results in horizontal overflow when you have super long words, which you can counter with `break-all`, but I don't think you want words breaking to a new line in the middle of a word.

`line-clamp-2 break-word`:
Screen Shot 2021-12-08 at 4 33 28 PM

`line-clamp-2 break-all`:
Screen Shot 2021-12-08 at 4 34 27 PM

joemasilotti:
What about line clamp 2, word break, and hidden horizontal overflow?

danieldpence:
I tried that, actually. `line-clamp` includes `overflow: hidden`. The problem is, the hero doesn't have a fixed width, nor does it have a parent element with a defined width, so long words continue to grow without _overflowing_ any parent element (since it's the hero determining the width of the parent).

I'll keep playing with it and see if I can come up with another solution, unless you have other ideas.

joemasilotti:
Weird! But good to know. 

I think the card has a max width. What if we changed that to an explicit width? Or added `w-full`?

danieldpence:
@joemasilotti Turns out it was a flexbox issue, and setting `min-width-0` on one of the parent containers solved the issue. We don't get the ellipsis on long words with the overflow hidden, but hopefully those will be the exception and not the rule.

Screen Capture on 2021-12-08 at 20-02-24

joemasilotti:
Amazing! Great find. I'll merge this in as soon as CI passes. Thanks for your contribution :D

PR #138, opened 2021-12-08
Add time zone to developer profile

caseyjenks:
## Purpose
Adds timezone field to developers

## Approach
* Adds new field `time_zone` to `developers` table.
* Uses `time_zone_select` for the time zone dropdown
* Renders the developer's timezone only if they have it filled out on their profile.

## Screenshot
image


joemasilotti:
It looks like there's a merge conflict in the scheme file and standard is complaining about some formatting.

db/schema.rb
```diff
@@ -10,8 +10,7 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema.define(version: 2021_11_28_095707) do
-
```
joemasilotti, line 14: Why did this blank line get removed?

app/views/developers/_details.html.erb
```diff
@@ -18,4 +18,6 @@
       </div>
     </div>
   </div>
+
+  <%= render TimeZoneComponent.new(developer.time_zone) %>
```
joemasilotti, line 22: If you add a time zone but no social links you get two lines.

image

Let's use [divide](https://tailwindcss.com/docs/divide-color) and remove the individual borders. That will be more future proof in case we add another section/area, too.

app/views/developers/_form.html.erb
```diff
@@ -39,6 +39,13 @@
               <p class="mt-2 text-sm text-gray-500"><%= t(".profile.bio_help_text") %></p>
             </div>
 
+            <div class="sm:col-span-6">
+              <%= form.label :time_zone, class: "block text-sm font-medium text-gray-700" %>
+              <div class="mt-1">
+                <%= form.time_zone_select(:time_zone, ActiveSupport::TimeZone.us_zones, include_blank: true) %>
+              </div>
+            </div>
```
joemasilotti, line 47: image

Can you style this? The screenshot is from desktop Safari.

Tailwind UI has a [Simple native select menu](https://tailwindui.com/components/application-ui/forms/select-menus#component-79592f355ca6f6e3b081e0158109981c) that doesn't require JavaScript. If you don't have access let me know and I can add the classes myself.

joemasilotti (changes requested):
Looking good! Just a few more changes and we should be good to merge.

app/views/developers/_details.html.erb
joemasilotti: I think this should be taken care of in `main` now.

caseyjenks: good deal, I'll grab the latest and rebase.

db/schema.rb
caseyjenks: resolved

app/views/developers/_form.html.erb
caseyjenks: This has been updated as well

joemasilotti (approved):

joemasilotti:
Looks great, thanks @caseyjenks! I made one change to ensure lines don't get rendered between empty sections.

I'll merge when CI finishes.

PR #171, opened 2021-12-19
Enable source maps.

dixpac:
Upgrade `jsbundling-rails, cssbundling-rails and sprockets-rails` in
order to have a source maps generated.
Screenshot 2021-12-19 at 19 18 18


joemasilotti (changes requested):
It looks like something went wrong with the CSS, the site looks very odd in dev.

Also, can you merge in changes from `main`?

image

PR #175, opened 2021-12-21
Test: Model and integration test for developer uploads

### LINKED ISSUE #21: Add tests for avatar and cover photo uploads

jacobdaddario:
Resolves #21. The PR adds model tests and integration tests for avatar and cover image uploads to developer profiles. Also in doing this, I noticed that we need some upload validations to be added to businesses. The avatar concern should probably own those validations.  

Also, I opted to stub out some of the model tests. I think it's efficient and prevents the need for a ton of file fixtures, but it may be a design decision worth taking a second look at.

### Pull request checklist


app/models/developer.rb
```diff
@@ -18,9 +18,9 @@ class Developer < ApplicationRecord
   validates :name, presence: true
   validates :hero, presence: true
   validates :bio, presence: true
-  validates :avatar, content_type: ["image/png", "image/jpg", "image/jpeg"],
+  validates :avatar, content_type: ["image/png", "image/jpeg", "image/jpg"],
     max_file_size: 2.megabytes
-  validates :cover_image, content_type: ["image/png", "image/jpg", "image/jpeg", "image/gif"],
+  validates :cover_image, content_type: ["image/png", "image/jpeg", "image/jpg", "image/gif"],
```
joemasilotti, line 23: Do you remember why we accept a gif for the cover image and not for the avatar?

I just checked production and no one has uploaded a gif - let's remove this accepted format.

test/integration/developers_test.rb
```diff
@@ -47,6 +47,7 @@ class DevelopersTest < ActionDispatch::IntegrationTest
     end
 
     assert user.developer.role_type.part_time_contract?
+    assert_equal "Lovelace.jpg", user.developer.avatar.filename.to_s
```
joemasilotti, line 50: Let's rename this to `lovelace.jpg` (without the capital "L") to match the rest of the files.

test/models/developer_test.rb
```diff
@@ -1,4 +1,5 @@
 require "test_helper"
+require "minitest/mock"
```
joemasilotti, line 2: Let's move this to the test helper.

test/models/developer_test.rb
```diff
@@ -104,4 +105,56 @@ class DeveloperTest < ActiveSupport::TestCase
...
+      end
+    end
+  end
+
+  test "should reject avatars of invalid file formats" do
+    invalid_formats = %w[image/bmp image/gif video/mp4]
+
+    invalid_formats.each do |file_format|
+      @developer.avatar.stub :content_type, file_format do
+        assert_not @developer.valid?, "#{file_format} should be an invalid format"
```
joemasilotti, line 124: Let's use `refute` over `assert_not` to match the rest of the tests.

test/models/developer_test.rb
```diff
@@ -104,4 +105,56 @@ class DeveloperTest < ActiveSupport::TestCase
...
+      end
+    end
+  end
+
+  test "should enforce a maximum avatar file size" do
+    @developer.avatar.blob.stub :byte_size, 3.megabytes do
+      assert_not @developer.valid?
```
joemasilotti, line 131: Same as above.

joemasilotti (changes requested):
Looks great outside of a few small requests. Thanks for getting around to this!

app/models/developer.rb
jacobdaddario: Honestly, I think I was thinking of how Dev.to does it when I added that validation. It was totally arbitrary aside from that though. We never really discussed it.

jacobdaddario:
Made the changes!

joemasilotti (approved):

joemasilotti:
And merged! Thanks again Jacob - excited to have tests around this code.

PR #182, opened 2021-12-23
Build notification UI

### LINKED ISSUE #154: Notification UI
- [ ] A developer or business can view their _unread_ notifications via a link from the navigation bar
- [ ] When a notification is clicked, `read_at` is set

### LINKED ISSUE #155: Notification badge
- [ ] A red dot is shown on the notification icon in the header when a business or developer has unread notifications

jayeclark:
Fixes #154 and fixes #155 

Raising a draft PR for feedback on layout/UX. I'm going to close the earlier one out as that branch is significantly behind the main repo now.

It's working fairly well on the front end, but it isn't the prettiest or most efficient under the hood yet (I'm extremely junior at Ruby on Rails and Ruby in general.)

I've added a gif below of the UI.
- The notification bell will always appear when a user is logged in. 
- If there are no new notifications, the bell will not have the red 'alert' dot. Clicking on the bell will take you to the ‘history’ page
- If there are new notifications, the bell will have the red ‘alert’ dot. Clicking on the bell will take you to the ‘notifications’ page
- On the new notifications page, if there are old notifications a button will appear that links to the history route.
- On the history page, if there are new notifications a button will appear that links to the notifications route.
- Notifications display the sender's name and the business name (if applicable), and contain a link to the conversation that the new message was sent in.
- There is a clickable 'x' in the upper right corner of each new notification. Clicking on the x will dismiss the new notification by adding a 'read_at' time. This notification will then be visible in 'view history'. 

Questions:
- Does this seem like the right set of functionality? 
- Would this affect the admin views in any way? I haven't even started looking at that side of things.
- If a user has both a developer and a business profile, they can receive notifications for both of these profiles. Should there be any distinction or separation between them in the notifications view? 
- Right now I have the new notifications route set to /notifications and the old (read) notifications route set to /history but am open to suggestions on a better name or path for either.
- Currently the new notifications alert will only appear when a user navigates to a new area of the site. Should I add a listener that checks every X minutes and updates the sign in component if there are new notifications? Use case: I sign in, send a message, navigate to another tab… and when I come back it still looks like I have no new notifications until I reload that page or navigate to another … even though I have indeed gotten a response to my message in the interim.

### Pull request checklist

functionality


jayeclark:
Just a heads up, the continuous integration actions are catching standardrb errors that bin/check for some reason is not flagging. (Apologies for the extra commits as a result - let me know if you want me to reorganize everything into a single commit or a specific set of commits.)

When I run `standardrb --fix` in the CLI, it will flag and/or correct the errors, but the same errors are not always caught or corrected with `bin/check`. The check script itself looks fine, and works correctly *most* of the time, so I'm not sure what's up.

joemasilotti:
This is looking great so far! Awesome progress.

* I don't think we need the "new notifications" link on the history page. The red dot in the navigation bar will always be there to navigate to new notifications.
* Ideally, we can consolidate some of the display logic of the notification and email that gets sent into the notification model via `#title` or `#message` or similar. I think they should be the same, or at least very similar, so that should work well. The Noticed docs and example should help here.
* I think we can skip the "X" functionality for this PR. I like the idea, but without a remote/Turbo request it doesn't feel any better than clicking through. Once we move this to a drop down or lazy-loaded menu item we can add the "dismiss" X back in.

For your questions:

1. It definitely feels like the right set of functionality! I'm calling out 2 small bits that we can even cut from this PR and branch out to new change requests.
2. Nope, don't worry about admin. There's only a single admin screen to view _blocked_ conversations. It's more of a moderation thing than anything.
3. Correct, a user can be both a developer and a business. I don't think there needs to be any distinction between the two. `/conversations`, for example, calls into `User#conversations` which merges developer and business ones.
4. I love the idea of `/notifications`, but let's nest history so it's `/notifications/history` or `/notifications/read`. Ideally that's `HistoryNotificationsController#index` (which sounds a little weird, so maybe `ReadNotificationsController` or something?) and not a custom method `#history` on `NotificationsController`.
5. I think there's 2 things going on here that we can address. First, the bell and dot should be a [permanent Turbo element](https://turbo.hotwired.dev/handbook/building#persisting-elements-across-page-loads). I think that should fix the issue of navigating back and seeing an incorrect bell dot. Second, we will eventually want to wire the notifications up to Action Cable from the notification so the bell is updates asynchronously. But that's not for this PR and can come later.

Excited to see this progress!

jayeclark:
Thanks so much for the feedback! I’ll take a crack at implementing those changes and also rebase all my commits so the code changes are more organized / easier to review.

joemasilotti:
Sounds great! And feel free to merge in master and not worry about rebasing. I'll squash everything into one commit so the history won't be preserved anyway.

jayeclark:
I've implemented all the suggestions except Turbo - I can probably tackle that as well, but need some time to digest the docs.

**"I don't think we need the "new notifications" link on the history page..."**  
I removed this.  
  
**"Ideally, we can consolidate some of the display logic of the notification and email that gets sent into the notification model..."**  
I looked through the docs but didn't see anything particularly helpful as only the message is passed as a parameter to the db and importing the `new_message` template from the mailers directory into the notification partial seemed very tricky. For now, pending more feedback on how you want to achieve that consistency, I just copied the template for the mailer into the notification partial.   
  
**"I think we can skip the "X" functionality for this PR..."**  
Removed the X but added a parameter 'redirect' to the update method, which should allow the X to be added back in more easily in the future. (Clicking X would dismiss & redirect to the notifications page, while a general click on the notification dismisses and redirects to the relevant conversation.)  
  
**"Don't worry about rebasing or cleaning up the git history. I will squash all the commits into one merge commit when merging into main."**  
I reorg'd of my commits so I could do a final check before moving out of draft. It definitely needs a good look at the code, I was at zero on Ruby (and Rails) before this PR and so much of the documentation I've been working with feels outdated/out of style.  
  
** "I love the idea of `/notifications`, but let's nest history so it's `/notifications/history` or `/notifications/read`."**  
I went with `/notifications/read` and `ReadNotificationsController`

joemasilotti:
This is really coming along nicely! All of your comments make sense with the exception of the `redirect` parameter. If code isn't going to be used then I'd prefer it removed from the code base. Also, I haven't looked at the code, yet.

Product-wise, I'm rethinking the two different notification pages now that I've used it. I'm curious if it would be a bit simpler and easier to understand if _all_ the notifications are always shown. Then, show a colored dot or strip of color on the left of a notification if it is unread. We can sort all notifications by when they were created and the dot can still indicate if any of those notifications are unread.

I'm thinking this will translate better to when we move the views to the dropdown from the menu. Also, having the bell icon link to a different screen depending on if you have unread notifications or not feels a little weird.

---

I'd love to hear what you think!

jayeclark:
Sure, notifications on dev.to work similarly to how you described (all on one page). It can quickly get visually cluttered, though, and it’s hard to tell which ones are new, so a fair amount of work would probably need to go into the logic of consolidating them. Feels like it might be simpler to just only display new notifications and not have a way to access the old ones.

Google’s Firebase admin interface has the bell link to the new notifications page whether they exist or not, and then has a ‘view history’ button for any old notifications that exist. But so far, looking at my notifications history, I can’t really see why as a user I would need t9 look at the old ones.

joemasilotti:
Great points. Let's meet somewhere in the middle.

1. Always link from the bell to `/notifications`
2. Always show a "View history" button at the bottom

That should simplify a bit of the logic and the UX, too. We won't have to worry about cluttering with a ton of notifications or adding a color/dot to unread ones. It also let's someone always view their old notifications in case they accidentally read one and missed it.

jayeclark:
I've removed the redirect parameter from the update method, and updated the signed_in_component UI to always link from the bell to /notifications (and then display a button linking to past notifications if any past notifications exist.)

app/controllers/read_notifications_controller.rb
```diff
@@ -0,0 +1,15 @@
+class ReadNotificationsController < ApplicationController
+  before_action :authenticate_user!
+
+  def index
+    @read_notifications = (read_biz_notifications + read_dev_notifications).sort { |a, b| b.created_at <=> a.created_at }
+  end
+
+  def read_dev_notifications
```
joemasilotti, line 8: Same comment about `User.conversations` as above.

app/views/notifications/_notification.html.erb
```diff
@@ -0,0 +1,17 @@
...
+<div role="list" class="w-full max-w-xl mx-auto rounded-lg overflow-hidden border border-gray-200 divide-y divide-gray-200 mt-3">
+  <%= button_to notification_path(notification), method: :patch, class: "w-full" do %>
+  <div class="relative bg-white py-5 px-6 hover:bg-gray-50 focus-within:ring-2 focus-within:ring-inset focus-within:ring-blue-600">
+    <div class="flex justify-between space-x-3">
+      <span class="shrink-0 whitespace-nowrap text-sm text-gray-500">
+      <%= render TimeComponent.new(notification.created_at) %>
+      </span>
+    </div>
+    <div class="mt-3">
+    <p class="line-clamp-2 text-sm text-gray-600 text-left">
```
joemasilotti, line 11: I think this line needs to be indented.

app/views/notifications/index.html.erb
```diff
@@ -0,0 +1,18 @@
...
+    <% if @notifications.any? %>
+    <h1 class="mt-6 text-center text-3xl font-extrabold mb-5"><%= t(".title") %></h1>
+    <%= render @notifications %>
+  <% else %>
+    <div class="mt-8">
+      <%= render "empty_state" %>
+    </div>
+  <% end %>
+    <div class="mt-6 mx-auto">
+    <% if @has_read_notifications %>
```
joemasilotti, line 11: This line also needs to be indented.

config/locales/en.yml
```diff
@@ -165,6 +165,13 @@ en:
     message:
       wrote: wrote
       you: You
+  notifications:
+    empty_state:
+      body: You have no new notifications available to display.
```
joemasilotti, line 170:
```suggestion
      body: You have no unread notifications.
```

config/locales/en.yml
```diff
@@ -201,6 +208,12 @@ en:
   pundit:
     errors:
       unauthorized: You don't have access to that.
+  read_notifications:
+    empty_state:
+      body: You have no past notifications available to display.
```
joemasilotti, line 213:
```suggestion
      body: You have no read notifications.
```

config/routes.rb
```diff
@@ -9,6 +9,8 @@
   resource :pricing, only: :show, controller: :pricing
   resource :role, only: :new
   resources :businesses, except: :destroy
+  resources :notifications, only: [:index, :update], controller: :notifications
```
joemasilotti, line 12:
```suggestion
  resources :notifications, only: %i[index update], controller: :notifications
```

config/locales/en.yml
```diff
@@ -201,6 +208,12 @@ en:
...
+      title: No past notifications
+    index:
+      title: Past notifications
```
joemasilotti, line 216:
```suggestion
      title: No read notifications
    index:
      title: Read notifications
```

app/components/user_menu/signed_in_component.html.erb
```diff
@@ -1,4 +1,12 @@
 <div class="absolute inset-y-0 right-0 flex items-center pr-2 sm:static sm:inset-auto sm:ml-6 sm:pr-0">
+  <div class="text-white px-2" style="position:relative">
+  <%= link_to notifications_path do %>
+    <%= inline_svg_tag "icons/outline/notification.svg", class: "mx-auto h-6 w-6 text-gray-400" %>
+        <% if unread_notifications.any? %>
+        <div id="notification-alert" class="absolute rounded-full bg-red-600 border-2 border-white w-3 h-3 top-1 right-2"></div>
```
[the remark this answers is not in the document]
joemasilotti: Also also, this should be indented.

test/components/user_menu/signed_in_component_test.rb
```diff
@@ -60,6 +60,46 @@ class UserMenu::SignedInComponentTest < ViewComponent::TestCase
...
+    user = users(:with_business)
+
+    render_inline UserMenu::SignedInComponent.new(user)
+    assert_link_to notifications_path
+  end
+
+  test "links to new notifications view when new notifications exist" do
+    user = users(:with_business)
+    developer = developers(:available)
+    Message.create!(developer: developer, business: user.business, sender: developer, body: "Hello!")
```
joemasilotti, line 73: We can use the `users(:with_business_conversation)` fixture instead of creating a new `Message` in the test. This also applies to the rest of this test file.

test/components/user_menu/signed_in_component_test.rb
```diff
@@ -60,6 +60,46 @@ class UserMenu::SignedInComponentTest < ViewComponent::TestCase
...
+  test "shows red alert dot when new notifications exist" do
+    user = users(:with_business)
+    developer = developers(:available)
+    Message.create!(developer: developer, business: user.business, sender: developer, body: "Hello!")
+
+    render_inline UserMenu::SignedInComponent.new(user)
+    assert_element "notification-alert"
+  end
+
+  test "does not show red alert dot when no new notifications exist" do
```
joemasilotti, line 88: This test tests that no notification exists at all, it doesn't test that there is a notification but it is _read_.

test/integration/notifications_test.rb
```diff
@@ -0,0 +1,58 @@
...
+
+    get notifications_path
+    assert_select "h3", "No new notifications"
+  end
+
+  test "you can view your new notifications if you have new (unread) notifications" do
+    user = users(:with_business)
+    developer = developers(:available)
+    sign_in user
+    Message.create!(developer: developer, business: user.business, sender: developer, body: "Hello!")
```
joemasilotti, line 20: Same comment as above regarding the fixture.

test/integration/notifications_test.rb
```diff
@@ -0,0 +1,58 @@
...
+    notification = Notification.last
+
+    patch notification_path(notification.id), params: {id: notification.id, redirect: conversation_path(notification.conversation)}
+
+    get notifications_path
+    assert_select "h3", "No new notifications"
+  end
+
+  test "you cannot mark notifications as read if you are not their recipient" do
```
joemasilotti, line 39: We might be able to clean this test up by grabbing a message's notification from a fixture that isn't owned by the signed in user.

joemasilotti (changes requested):
This is looking great so far! I requested a fair amount of changes but they are mostly minor. I think there are two big things that need to be addressed.

First, and this is outlined in the review comments, the querying for notifications. `User.notifications` can be copied and modified to grab all of a user's notifications at once. Then we can keep them in a query instead of converting to an in-memory array.

Second, we need a way to mark notifications as read when clicking through an email. When a developer receives a message they get an email linking them directly to the conversation. We should either a) mark a notification as read when viewing a conversation or b) link the email to the notification and use the redirect logic in the controller you've built already out.

I think the first one will require more changes to your PR but might be more robust in the long term. The second should be only a line or two of changes. I'm open to your feedback and thoughts on this as I'm not sure which is better right now.

---

As always, I really appreciate the work you've put in on this! Let me know if you need more guidance on some of my comments or want to bump any of the work to me.

jayeclark:
Hey @joemasilotti , thanks for the feedback - I especially appreciate the stylistic comments as like I said, I'm completely new to Ruby and Ruby on Rails and have been having a hard time finding docs and resources that aren't super outdated. 

May I ask, have you tried your suggested code changes locally, specifically the one about querying for user.notifications instead of business.notifications + developer.notifications? 

My understanding from how the code is written (and my observation from testing both approaches - I started with user.notifications and had to adjust from there when it didn't work) is that the user.notifications query will only retrieve notifications where the recipient_type is user. It will not retrieve notifications where the recipient_type is business or developer (which are the recipient_types that the message notifications get sent to.) As far as I can tell, only admin users will have a non-zero number of notifications when the user.notifications query is used, and if we wanted to show ALL notifications that would mean combining user.notifications + business.notifications + developer.notifications...

If I change the controller query to user.notifications (well, technically `current_user.notifications`) I end up getting zero notifications on both the new and read pages (even though notifications exist), because only the admin notifications are sent to recipient_type "User".  

Screenshot from the previously 'working' read notifications page to show what happens when I change the query to user.notifications:
Screen Shot 2021-12-29 at 12 50 14 PM

I like the idea about marking all notifications for a conversation as 'read' when the conversation is viewed, I think that's a more elegant way to do it. It covers several use cases with one block of code, including one we haven't discussed yet: if I manually navigated to my conversations on the site, and read a conversation for which there was a new notification, I would expect the new notification dot to disappear and the notification to no longer appear in my new notifications - even though I didn't interact with the notifications page at all along the way. (I should not be receiving a visual prompt from the site about something 'new' that I've actually already seen. 🙂 )

app/components/user_menu/signed_in_component.html.erb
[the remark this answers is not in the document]
jayeclark: Yep, id is not necessary - it's left over from an earlier unnecessarily complicated version of the code that didn't even make it to the PR.

joemasilotti:
Ah, great call on navigating directly to a conversation! Let's go with that approach then: mark all of the conversation's notifications (for that user) as read when viewing a conversation.

For the query, I was thinking of something like this. But I tried it and it doesn't actually work! So I'm thinking of a better query (or this is wrong) that can work. Great catch.

```ruby
has_many :message_notifications, ->(user) {
  unscope(where: :user_id)
    .left_joins(:business, :developer)
    .where("businesses.user_id = ? OR developers.user_id = ?", user.id, user.id)
}, class_name: "Notification"
```

app/components/user_menu/signed_in_component.html.erb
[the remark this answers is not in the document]
jayeclark: For the screen reader element, what do you think of:
1. Always-on screen reader only span "Open notifications page" for the bell itself
2. Conditional screen reader only span "You have a new notification" if a new notification exists/the red dot is displayed

app/components/user_menu/signed_in_component.rb
```diff
@@ -24,4 +24,16 @@ def conversations?
   def admin?
     user.admin?
   end
+
+  def unread_notifications
```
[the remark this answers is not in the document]
jayeclark: The methods return an array rather than a boolean... should they still be questions? 

I could also simplify things by making them actually return a boolean - I think the reason they aren't is that I was trying to plan ahead a little bit as we had talked about making new notifications be a dropdown from the notification bell in the future.

test/components/user_menu/signed_in_component_test.rb
jayeclark: I'm a little confused by this. I read the documentation on Rails unit testing but still have so many questions.

I see the users.yml file that defines the various fixtures. But I don't see anything creating any business conversations for the user "with_business_conversation", . How do I know that they have a new notification or a read notification? Does the testing suite create a copy of the database, so that any changes are not permanent? (ie if the test updates the db so that the new notification gets set as read, that won't affect future tests that depend on a new notification existing.)

test/components/user_menu/signed_in_component_test.rb
jayeclark: Going to hold off on this until I get my head straight about the testing suite.

jayeclark:
I tried something almost identical (using :recipient_id instead of :user_id IIRC), and also couldn't get it to work. 

If you inspect a random Notification record itself, it becomes clearer where the issue is (I think.) Because the record is polymorphic, the `recipient_id` associated with the record could be a user id, business id, or developer id. So it's possible to have three notifications with `recipient_id == 10` that all belong to different users (i.e. User 10 who has Business 1 and Developer 1, some Developer 10 who belongs to some other User, and Business 10 who belongs to a third unique User.)

joemasilotti:
That's almost exactly the same conclusion I came to on the plane last night! Here's something that technically works, but is very ugly.

```
# app/models/user.rb:34

  def message_notifications
    if developer.present? && business.present?
      developer_notifications.or(business_notifications)
    elsif developer.present?
      developer_notifications
    elsif business.present?
      business_notifications
    end
  end

  private

  def developer_notifications
    Notification.where(recipient_type: "Developer", recipient: developer)
  end

  def business_notifications
    Notification.where(recipient_type: "Business", recipient: business)
  end
```

The "magic" is the `#or` chain which merges the two queries into one. Also, setting `recipient_type` ensures we don't run into the issue you called out.

---

We might be able to clean this up by doing an actual join on the polymorphic relationship, too. https://veelenga.github.io/joining-polymorphic-associations/

For now, let's roll with the Ruby code above (or similar), we can come back to the query later and clean it up if this approach ends up feeling better!

jayeclark:
I've been playing around with it a bit too. I may be getting into 'dumb question' territory here, but... what's the difference between using the code above vs something like this?

```
  def message_notifications
    developer&.notifications.or(business&.notifications)
  end
```

joemasilotti:
Great question!

If the user doesn't have a business then you will get an error. Since `&` is returning `nil`, the entire `#or` clause is empty.

> ArgumentError: You have passed NilClass object to #or. Pass an ActiveRecord::Relation object instead.

And if the user doesn't have a developer than the `#or` can't chain on `nil.

> NoMethodError: undefined method `or' for nil:NilClass

jayeclark:
Ah, makes sense. I was testing it on an account with both a business and a developer.

jayeclark:
OK, I've pushed two new commits with my changes so far based on the review, plus an additional commit to deal with a random test failure (not entirely sure why it was suddenly coming up), and I've marked the relevant conversations as resolved. 

I still have a few questions above (in the review section) that are waiting for feedback before I can resolve them.

Once all the convos are resolved and I've got commits in implementing their changes, I'll add a further commit implementing the 'conversation notifications get marked as read when conversation is viewed' logic.

test/components/user_menu/signed_in_component_test.rb
joemasilotti: I'm new to fixtures, too, so we can learn together! I usually use FactoryBot, but am trying minitest/fixtures on this project because it is the default with Rails.

From my understanding, when the test suite starts each fixture is inserted into the database. Each test is also run in a transaction, so any db changes are reverted at the end of the test.

Figuring out which fixtures rely on others is a bit tricky, I'll admit. You kind of have to think of it inside-out. For example, to see the message for `users(:with_business_conversation)` I have to dig into `businesses(:with_conversation)` which is referenced from `conversations(:with_conversation)` which in turn is referenced from `messages(:from_business)`.

I just learned that [fixtures don't trigger callbacks](https://stackoverflow.com/a/1223192/384110). So on a test run there actually won't be any notifications getting created until you create a message manually. 

Does that make sense? I realize my initial advice was wrong!

test/integration/read_notifications_test.rb
```diff
@@ -0,0 +1,28 @@
...
+  end
+
+  test "you can view your past notifications if you have past (read) notifications" do
+    user = users(:with_business)
+    developer = developers(:available)
+    sign_in user
+    Message.create!(developer: developer, business: user.business, sender: developer, body: "Hello!")
+    notification = Notification.last
+
+    patch notification_path(notification), params: {id: notification.id, redirect: conversation_path(notification.conversation)}
```
joemasilotti, line 23: I think that a fixture for a read notification will go a long way for your new tests, especially this one. Then we can reference that one, or the business with that one, directly without having to rebuild state.

test/components/availability_component_test.rb
```diff
@@ -26,7 +26,7 @@ class AvailabilityComponentTest < ViewComponent::TestCase
     render_inline(AvailabilityComponent.new(developer: @developer))
 
     assert_no_selector("svg")
-    assert_selector("span", text: /Available\s*in\s*2 months/m)
+    assert_selector("span", text: /Available\s*in(\sabout){0,1}\s*2 months/m)
```
joemasilotti, line 29:
```suggestion
    assert_selector("span", text: /Available\s*in about\s*2 months/m)
```

This passed 20x in a row for me. Does it work for you?

app/controllers/notifications_controller.rb
```diff
@@ -0,0 +1,21 @@
...
+    @has_read_notifications = current_user.message_notifications.read.any?
+  end
+
+  def update
+    notifications = current_user.message_notifications.unread
+    @notification = Notification.find(params[:id])
+
+    if notifications.include?(@notification)
+      @notification.mark_as_read!
+    end
```
joemasilotti, line 15: This is handling permissions/authorization, right? Like, if I `PATCH` to this route for your notification it shouldn't mark it as read?

If that's the case, then I'd like to replace this logic with a pundit authorization policy. `app/controllers/messages_controller.rb:7` and `app/policies/messaging_policy.rb:6` are a good example of how we can move that logic into a new class like the rest of the codebase.

joemasilotti (changes requested):

config/routes.rb
```diff
@@ -9,6 +9,8 @@
   resource :pricing, only: :show, controller: :pricing
   resource :role, only: :new
   resources :businesses, except: :destroy
+  resources :notifications, only: %i[index update], controller: :notifications
+  resources :read_notifications, path: "notifications/read", only: :index, controller: :read_notifications
```
joemasilotti, line 13:
```suggestion
  resources :notifications, only: %i[index update]
  namespace :notifications do
    resources :read, only: :index
  end
```

This will route to a `Notifications::ReadController` and I think is a little cleaner. Less code, and more conventional, at least!

app/components/user_menu/signed_in_component.rb
```diff
@@ -24,4 +24,8 @@ def conversations?
   def admin?
     user.admin?
   end
+
+  def unread_notifications?
+    user.message_notifications&.unread&.any?
```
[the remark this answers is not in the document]
jayeclark: Hmmm ok, I’ll read up on this so I don’t make the same error in the future. I could have sworn I was getting errors when I tried to call @current_user.message_notifications.read on an empty collection.

test/components/availability_component_test.rb
jayeclark: @joemasilotti I was going to call this out in a separate issue actually. Yes, your version should work too with how the availability is currently coming back. Problem is, I have no idea why the span suddenly changed from ‘in two months’ to ‘in about two months’ - I didn’t make any changes to my code that would have caused it, which makes me think it might be due to an incorrectly set up fixture that might have its availability output change again in the future (ie change to ‘in 1 month.’)  

The other possibility is that February being an unusually short month changes the logic of Timecomponent’s output. If the content of the span changes based on which months of the year are ahead at the time the test is run, you’re probably going to want to keep my regex in there or change the availability date so that the output always has ‘about’ in it. (But I haven’t had time to troubleshoot as I’m try to stay focused on the initial PR review issues.)

app/controllers/notifications_controller.rb
jayeclark: The notification update (mark as read) logic is moving to conversations in my next major commit on this PR so I don’t think this work is necessary. This whole method is disappearing.

config/routes.rb
jayeclark: I will test locally. I thought I’d tried that early on and got unresolvable errors, but it may have been my syntax.

app/views/notifications/_notification.html.erb
```diff
@@ -0,0 +1,17 @@
+
+<div role="list" class="w-full max-w-xl mx-auto rounded-lg overflow-hidden border border-gray-200 divide-y divide-gray-200 mt-3">
+  <%= button_to notification_path(notification), method: :patch, class: "w-full" do %>
+    <div class="relative bg-white py-5 px-6 hover:bg-gray-50 focus-within:ring-2 focus-within:ring-inset focus-within:ring-blue-600">
+      <div class="flex justify-between space-x-3">
+        <span class="shrink-0 whitespace-nowrap text-sm text-gray-500">
+        <%= render TimeComponent.new(notification.created_at) %>
```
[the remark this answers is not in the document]
jayeclark: Yeah, sorry about all these indentation errors. They're happening when I copy/paste to move things around - VSCode is changing properly indented blocks to have incorrect indentation. The erb-lint extension isn't showing any errors either so I originally thought maybe the behavior was intentional.

test/components/availability_component_test.rb
jayeclark: OK @joemasilotti I just looked more closely at the test (the answer I needed was just a couple lines above, duh!) and have found some unexpected behavior within the test suite.

@developer.available_on = Date.today + 2.months : 'Available in about 2 months'
@developer.available_on = Date.today + 45.days (up to 60 days) : 'Available in about 2 months'
@developer.available_on = Date.today + 61.days (up to 75 days): 'Available in 2 months'

What is happening is that + 2.months handles February weirdly. Today is 12/31, adding 2 months would bring me to 2/31.... since 2/31 doesn't exist, what is happening is that it is just defaulting to the last day in February. However, the time_in_words is being generated based on the number of days between today and the available_on date.

I'd suggest changing the test to add a specific number of days to today's date, OR leaving the regex in. If you change it to absolutely test for the span 'in about 2 months', the test is going to fail again starting tomorrow.

app/components/user_menu/signed_in_component.rb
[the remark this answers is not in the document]
joemasilotti: Yeah, let's keep it as a question so it moves all the logic, `#any?`, to the Ruby code instead of the view.

app/components/user_menu/signed_in_component.rb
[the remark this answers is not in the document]
joemasilotti: It may have been an empty array, not an Active Record query result.

test/components/availability_component_test.rb
joemasilotti: Ah, nice find! Let's change the date manipulation so it doesn't ever render the "about", regardless of February and such.

app/controllers/notifications_controller.rb
joemasilotti: Gotcha, sounds good.

config/routes.rb
joemasilotti: FYI, the benefit of this type of routing is explained here: http://jeromedalbert.com/how-dhh-organizes-his-rails-controllers/

Also, you can check the output of the routes file by visiting `http://localhost:3000/rails/info/routes` or running `rails routes`

jayeclark: I'm making progress on getting this to work, I think. It feels very convoluted though - and definitely not less code. So I'm wondering if I'm doing it wrong.

app/components/user_menu/signed_in_component.rb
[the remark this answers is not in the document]
jayeclark: Yes, this was corrected in a recent commit. Marking this convo as resolved.

app/components/user_menu/signed_in_component.rb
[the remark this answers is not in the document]
jayeclark: Here's the error that comes up on a newly registered account (ie one for which message_notifications is totally empty) when I make that change (and why I had to add in the safe navigation in the first place): 
Screen Shot 2021-12-31 at 12 25 46 PM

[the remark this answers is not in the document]
jayeclark: The reason for the error is that the user hasn't set up a developer or business profile yet, so message_notifications is nil instead of an empty query result.

config/locales/en.yml
```diff
@@ -225,6 +238,8 @@ en:
       my_business_profile: My business profile
       my_conversations: My conversations
       my_developer_profile: My developer profile
+      new_notifications: You have a new notification
+      notifications_icon: Open new notifications page
```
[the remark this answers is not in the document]
jayeclark: Sounds good. This change was included on my recent commit refactoring routes.rb (and the changes that needed to be made elsewhere in the code to accommodate that.)

test/components/availability_component_test.rb
jayeclark: Sounds good - I've updated the code locally and will include that change on an upcoming 'miscellaneous PR review fixes' commit. Marking as resolved to keep my to do list clean.

test/components/user_menu/signed_in_component_test.rb
jayeclark: What are the action items for me on this one? 

Should I be trying to create fixtures for an unread notification and a read notification, and associating them with a message and a business or developer?

app/components/user_menu/signed_in_component.rb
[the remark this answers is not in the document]
joemasilotti: Ah! Yes, that makes sense, I forgot about that edge case. So we can do `!! user.message_notifications&.unread&.any?` which will turn a `nil` into `false.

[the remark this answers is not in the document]
joemasilotti: Actually, I take that back. We should always return an association from that method.

```
def message_notifications
  if developer.present? && business.present?
    developer_notifications.or(business_notifications).order(created_at: :desc)
  elsif developer.present?
    developer_notifications.order(created_at: :desc)
  elsif business.present?
    business_notifications.order(created_at: :desc)
  else
    Notification.none
  end
end
```

This will ensure we always get something back and can chain our scopes.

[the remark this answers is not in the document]
jayeclark: Got it, will make those changes.

config/routes.rb
joemasilotti: Interesting - I thought you should be able to move the `ReadNotificationsController` code into `Notifications::ReadController` and rename the `views/` folders. What additional code is required?

test/components/user_menu/signed_in_component_test.rb
joemasilotti: Let's punt on this one for now. The tests pass and test the behavior - we can refactor later.

config/routes.rb
jayeclark: I've added a commit with the basic changes (I did eventually get it to work.) 

The main thing that seemed convoluted to me was needing to make both 'read' and 'notifications' folder within views/notifications. And how I have two partials for notifications instead of one, even though they should be virtually identical. Should I/could I create a component for notification and just have each _notification.html.erb partial import that component? Is there a better way to do it that I'm not thinking about? I don't want someone in the future to have to update two templates every time a small tweak is made to the general layout of a notification.

Also I didn't originally catch that if I used a namespace, I'd need to create a subfolder within controllers called notifications to hold `Notifications::ReadController`. The docs on routes are... not the clearest. But I think I'm getting up to speed.

joemasilotti: Ah, we definitely don't want that much duplicated logic. Great catch.

I'm fine with duplicating the `index` views, especially since we have slightly different logic going on. However, for the notification partials, we can re-use `app/views/notifications/_notification.html.erb` by explicitly rendering the partial from `app/views/notifications/read/index.html.erb:4`.

```ruby
<%= render partial: "notifications/notification", collection: @read %>
```

I also think we can rename `@read` to `@notifications` since we know the context based on the controller and view names.

jayeclark: OK perfect, this is really helpful.

jayeclark:
OK, I *think* I've taken care of everything other than the tests.

The logic to mark a conversation's new notifications as read when the conversation is viewed is running through arrays, however, rather than ActiveRecord collections... I know you had commented on wanting to avoid that in an earlier review, but I tried for a few hours and could not figure out a query that would work.

app/controllers/conversations_controller.rb
```diff
@@ -9,5 +9,10 @@ def show
     @conversation = Conversation.find(params[:id])
     @message = Message.new
     authorize @conversation, policy_class: MessagingPolicy
+    @unread_notifications = current_user.conversation_notifications(@conversation)
+    if @unread_notifications.any?
+      @unread_notifications.each { |n| n.mark_as_read! }
+      redirect_to conversation_path(@conversation)
+    end
```
joemasilotti, line 16:
```suggestion
    unread_notifications = current_user.conversation_notifications(@conversation)
    unread_notifications.mark_as_read!
    redirect_to conversation_path(@conversation)
```

joemasilotti (changes requested):

joemasilotti:
Awesome! I'm excited, this is looking great.

I'm trying a bunch of different things locally and stumbled across [these docs](https://github.com/excid3/noticed#associating-notifications) from the noticed gem. Adding notification-type-specific methods to `Notification` felt weird, but I couldn't put my finger on why. Now I'm realizing that we have other notification types that aren't necessarily related to a message or conversation.

Without derailing this entire PR, I'm wondering if we want to refactor a bit to use the helper outlined in the linked docs. That said, I still can't get around getting message notifications from a conversation without converting to an array like you're doing.

I'm leaning towards merging all of this in and then coming back and refactoring later. But if you have the appetite to try and refactor now, go for it!

app/models/user.rb
```diff
@@ -23,4 +23,30 @@ class User < ApplicationRecord
...
+    elsif developer.present?
+      developer_notifications.order(created_at: :desc)
+    elsif business.present?
+      business_notifications.order(created_at: :desc)
+    else
+      Notification.none
+    end
+  end
+
+  def conversation_notifications(conversation)
```
joemasilotti, line 39:
```suggestion
  def unread_conversation_notifications(conversation)
```

joemasilotti, line 39:
```suggestion
  def unread_conversation_notifications(conversation)
```

app/controllers/conversations_controller.rb
joemasilotti, line 16:
```suggestion
    unread_notifications = current_user.unread_conversation_notifications(@conversation)
    unread_notifications.each { |n| n.mark_as_read! }
    redirect_to conversation_path(@conversation)
```

jayeclark: Isn't this going to create an infinite loop? Shouldn't the redirect only happen if there are no unread conversation notifications that need to be marked as read?

joemasilotti:
Oh... I think something just clicked. If each `Notification` had two params, both `message:` and `conversation:`, then we can add `has_noticed_notifications` to `Conversation`. Then, in the controller, `conversation.notifications_as_conversation.where(recipient: user.business).mark_as_read!`. However, we would need to know if `current_user` is the business or developer context for the conversation.

Again, not sure if we want to address this in this PR. This is starting to spiral into me wondering if we want to move the ownership of notifications directly to the user instead of developer or business, too. 🤦‍♂️

app/controllers/conversations_controller.rb
joemasilotti: Oh, you're absolutely right. Why, then, are we redirecting at all? We aren't showing any UI for read vs. unread messages so the view should look exactly the same.

jayeclark:
It would simplify things considerably to move ownership to the user and then just have the controller select a specific type of Notification to display . Especially since there's no difference in the display logic between Business owners and Developer owners. What was the original reasoning behind associating them with the different roles in the first place? 

Though I would also very much like to get some of this merged. 😅  I have 15 PR's left in my 100 PR challenge so I'm going to have to focus soon on wrapping those up before my 100 day deadline arrives.

joemasilotti:
I hear you 100%! I was checking out your challenge and admire it; I definitely don't want to hold it up.

Let's do this – let's resolve the last 2 open changes, get this merged in, then I can take a deeper dive outside of this PR. Does that sound good?

app/controllers/conversations_controller.rb
jayeclark: Yeah, you're right, it totally does not need to be in there. 🤦‍♀️ I mistakenly assumed the redirect was necessary in order to make the red dot disappear.

jayeclark:
Sounds good, signing off for now but I'll take care of those in the morning!

joemasilotti:
@jayeclark, is this ready? I see your last commit but wanted to double check.

jayeclark:
Yes, sorry, I must not have saved my comment! I think this should be good to merge now. The tests still need work as we discussed earlier, but they _are_ functioning and should catch breaking changes (even if they aren't using the ideal execution yet.)

I think I should probably wait on refactoring the tests & adding fixtures until some decisions are made about whether to refactor Notification. The idea to add a conversation parameter made a **lot** of sense to me, and so did just associating all Notifications with the user and filtering them by type in order to determine which count as 'new' and 'read' notifications for the notification UI. 

Totally happy to do more work on this/additional PRs btw. I just need to focus the next few days on getting in some of those final PRs in my challenge. 🙂

joemasilotti:
I agree on all accounts. I'm adding a feature flag until those items are taken care of, but this is ready to merge in my opinion.

Thanks so much for the help with this! Almost 100 messages on GitHub while we figured this out together.

Good luck with your other PRs, and let me know when you're ready to dive back in to this. I have a lot of ideas!

jayeclark:
Thanks - I’ll definitely loop back around once I’ve made some progress on other PRs!

Issue #40, opened 2021-11-05
Index site on Google via a sitemap generator


joemasilotti:
I've used [this fork of sitemap_generator](https://github.com/kjvarga/sitemap_generator) in the past, but am open to newer/better tools!

I'll then set up Google Search Console.

vamsipavanmahesh:
A couple of questions:

- Do you want individual developer profiles to be part of the sitemap?
If that's the case, we might want to run a cron job to regenerate this sitemap every day via a cron job

- Do you envision at some time in the future that we will move to multiple servers, if that's the case we cannot have a sitemap file in a single server or multiple servers, need to be in s3 or equivalent and a route will point to that file

But the second point need not be done right now! we can do it when the time comes

davegudge:
Using a [dynamically generated sitemap](https://webdevchallenges.com/how-to-add-a-dynamically-generated-sitemap-with-rails) would remove the need for regenerating and it should work regardless of how the site is hosted.

vamsipavanmahesh:
For a small site it should be fine, but when you have thousands of records, might not be a good solution?

A lot of this will obviously depend on what is the time crawler is willing to time just to get the sitemap?

After getting the sitemap, it has to actually crawl

When I have worked on an SEO project at my company, SEO consultants used to talk about a thing called "crawl budget", basically our question was why only 40 k things were indexed, whereas 100 k URLs. So they used to be talk about how google only gives a budget for each site etc

But overall I would vote for not doing it dynamically assuming that thousands of people are going to come in

davegudge:
You've raised some very good points. 

SEO is such a large topic these days!

joemasilotti:
I've used the fork I linked to at the top for a variety of projects. You set up a single Ruby file that looks a lot like a Rails routing file. You can dynamically index pages pages on path, or resource. It also sets up a single rake task that will publish the sitemap on an S3 bucket and ping Google (and other search engines).

That still feels like the easiest approach for setting up a sitemap on a Rails project. But I'm still open to other ideas or implementations. I want to stay away from manually generating and splitting XML files and am fine bringing in a gem for this.

### What to index?

* the homepage
* `/developers`
* `/about`
* every developer profile
* sign in, sign up, reset password, and resend confirmation instruction pages

vamsipavanmahesh:
Yes, I was talking about using the mentioned gem

> every developer profile

this is where the complexity lies in, example there are 100 developers right now and our URLs are like this

/developers/1
[...]

PR #183, opened 2021-12-24
Add sitemap generator

hungmi:
This PR adds sitemap_generator and add some links mentioned in https://github.com/joemasilotti/railsdevs.com/issues/40.

I also add `/developers/new` & `/businesses/new` since they are part of sign up process. I can remove them If that is not needed.

I have tested it locally and the sitemap.xml seems to be ok.
But I don't know how to write test for it 😅. Does anyone have any suggestion?

And the next step is set `rake "-s sitemap:refresh"`(according to the [doc](https://github.com/kjvarga/sitemap_generator#crontab)) on Heroku scheduler.

### Pull request checklist


Gemfile
```diff
@@ -47,5 +47,5 @@ gem "pundit", "~> 2.1"
 gem "redcarpet", "~> 3.5"
 gem "sidekiq", "~> 6.3"
 gem "stripe", ">= 2.8", "< 6.0"
-
+gem "sitemap_generator"
```
joemasilotti, line 50: Please move this one line above and sort the chunk of gems.

config/sitemap.rb
```diff
@@ -0,0 +1,44 @@
+# Set the host name for URL creation
```
joemasilotti, line 1: Please remove all comments from this file.

config/sitemap.rb
```diff
@@ -0,0 +1,44 @@
...
+  #
+  #   add articles_path, :priority => 0.7, :changefreq => 'daily'
+  #
+  # Add all articles:
+  #
+  #   Article.find_each do |article|
+  #     add article_path(article), :lastmod => article.updated_at
+  #   end
+
+  add developers_path, changefreq: "always", priority: 1, lastmod: Developer.order(id: :desc).first.try(:updated_at)
```
joemasilotti, line 28: `Developer.maximum(:updated_at)` will return the most recently updated timestamp and can be used for the `lastmod:` option.

joemasilotti (changes requested):
Thanks for kicking this off! I made a few small change requests, other than that it looks good.

We will have to host the sitemap somewhere else, like S3, as Heroku doesn't have persistent storage. But I can take care of that.

joemasilotti:
I'm fine with a single smoke test if it doesn't take too long. Running the command and verifying that the XML file is created without any raised errors should be good enough!

Gemfile
hungmi: Updated.
 I thought these names are started with **s**. 🤣

config/sitemap.rb
hungmi: Done!

config/sitemap.rb
hungmi: Updated!

hungmi:
I have added a test to run `rails sitemap:refresh:no_ping`.
But I can't find a way to add `-s` when I invoke it, so I followed [this open issue](https://github.com/kjvarga/sitemap_generator/issues/332) to prevent unneccessarily outputs.
Thanks for your review!

joemasilotti:
This is looking great. I'm just figuring out the S3 hosting and will merge in.

joemasilotti (approved):

joemasilotti:
OK, I think we're good! Turned out I need to add a bunch of stuff to get the S3 upload working and Google Search Console pointing to the sitemap correctly. I'm going to document all of the non-code stuff on the Wiki soon.

Thanks again for the PR!

joemasilotti:
And we're good! Thanks again 🎉

image

PR #199, opened 2022-01-02
Refactor notifications

joemasilotti:
This PR refactors notifications to make them a little easier to interact with. Most of the ideas implemented here come from discussions with @jayeclark in #182. So a huge thank you there!

Most importantly, notifications are now tied to a `User`, not a `Business` or `Developer`. This makes queries much more straightforward as `current_user.notifications` gets them all at once. No more building custom queries.

Second, notification logic now lives on the notification models (not records) themselves. `NewMessageNotification` now has methods for the conversation and message, not `Notification`. This is largely thanks to the discovered `#to_notification` method from noticed that rehydrates the instance from the params initially sent in.

Finally, it refactors how notifications are "read" by navigating to `/notifications/:id` and then redirecting. The second refactor enabled this to be much simpler, mostly due to the `#to_notification` method from noticed. If a `Notification` instance doesn't have a url then it is marked as read but you are redirected to `/notifications` (with a flash message).

A backfill was added that addresses a few things. First, it sets the recipient of all `NewMessageNotification` type records to the user associated with the developer or business. Second, it deletes some dangling notification records from deleted developer profiles. Finally, a separate rake task marks all existing notifications as read.

Edit: The backfill now also adds the conversation to the notification records (as params) which allows easy querying for batch updating: `conversation.notifications_as_conversation` grabs all the notifications triggered by sent messages in the convo. Adding this to `ConversationsController` means marking all (unread) notifications as read when viewing a conversation, all with a single query!

---

I'm going to leave this in draft tonight and revisit tomorrow morning as there's a lot going on. Especially with the backfills!

### Pull request checklist


app/controllers/conversations_controller.rb
```diff
@@ -9,13 +9,5 @@ def show
...
     @message = Message.new
     authorize @conversation, policy_class: MessagingPolicy
-
-    if Feature.enabled?(:notifications)
-      @unread_notifications = current_user.conversation_notifications(@conversation)
-      if @unread_notifications.any?
-        @unread_notifications.each { |n| n.mark_as_read! }
-        redirect_to conversation_path(@conversation)
-      end
-    end
```
joemasilotti, line 19: I'm still not sold on removing this. If someone navigates directly to a conversation from the user menu then all of the associated message notifications will still be unread.

joemasilotti, line 19:
```ruby
    @conversation.notifications_as_conversation.where(recipient: recipient).mark_as_read!
  end

  private

  def conversation
    @conversation ||= Conversation.find(params[:id])
  end

  def recipient
    if conversation.business?(current_user)
      current_user.business
    elsif conversation.developer?(current_user)
      current_user.developer
    end
  end
```

```ruby
# lib/tasks/message_notifications.rake:2
  task backfill_conversations: :environment do
    Notification.where(type: "NewMessageNotification").find_each do |notification|
      instance = notification.to_notification
      notification.update!(
        recipient: notification.recipient.user,
        params: {
          message: instance.message,
          conversation: instance.conversation
        }
      )

      # The associated developers with these notifications where deleted.
      Notification.where(id: [1, 98]).destroy_all
    end
  end
```

joemasilotti:
```ruby
    @conversation.notifications_as_conversation.where(recipient: recipient).mark_as_read!
  end

  private

  def conversation
    @conversation ||= Conversation.find(params[:id])
  end

  def recipient
    if conversation.business?(current_user)
      current_user.business
    elsif conversation.developer?(current_user)
      current_user.developer
    end
  end
```

```ruby
# lib/tasks/message_notifications.rake:2
  task backfill_conversations: :environment do
    Notification.where(type: "NewMessageNotification").find_each do |notification|
      instance = notification.to_notification
      notification.update!(
        recipient: notification.recipient.user,
        params: {
          message: instance.message,
          conversation: instance.conversation
        }
      )

      # The associated developers with these notifications where deleted.
      Notification.where(id: [1, 98]).destroy_all
    end
  end
```

jayeclark: It feels like an important part of user experience that I wouldn’t be promoted with new notifications for a conversation I’d already read directly. I could see people getting frustrated/confused — ‘It said I had a new notification from Jay Clark but when I went to the conversation it was just messages I’d read already.’ 

What I originally wanted to implement was a conversation update path. The notification could link to the update path and any conversations with new messages would also link to the update method while those without notifications would just link to the regular get path. 

I eventually abandoned that because the query for new notifications would have to run on every conversation rather than just the one the viewer was looking at, so this seemed like a lot of extra work. Also I would have had to do something with routes that seemed like it probably wasn’t The Rails Way(TM).

Another idea I briefly toyed with, FWIW, was that sending a message notification would also update a column in the conversation record. Then rather than being a separate query, the ‘unread notifications’ column in the table could determine whether or not to do the work of querying for new notifications and marking them as read if the user was the recipient (and update the column to false if both recipients had read all notifications), or just display the conversation.

joemasilotti: > It feels like an important part of user experience that I wouldn’t be promoted with new notifications for a conversation I’d already read directly.

After sleeping on it, I totally agree. I think this needs to be implemented one way or another.

> Another idea I briefly toyed with, FWIW, was that sending a message notification would also update a column in the conversation record.

Oh, that's interesting. Kind of like a cache to determine where/how to link someone. I like that! One wrinkle is if someone bookmarks the conversation page and refreshes it without going through this new route. So I think any logic that marks stuff as read still needs to be on the conversation page.

jayeclark: Instead of that extra field determining where/how to link the conversation, I was envisioning it being a logic gate inside the ‘show’ method of the conversation controller.

All links would just link to the conversation. If the conversation was flagged as having new notifications, some additional logic would run within the show method to clean up unread notifications to the user about that conversation. If the conversation was not flagged as having new notifications, there would be no cleanup action. 

IDK, maybe that’s not how it’s supposed to be done? Is it ‘normal’ for get requests in Rails to trigger side effects like that?

The way I wrote it in my last PR, *every* time a conversation was viewed the db was being queried for notifications related to the conversation and user, and any unread ones were marked as read. I’m not really far enough along in understanding resource usage in this stack yet, but querying the db for unread notifications on every conversation view seems inefficient compared to changing a field in the relevant conversation when a message notification is sent, which then determines whether the query needs to be run (and notifications updated) when the conversation is viewed.

joemasilotti: Great points, I think there is something a little screwy with making db updates on a GET request.

That said, for now, I've gone and re-implemented your previous workflow. 

> The backfill now also adds the conversation to the notification records (as params) which allows easy querying for batch updating: `conversation.notifications_as_conversation` grabs all the notifications triggered by sent messages in the convo. Adding this to ConversationsController means marking all (unread) notifications as read when viewing a conversation, all with a single query!

This feels like the most pragmatic solution for now, at the small expense of an additional query every time a conversation is viewed.

Issue #201, opened 2022-01-02
Consolidate empty state image sizes


joemasilotti:
`/notifications/read` and `/conversations` have different sized icons for their empty state. They should all be the same and match what's currently being shown for `/conversations`.

<img width="425" alt="image" src="https://user-images.githubusercontent.com/2092156/147883766-e8946316-583c-43d5-bb40-bf630932a0bc.png">

<img width="387" alt="image" src="https://user-images.githubusercontent.com/2092156/147883770-6f82da3c-8269-49fb-87be-f023bb941ccc.png">

joemasilotti:
Closed via #202.

PR #202, opened 2022-01-02
Introduce EmptyStateComponent

jespr:
This implements an EmptyStateComponent for #201 - a bit unsure how to best test the inline svg

### Pull request checklist

### DIFF
--- a/app/components/empty_state_component.rb
+++ b/app/components/empty_state_component.rb
@@ -1,0 +1,9 @@
+class EmptyStateComponent < ApplicationComponent
+  attr_reader :title, :body, :icon
+
+  def initialize(title:, body:, icon:)
+    @title = title
+    @body = body
+    @icon = icon
+  end
+end
--- a/app/views/notifications/index.html.erb
+++ b/app/views/notifications/index.html.erb
@@ -4,7 +4,7 @@
     <%= render @notifications %>
   <% else %>
     <div class="mt-8">
-      <%= render "empty_state" %>
+      <%= render EmptyStateComponent.new(title: t(".empty_state.title"), body: t(".empty_state.body"), icon: "icons/outline/bell.svg") %>
     </div>
   <% end %>
   <div class="mt-6 mx-auto">
--- a/test/components/empty_state_component_test.rb
+++ b/test/components/empty_state_component_test.rb
@@ -1,0 +1,14 @@
+require "test_helper"
+
+class EmptyStateComponentTest < ViewComponent::TestCase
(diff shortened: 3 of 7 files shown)


app/components/empty_state_component.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
```
joemasilotti, line 1: Please remove this comment.

app/views/notifications/index.html.erb
```diff
@@ -4,7 +4,7 @@
     <%= render @notifications %>
   <% else %>
     <div class="mt-8">
-      <%= render "empty_state" %>
+      <%= render EmptyStateComponent.new(title: t("notifications.empty_state.title"), body: t("notifications.empty_state.body"), icon: "icons/outline/bell.svg") %>
```
joemasilotti, line 7: Let's move the translations so they aren't so long to access. For example,

```yaml
en:
  notifications:
    index:
      empty_state:
        title:
        body:
```

Then we can use `t(".empty_state.title")` and `t(".empty_state.body")` here.

test/components/empty_state_component_test.rb
```diff
@@ -0,0 +1,14 @@
...
+class EmptyStateComponentTest < ViewComponent::TestCase
+  test "Renders the title, body and icon" do
+    title = "No notifications"
+    body = "Nothing to see"
+
+    render_inline(EmptyStateComponent.new(title: title, body: body, icon: "icons/outline/bell.svg"))
+
+    assert_selector("h3", text: title)
+    assert_selector("p", text: body)
+    assert_selector("svg")
```
joemasilotti, line 12: If you use an SVG from `icons/brands/` they have the `title` attribute set. And you can test them like `test/components/social_link_component_test.rb`.

joemasilotti (changes requested):
Thanks for the PR! I requested a few changes and commented on how you can test the SVG.

app/views/notifications/index.html.erb
jespr: I went with this approach first, but wasn't sure if there was previous history with how it was placed in `en.yml` - will change this!

test/components/empty_state_component_test.rb
jespr: Oh!! Neat

jespr:
@joemasilotti should be good now :)

joemasilotti (approved):

joemasilotti:
Awesome! Thanks for the contribution.

PR #204, opened 2022-01-03
refactor: Developer ux change

jacobdaddario:
Adds in the change to the UI flow for looking at developer profiles as discussed in #197. Now the primary action on the developer page will conditionally change based on the user looking at it. Additionally, the link in the menu under the avatar links to the developer's own profile.

### Pull request checklist

### DIFF
--- a/app/components/developer_primary_action_component.html.erb
+++ b/app/components/developer_primary_action_component.html.erb
@@ -1,11 +1,11 @@
-<% if business? %>
-  <%= link_to new_developer_message_path(@developer), rel: "nofollow", class: "shrink-0 mt-4 md:mt-0 inline-flex items-center px-4 py-2 border border-gray-300 shadow-sm text-base font-medium rounded-md text-gray-700 bg-white hover:bg-gray-50 focus:outline-none focus:ring-2 focus:ring-offset-2 focus:ring-gray-500" do %>
-    <%= inline_svg_tag "icons/solid/briefcase.svg", class: "-ml-1 mr-3 h-5 w-5" %>
-    <%= t("developers.show.hire") %>
-  <% end %>
-<% elsif owner? %>
+<% if owner? %>
   <%= link_to edit_developer_path(@developer), class: "shrink-0 mt-4 md:mt-0 inline-flex items-center px-4 py-2 border border-gray-300 shadow-sm text-base font-medium rounded-md text-gray-700 bg-white hover:bg-gray-50 focus:outline-none focus:ring-2 focus:ring-offset-2 focus:ring-gray-500" do %>
     <%= inline_svg_tag "icons/solid/pencil.svg", class: "-ml-1 mr-3 h-5 w-5" %>
     <%= t("developers.show.edit") %>
+  <% end %>
+<% else %>
+  <%= link_to new_developer_message_path(@developer), rel: "nofollow", class: "shrink-0 mt-4 md:mt-0 inline-flex items-center px-4 py-2 border border-gray-300 shadow-sm text-base font-medium rounded-md text-gray-700 bg-white hover:bg-gray-50 focus:outline-none focus:ring-2 focus:ring-offset-2 focus:ring-gray-500" do %>
+    <%= inline_svg_tag "icons/solid/briefcase.svg", class: "-ml-1 mr-3 h-5 w-5" %>
+    <%= t("developers.show.hire") %>
   <% end %>
 <% end %>
--- a/app/components/developer_primary_action_component.html.erb
+++ b/app/components/developer_primary_action_component.html.erb
@@ -1,0 +1,11 @@
+<% if business? %>
+  <%= link_to new_developer_message_path(@developer), rel: "nofollow", class: "shrink-0 mt-4 md:mt-0 inline-flex items-center px-4 py-2 border border-gray-300 shadow-sm text-base font-medium rounded-md text-gray-700 bg-white hover:bg-gray-50 focus:outline-none focus:ring-2 focus:ring-offset-2 focus:ring-gray-500" do %>
+    <%= inline_svg_tag "icons/solid/briefcase.svg", class: "-ml-1 mr-3 h-5 w-5" %>
+    <%= t("developers.show.hire") %>
+  <% end %>
+<% elsif owner? %>
+  <%= link_to edit_developer_path(@developer), class: "shrink-0 mt-4 md:mt-0 inline-flex items-center px-4 py-2 border border-gray-300 shadow-sm text-base font-medium rounded-md text-gray-700 bg-white hover:bg-gray-50 focus:outline-none focus:ring-2 focus:ring-offset-2 focus:ring-gray-500" do %>
+    <%= inline_svg_tag "icons/solid/pencil.svg", class: "-ml-1 mr-3 h-5 w-5" %>
+    <%= t("developers.show.edit") %>
+  <% end %>
+<% end %>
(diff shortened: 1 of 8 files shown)


app/components/developer_primary_action_component.rb
```diff
@@ -0,0 +1,16 @@
+# frozen_string_literal: true
```
joemasilotti, line 1: Please remove this comment.

app/components/developer_primary_action_component.rb
```diff
@@ -0,0 +1,16 @@
...
+
+class DeveloperPrimaryActionComponent < ApplicationComponent
+  def initialize(user:, developer:)
+    @user = user
+    @developer = developer
+  end
+
+  def owner?
+    @user&.developer == @developer && !@developer.nil?
```
joemasilotti, line 10:
```suggestion
    @user&.developer == @developer && @developer.present?
```

joemasilotti: Does this also need a `#persisted?` check like `#business?`?

app/components/user_menu/signed_in_component.html.erb
```diff
@@ -22,7 +22,9 @@
     <div data-toggle-target="element" class="hidden origin-top-right absolute right-0 z-10 mt-2 w-48 rounded-md shadow-lg py-1 bg-white ring-1 ring-black ring-opacity-5 focus:outline-none" role="menu" aria-orientation="vertical" aria-labelledby="user-menu-button" tabindex="-1">
       <%= render UserMenu::MenuItemComponent.new(t(".get_started"), new_role_path, condition: neither?) %>
       <%= render UserMenu::MenuItemComponent.new(t(".my_business_profile"), new_business_path, condition: business?) %>
-      <%= render UserMenu::MenuItemComponent.new(t(".my_developer_profile"), new_developer_path, condition: developer?) %>
+      <% if developer? %>
+        <%= render UserMenu::MenuItemComponent.new(t(".my_developer_profile"), developer_path(user.developer), condition: developer?) %>
+      <% end %>
```
joemasilotti, line 27: I don't think this change is necessary because of the `condition:` option, right?

app/components/developer_primary_action_component.html.erb
```diff
@@ -0,0 +1,11 @@
+<% if business? %>
+  <%= link_to new_developer_message_path(@developer), rel: "nofollow", class: "shrink-0 mt-4 md:mt-0 inline-flex items-center px-4 py-2 border border-gray-300 shadow-sm text-base font-medium rounded-md text-gray-700 bg-white hover:bg-gray-50 focus:outline-none focus:ring-2 focus:ring-offset-2 focus:ring-gray-500" do %>
+    <%= inline_svg_tag "icons/solid/briefcase.svg", class: "-ml-1 mr-3 h-5 w-5" %>
+    <%= t("developers.show.hire") %>
+  <% end %>
+<% elsif owner? %>
```
joemasilotti, line 6: I think the priority needs to be reversed here. A user can be both a business _and_ developer and when they look at their own profile they should see Edit, not Hire.

joemasilotti: Actually, the check should be `if owner?` and `else`. We want to shot the Hire button even if the user isn't a business (yet).

joemasilotti (changes requested):

jacobdaddario:
For some reason GitHub isn't letting me comment on the code reviews. Did you change the permissions on the repos?

jacobdaddario:
@joemasilotti Had a chance to sit down and look over the code review tonight. Made some of the changes, but there's two that I feel need more discussion. I've left those threads unresolved.

joemasilotti:
Awesome! What would you like to discuss with the two open threads?

jacobdaddario:
I left comments in the threads.

joemasilotti:
Weird, I'm not seeing anything. Can you double check that you submitted (if a review) or refresh and ensure they are still present?

If so, please comment with the line number here, there might be a bug in GitHub.

jacobdaddario:
I'll jus post links.

Here's the first one:
https://github.com/joemasilotti/railsdevs.com/pull/204#discussion_r778531733

joemasilotti:
Oh! It looks like those comments are still pending. I think you need to submit the review in the upper right.

jacobdaddario:
Just did it while eating some lunch. TIL, sorry for that.

app/components/user_menu/signed_in_component.html.erb
```diff
@@ -22,7 +22,9 @@
     <div data-toggle-target="element" class="hidden origin-top-right absolute right-0 z-10 mt-2 w-48 rounded-md shadow-lg py-1 bg-white ring-1 ring-black ring-opacity-5 focus:outline-none" role="menu" aria-orientation="vertical" aria-labelledby="user-menu-button" tabindex="-1">
       <%= render UserMenu::MenuItemComponent.new(t(".get_started"), new_role_path, condition: neither?) %>
       <%= render UserMenu::MenuItemComponent.new(t(".my_business_profile"), new_business_path, condition: business?) %>
-      <%= render UserMenu::MenuItemComponent.new(t(".my_developer_profile"), new_developer_path, condition: developer?) %>
+      <% if developer? %>
```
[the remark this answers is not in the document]
joemasilotti: Ah, good catch. I think we should be able to nil-chain off of `app/components/user_menu/signed_in_component.rb:17` then, right?

`user.developer&.persisted?`

app/components/developer_primary_action_component.rb
joemasilotti: Ah, that makes sense. I'm find without the `#persisted?` check then. Let's apply the `nil` vs. `present?` and wrap this up.

jacobdaddario: I think we can entirely eliminate it for the same reason as `persisted?`. You can check our my recent commit to see what the could look like. Either way I think we're fine. If you don't care one way or another, I'll make your suggested change in the interest of wrapping this up.

test/components/developer_primary_action_component_test.rb
```diff
@@ -0,0 +1,28 @@
...
+  end
+
+  test "should show edit to owner of profile" do
+    user = users(:with_available_profile)
+    render_inline DeveloperPrimaryActionComponent.new(user: user, developer: @developer)
+
+    assert_text "Edit"
+  end
+
+  test "should show message button to business account" do
```
joemasilotti, line 22: There should only be two tests now - owner or not.

app/components/user_menu/signed_in_component.html.erb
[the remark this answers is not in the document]
joemasilotti: Ahhh, I see. It's `url:` key that gets executed before `render?` is hit. Ideally, we would move this inside a block or something, but I'm not sure I have the appetite for that right now. I'm happy to move forward with what you have and create an issue to refactor later.

joemasilotti (changes requested):

joemasilotti:
Just one change request regarding the tests and we are good to go.

jacobdaddario:
Will do. Work is wrapping up for the day, so I had a second to check my phone, but won't be at my computer until a bit later.

jacobdaddario:
@joemasilotti Ready for merge. Sorry this took so much fiddling with. I still have some more to learn about how the interface for GH code reviews work 😅 Hopefully my replies to your code review on other branches has been showing up.

joemasilotti (approved):

joemasilotti:
Merged! Thanks for the PR and great initiative – I hadn't even thought about the weird UX before you brought this up.

PR #209, opened 2022-01-03
Fix: the Date.today usages in test cases

hungmi:
Hi! This PR fixes a really small issue about the `Date.today` usages in test cases.

`Date.today.future?` will return `true` when it's midnight here (GMT +8), since it's based on the machine timezone.
I replaced them with `Date.current` so that these tests will not fail depends on machine settings.

### Pull request checklist


joemasilotti:
Great catch, thanks for the PR!

Issue #200, opened 2022-01-02
Paginate notifications


joemasilotti:
`/notifications` and `read_notifications` can get long, especially for admins, so should be paged.

`/developers` has an infinite scroll example powered by pagy and Turbo which could be replicated for these two views.

Now that #133 is merged in this can be addressed with the new component!

imella:
Hey @joemasilotti! If you don't mind, I can pick this one!

joemasilotti:
Sure, go for it! I'll add it to the [In progress column](https://github.com/joemasilotti/railsdevs.com/projects/6).

PR #210, opened 2022-01-03
add paginitation to notifications

imella:
This PR adds pagination to `/notifications` and `/notifications/read`
pages using the `PagyPaginatorComponent`

Closes: #200 

### Pull request checklist

### DIFF
--- a/app/components/pagy_paginator_component.rb
+++ b/app/components/pagy_paginator_component.rb
@@ -5,7 +5,7 @@ class PagyPaginatorComponent < ApplicationComponent
 
   attr_reader :id, :pagy, :url_array, :container_classes, :view_component
 
-  def initialize(id:, pagy:, url_array:, container_classes:)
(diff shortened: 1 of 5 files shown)


joemasilotti:
Thanks for the PR!

One small change request: it looks like the `"space-y-8"` isn't needed for the `container_classes:` option in the initializer. Can you make `container_classes:` optional by defaulting it to `nil` and remove the call from the two new component renders? For example,

```ruby
# app/views/notifications/index.html.erb:11 and app/views/read_notifications/index.html.erb:5
<%= render PagyPaginatorComponent.new(id: "notifications",
  pagy: @pagy,
  url_array: [:notifications]) do |component| %>

# app/components/pagy_paginator_component.rb:8
def initialize(id:, pagy:, url_array:, container_classes: nil)
```

imella:
@joemasilotti changes made! Should be good now.

joemasilotti:
Looks great, thanks for the quick turnaround and contribution!

PR #212, opened 2022-01-05
feat: New developer notifications emailed to businesses

### LINKED ISSUE #208: Email paying businesses when new developer profiles are added
Paying customers shouldn't have to manually check railsdevs to see if there are any new developer profiles. Instead, they should receive daily or weekly emails linking to all of the new profiles added that day or week, respectively.

A PR addressing this should handle the following:

- [ ] Add a new column to the `Business` model,`new_developer_notifications`, as an enum with values `none`, `daily`, and `weekly`; defaulted to `none`
- [ ] Add an option for this field on `/businesses/:id/edit`
- [ ] Add a job that gathers developer profiles added in the last 24 hours and last 7 days and sends a respective email to paying customers based on their preference
- [ ] Run this job every day at 9am EST
- [ ] Add an unsubscribe link to the bottom of the email that links to editing their business profile
- [ ] Highlight the feature on the `/pricing` page

jacobdaddario:
This is a work-in-progress PR that aims to resolve #208. In order to deploy these changes, the Heroku scheduler add-on will eventually need to be added to the website.

Tracking feature progress:
- [x] Add a new column to the Business model,new_developer_notifications, as an enum with values none, daily, and weekly; defaulted to none
- [x] Add an option for this field on /businesses/:id/edit
- [x] Add a job that gathers developer profiles added in the last 24 hours and last 7 days and sends a respective email to paying customers based on their preference
- [x] Run the daily job every day at 9am EST `bundle exec rake developer_digest:daily`
- [x] Run the ~weekly~ daily job every day at 9am EST `bundle exec rake developer_digest:weekly`
- [x] Add an unsubscribe link to the bottom of the email that links to editing their business profile
- [x] Highlight the feature on the /pricing page

### Pull request checklist


app/models/business.rb
```diff
@@ -1,12 +1,16 @@
...
+  enum :new_developer_notifications, [:no, :daily, :weekly], default: :no,
+    suffix: :developer_notifications
+
   belongs_to :user
   has_many :conversations, -> { visible }
 
   validates :name, presence: true
   validates :company, presence: true
   validates :bio, presence: true
+  validates :new_developer_notifications, presence: true
```
jacobdaddario, line 13: I think we can avoid an inclusion validation here. Originally, I tried adding one, but it gets weird with how enumerables work. Additionally, the enumerable will through an argument air if it's given weird. Do we need to do something to catch those errors?

app/views/business_mailer/new_developer_profile.html.erb
```diff
@@ -0,0 +1,10 @@
+<h1>New developers at Railsdevs.com!</h1>
```
jacobdaddario, line 1: Kept the styling simple here like you did in the other mailers.

config/locales/en.yml
```diff
@@ -208,7 +214,7 @@ en:
         '3': Skip job descriptions and formal postings.
         '4': Talk directly to candidates.
         '5': Unlock private developer information.
-        '6': Hire as much or as little as you need.
+        '6': Receive email digests of new candidates.
```
jacobdaddario, line 217: I didn't want to disrupt the 2x4 design, so I opted to just replace one of the lines. Not sure if this is preferred by you.

lib/tasks/developer_digest.rake
```diff
@@ -0,0 +1,37 @@
+namespace :developer_digest do
```
jacobdaddario, line 1: I followed your lead and omitted tests for the mailer itself. I'm not sure if we want them for these rake tasks. I figured I'd ask before writing them, since after some reading, it looked like writing rake task tests is a bit complicated.

jacobdaddario (commented):
@joemasilotti Ready to go. Just have a few things I wanted to make note of in advance of your review. Additionally, the only thing that isn't finished is scheduling the rake tasks. Unless you give me access to Heroku, you'll have to do it. It's a pretty simple configuration with the [scheduler add-on](https://devcenter.heroku.com/articles/scheduler#defining-tasks). You should be able to set the appropriate frequency for each task using that tool as well.

app/views/businesses/_form.html.erb
```diff
@@ -59,6 +59,40 @@
       </div>
     </div>
 
+    <div class="bg-white shadow mt-8 px-4 py-5 lg:rounded-lg sm:p-6">
+      <div class="md:grid md:grid-cols-3 md:gap-6">
+        <div class="md:col-span-1">
+          <h3 class="text-lg font-medium leading-6 text-gray-900"><%= t(".notifications.title") %></h3>
+          <p class="mt-1 text-sm text-gray-500"><%= t(".notifications.help_text") %></p>
```
joemasilotti, line 66: Let's drop the help text until we have another field in this section.

app/views/businesses/_form.html.erb
```diff
@@ -59,6 +59,40 @@
...
+        </div>
+        <div class="mt-5 md:mt-0 md:col-span-2">
+          <div>
+            <%= form.label :new_developer_notifications, class: "text-base font-medium text-gray-900" %>
+            <p class="text-sm leading-5 text-gray-500"><%= t(".new_developer_notifications.help_text") %></p>
+            <fieldset class="mt-4">
+              <legend class="sr-only"><%= t(".new_developer_notifications.legend") %></legend>
+              <div class="space-y-4">
+                <div class="flex items-center">
+                  <%= form.radio_button :new_developer_notifications, "no", class: "focus:ring-gray-500 h-4 w-4 text-gray-600 border-gray-300" %>
```
joemasilotti, line 76:
```suggestion
                  <%= form.radio_button :new_developer_notifications, :no, class: "focus:ring-gray-500 h-4 w-4 text-gray-600 border-gray-300" %>
```

app/views/businesses/_form.html.erb
```diff
@@ -59,6 +59,40 @@
...
+                  <%= form.label :new_developer_notifications_no, "None", class: "ml-3 block text-sm font-medium text-gray-700" %>
```
joemasilotti, line 77:
```suggestion
                  <%= form.label :new_developer_notifications_no, "Never", class: "ml-3 block text-sm font-medium text-gray-700" %>
```

Also, this (and the next two) should be localized.

app/views/businesses/_form.html.erb
joemasilotti: Same applies for the next two, also.

lib/tasks/developer_digest.rake
joemasilotti: I'm fine omitting tests for the mailer, especially because you included the previews.

But I think because this rake task has some real logic we want some sort of tests for it. I recommend pulling out all of the code into a model and testing _that_. Then the rake task can call a single line and we don't have to worry about testing it.

lib/tasks/developer_digest.rake
```diff
@@ -0,0 +1,37 @@
...
+  desc "email subscribed businesses about new developers in the last day"
```
joemasilotti, line 2:
```suggestion
  desc "Email subscribed businesses about new developers in the last day"
```

lib/tasks/developer_digest.rake
```diff
@@ -0,0 +1,37 @@
...
+  task daily: :environment do
+    businesses = Business.daily_developer_notifications
+    developers = Developer.where("created_at > '#{1.day.ago}'")
```
joemasilotti, line 5:
```suggestion
    developers = Developer.where(created_at: Date.yesterday..Date.current)
```

This creates a concrete time span instead of basing it off of when the task was actually run. If the time the task run changes we could miss a developer otherwise.

lib/tasks/developer_digest.rake
```diff
@@ -0,0 +1,37 @@
...
+      puts "Sending daily digests to #{total_emails} businesses:"
+
+      businesses.each_with_index do |business, i|
+        puts "Progress - #{((i + 1.0) / total_emails) * 100}%"
+        BusinessMailer.with(business: business, developers: developers).new_developer_profile.deliver_now
+      end
+    end
+  end
+
+  desc "email subscribed businesses about new developer s in the last week"
```
joemasilotti, line 20:
```suggestion
  desc "Email subscribed businesses about new developers in the last week"
```

lib/tasks/developer_digest.rake
```diff
@@ -0,0 +1,37 @@
...
+  task weekly: :environment do
+    businesses = Business.weekly_developer_notifications
+    developers = Developer.where("created_at > '#{1.week.ago}'")
```
joemasilotti, line 23:
```suggestion
    developers = Developer.where(created_at: (Date.yesterday - 7.days)..Date.current)
```

app/views/business_mailer/new_developer_profile.html.erb
```diff
@@ -0,0 +1,10 @@
...
+
```
joemasilotti, line 2: Let's remove this intro, the subject says the same thing.

app/views/business_mailer/new_developer_profile.html.erb
```diff
@@ -0,0 +1,10 @@
...
+<p><%= @business.name %>,</p>
+<p>Here's your digest of new developers available for hire at Railsdevs.com:</p>
+<hr />
+<% @developers.each do |dev| %>
+  <p><%= link_to dev.name, dev %>: <%= dev.hero %></p>
```
joemasilotti, line 7:
```suggestion
<% @developers.each do |developer| %>
  <p><%= link_to developer.name, developer_url(developer) %>: <%= developer.hero %></p>
```

app/views/business_mailer/new_developer_profile.text.erb
```diff
@@ -0,0 +1,15 @@
...
+
+<%= "-" * 20 %>
+
+<% @developers.each do |dev| %>
+<%= dev.name %>: <%= dev.hero %>
+
+<%= "-" * 20 %>
+
+<% end %>
+Unsubscribe from these emails by editing your preferences in your business profile.
```
joemasilotti, line 15: Same suggestions as the HTML version.

app/views/business_mailer/new_developer_profile.text.erb
```diff
@@ -0,0 +1,15 @@
+New developers at Railsdevs.com!
+
+<%= @business.name %>,
+
+Here's your digest of new developers available for hire at Railsdevs.com:
+
+<%= "-" * 20 %>
+
+<% @developers.each do |dev| %>
+<%= dev.name %>: <%= dev.hero %>
```
joemasilotti, line 10: We can still link to the developer.

```suggestion
<%= developer_url(developer) %>
<%= dev.name %>: <%= dev.hero %>
```

app/views/business_mailer/new_developer_profile.html.erb
```diff
@@ -0,0 +1,10 @@
+<h1>New developers at Railsdevs.com!</h1>
+
+<p><%= @business.name %>,</p>
+<p>Here's your digest of new developers available for hire at Railsdevs.com:</p>
```
joemasilotti, line 4:
```suggestion
<p>Here's your digest of new developer profiles on railsdevs:</p>
```

The new folks technically might not be available right now.

app/views/business_mailer/new_developer_profile.text.erb
joemasilotti, line 15:
```suggestion
Unsubscribe from these emails by editing your preferences in your business profile.
<%= edit_business_url(@business) %>
```

app/views/business_mailer/new_developer_profile.html.erb
```diff
@@ -0,0 +1,10 @@
...
+<hr />
+<% @developers.each do |dev| %>
+  <p><%= link_to dev.name, dev %>: <%= dev.hero %></p>
+  <hr />
+<% end %>
```
joemasilotti, line 9: The horizontal rules are a little aggressive without any CSS. Let's change them to a `<ul>` and `<li>` instead.

joemasilotti (changes requested):
Amazing work, Jacob! Thanks so much for this.

I left a _ton_ of comments, but they are mostly tiny nitpicky suggestions that you can directly accept (assuming they work).

The big bit of work I requested is moving the mailing logic out of the rake task and into a new model that can be tested directly. And a bit of changes to the actual email template.

Let me know if you'd like me to handle any of this and thanks again!

app/views/business_mailer/new_developer_profile.text.erb
jacobdaddario: You can't hyperlink in plain text emails. That will just render out HTML that does nothing.

app/views/business_mailer/new_developer_profile.text.erb
jacobdaddario: See below. Plain text emails do not allow hyperlinking.

jacobdaddario:
Looked over things really briefly. I think most of what you're saying makes sense. Like you said, the model is the main thing, and the rest come primarily down to copy and code style.

Only thing I wanted to make sure that you were aware of is the fact that you can't hyperlink in plain text emails. Other than that I can implement all of the suggested changes. I particularly like the bit about extracting the rake task logic out into a model. Good thinking with that.

I want to stick to my personal commitment about another project, so I'll be spending time on that this weekend when I have time, so I actually won't get around to this until next week. Up to you if you want to push this across this finish line. I'm fine either way.

joemasilotti:
> Only thing I wanted to make sure that you were aware of is the fact that you can't hyperlink in plain text emails.

That's correct, but you _can_ add the URL for someone to copy/paste in their browser. I'm doing that in the `AdminMailer` text formats.

> I want to stick to my personal commitment about another project, so I'll be spending time on that this weekend when I have time, so I actually won't get around to this until next week. Up to you if you want to push this across this finish line. I'm fine either way.

Sounds great, thanks for the heads up! I might try to pick up bits of this here and there and will make sure to push whenever I make changes.

app/views/business_mailer/new_developer_profile.text.erb
joemasilotti: I mentioned this in my comment, but outputting the URL still has value because someone can copy/paste it into their browser.

joemasilotti:
@jacobdaddario, a heads up that I'm directly testing a rake task in PR #225 and it seems to be much less work than it used to. Check out [this test](https://github.com/joemasilotti/railsdevs.com/blob/cc36230c3669360b989b94fb22227a02be063f4d/test/lib/tasks/developers_utc_offset_task_test.rb) for an example, which tests [this task](https://github.com/joemasilotti/railsdevs.com/blob/cc36230c3669360b989b94fb22227a02be063f4d/lib/tasks/developers.rake).

I'm happy to go with this approach or a custom object; your call!

jacobdaddario:
I think I'm gonna go with a little command pattern/service object. Seems very suitable here. Any strong opinions on where that goes? I'm thinking about putting it under a Business namespace in the models folder.

joemasilotti:
Sounds great on both accounts! A PORO that gets called from the rake task sounds 👌

joemasilotti:
I applied all of the changes that I requested that were still unresolved. Most of the stuff was around copy/design in the emails, the enums, and naming. I think all that's left is is testing the rake tasks!

One thing that I overlooked is we need a paywall or CTA in the form for folks that don't have an active business subscription. I'm not sure what the best design is for that just yet, but wanted to call it out. We can merge this in with a feature flag if need be.

jacobdaddario:
@joemasilotti Thank you! I really appreciate it. By the way, I did some searching about the timestamp precision thing again, and I found this: https://github.com/rails/rails/blob/2ea5268cd2994a8d4628fbdb02c64e4d25549c7f/activerecord/lib/active_record/migration/compatibility.rb#L220

This was the only compatibility change for timestamps. It changed in Rails 6 which is why we're seeing the backwards compatibility change for Rails 5.2. Is there a reason you think Rails 7 shouldn't add precision by default? From what I can see, it never changed back. At least as evidenced in this compatibility file.

joemasilotti:
Weird! Thanks for diving in. I'm using PostgreSQL 14.0, which is one minor release behind latest.

Are you saying that the latest versions _remove_ the precision from db/schema or _add_ them in? Because there are other spots in the schema where precisions are explicitly set. Ideally, I'd like to keep them all consistent.

jacobdaddario:
I'm saying the latest version includes the precisions by default. You can see that the change kicked in in version 6.0. We can confirm that by seeing they added a backwards compatibility method for version 5.2 and below in that compatibility file.

The second file linked in the edit shows where the current definition for timestamps is. If you follow the conditional to see if it supports it, it shows that the postgres adapter supports it for all versions.

joemasilotti:
Ahhh gotcha! Then is it safe to remove all of the precision options in the schema?

jacobdaddario:
My thought process was that they'd generate in the schema by default, since that's the behavior for all Postgres versions since Rail 6.0, as far as I can tell.

joemasilotti:
Interesting. And if you run `rake db:migrate:reset` does it add all the instances of `precision:` back in, remove them all, or some weird mix?

jacobdaddario:
@joemasilotti It adds all of the datetime precision options back in.

joemasilotti:
Oh man, for real? And then if you re-run the most recent migration it... removes a few instances of `precision:`?

jacobdaddario:
I just pulled from main and ran it. I don't see anything different. Most recent commit seems to affect timezone, not precision. What makes you think it'd be different?

joemasilotti:
> What makes you think it'd be different?

Didn't this whole conversation spark from this PR removing a could instances of `precision:` in the schema? Or am I going crazy today.

jacobdaddario:
> > What makes you think it'd be different?
> 
> Didn't this whole conversation spark from this PR removing a could instances of `precision:` in the schema? Or am I going crazy today.

Oh sorry for the confusion. I'm remarking about the problem in general.

joemasilotti:
Gotcha. What do you propose moving forward? I think you have more knowledge around this then I do at this point.

jacobdaddario:
I'm in favor of going with the default. I just don't get why your machine isn't producing that result anyways.

Also, has less time to work on this today than I thought I would. Pushed my changes, but it's unfinished. Got hung up on something and I'm too tired to debug it.

Considering reverting back to not having a command object and just testing the rake tasks directly.

jacobdaddario:
@joemasilotti Alright, I think I've taken this as far as I can. There's a couple of things I want to mention before leaving this in your hands.

1. I've soft feature flagged email preferences for the time being. I imagine we'll need a separate PR to figure out how we want to gate premium content from companies that have not yet paid for a subscription. Obviously the field is still reachable, but it won't appear on the business form for now.
2. You'll need to set this up in Heroku's scheduler to run at the desired time whenever we complete this full feature.
3. I see you changed the query for developers into a date range as a matter of code quality. I'm not sure why you switched to `Date.current` though. This was breaking tests on my machine because `Date.current` is not timezone aware unlike `Time.now`. My PSQL database saves thing in UTC time, and without the timezone aware `Time.now`, developers were actually being created before the current time on my system. On a production environment this won't be a problem, but if there's no particular reason you used `Date.current` over `Time.now`, it might be best to keep the timezone aware methods.
4. Regarding the `datetime` precision thing. I'm fairly confident that default behavior should be having `precision: 6`, but it's up to you how you handle this. I'll say anecdotally, after resolving merge conflicts on this branch, whenever I tried to revert the `precision` values added to the schema, I'd get database integrity issues on my end. Rails would keep telling me I had pending migrations and wouldn't let me run tests. Consequently, I've committed he precision values in this PR.

After I get #220 merged, I'll probably take a break from committing for a while. I'll be honest that I struggled with this branch for some reason, and I don't want to take away from your development time by making do code reviews. Plus, it'll let me focus on my personal project that I've been neglecting.

joemasilotti:
1. Awesome, this is perfect for now. I'm going to change it to a "real" `Feature.rb` flag so we don't have the environment check in the view. From there, I'm hoping to add a Subscribe button to `/developers`.
2. 👍
3. Hmm, I that's backwards. Both [`Date.current`](https://github.com/rails/rails/blob/main/activesupport/lib/active_support/core_ext/date/calculations.rb#L48-L50) and [`Time.current`](https://apidock.com/rails/Time/current/class) take into account time zone. `Time.now` does not. Could they be failing because the created developers were "in between" UTC's midnight and your current time zone?
4. Gotcha, sounds good. I'll keep what you have and if I run into a different diff I'll address it later.

Outside of hearing your feedback on number 3. I can take this over the finish line. Thanks again for the contribution, this was a lot of work and I really appreciate it!

jacobdaddario:
Oh shoot, you're totally right about `Date.current`. I was in a rush last night, and I must have skimmed `TimeZone.now` as `Time.now`.

I'd hate to quit on you, so I can investigate this and the feature flag. Whatever you feel is most time efficient. I want to be respectful to your codebase and to your time.

joemasilotti:
All good! Don't ever feel obligated to work on this, every line of code you write is one less for me :)

If you get around to making changes go for it. If not, I'm hoping to dive into some work later this week and can take this over the finish line.

joemasilotti:
Holy crap, dates and time zones suuuuuuuuuuck.

I ended up reworking the logic of the mailer around dates. Using `1.day.ago` and friends uses the current time zone – that combined with `#beginning_of_day` and `#end_of_day` makes it a bit easier (IMHO) to reason about what days we are talking about. I also dropped the enum-ish param for two separate methods.

Thanks for the work you put in on this! I'm going to sleep on the PR, review it tomorrow, and hopefully merge this weekend if all still looks good.

joemasilotti (approved):

joemasilotti:
And we're live! 🎉

Wow that was a big one. That definitely snuck up on me, I did not expect this to be such a huge commitment. Thanks again for doing the majority of the work on this, @jacobdaddario! I always appreciate it :)

PR #221, opened 2022-01-06
Add basic configuration for system tests

salochara:
Hey there! 👋 

I've added the basic configuration for system tests. 
As for the pagy feature test, I'm honestly not familiar with pagy and I'm not sure what/how to test. 

Also, I've not added 
`bundle exec rails test:system` or even change it to `bundle exec rails test:all` in the `bin/check` script since things will eventually get slow 😅  haha. 

Let me know what you think about this!  

Cheers!

### Pull request checklist

### DIFF
--- a/README.md
+++ b/README.md
@@ -92,4 +92,6 @@ Application monitoring is powered by [Scout APM](https://scoutapm.com). This hel
 
 ## Testing
 
-Run `rails test` to run unit/integration tests.
+* Run `rails test` to run unit/integration tests.
+* Run `rails test:system` to run system tests, using `headless_chrome`.
+* Run `HEADFUL=1 rails test:system` to run system tests, using `headful_chrome`.
--- a/test/application_system_test_case.rb
+++ b/test/application_system_test_case.rb
@@ -1,0 +1,19 @@
+require "test_helper"
+require "webdrivers"
+
+Capybara.register_driver :headless_chrome do |app|
+  options = ::Selenium::WebDriver::Chrome::Options.new
+  options.add_argument("--headless") unless ActiveModel::Type::Boolean.new.cast(ENV["HEADFUL"])
+  options.add_argument("--window-size=1920,1080")
+
+  client = Selenium::WebDriver::Remote::Http::Default.new
+  client.read_timeout = 240
+
+  Capybara::Selenium::Driver.new(app, browser: :chrome, capabilities: [options], http_client: client)
+end
+
+Capybara.javascript_driver = :headless_chrome
+
+class ApplicationSystemTestCase < ActionDispatch::SystemTestCase
+  driven_by :headless_chrome
+end
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -113,3 +113,4 @@ jobs:
           bin/rails db:create
           bin/rails db:schema:load
           bin/rails test --fail-fast
+          bin/rails test:system
--- a/Gemfile
+++ b/Gemfile
@@ -32,6 +32,8 @@ end
 
 group :test do
   gem "capybara", "~> 3.36"
+  gem "selenium-webdriver"
+  gem "webdrivers"
 end
 
 gem "aws-sdk-s3", "~> 1", require: false
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -111,6 +111,7 @@ GEM
       rack-test (>= 0.6.3)
       regexp_parser (>= 1.5, < 3.0)
       xpath (~> 3.2)
+    childprocess (4.1.0)
     classy-yaml (0.7.0)
       actionpack (>= 6)
       activesupport (>= 6)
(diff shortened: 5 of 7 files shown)


joemasilotti:
This is awesome, thank you for the PR! I'd love just one test, even if it is a smoke test, to make sure everything is configured to work with JavaScript.

A good example could be opening/closing the user menu after signing inon desktop and making sure the "Get started" text appears!

salochara:
Hey 👋 
Not sure which text are you referring to 🤔 . Can you post here an image/gif please? 🙏 

I'll work on this during the weekend as well!

joemasilotti:
If you sign up for an account but don't add a developer or business profile you have "Get started" in your user dropdown. I'm happy to test anything in that drop down, though. Honestly, "Sign out" is probably the best bet because it is always there!

image

salochara:
Makes sense! Thanks for the clarification!

salochara:
Hey there! 👋 
I've added a (quite simple) login test and resized the window size for system tests.

Let me know what you think!

README.md
```diff
@@ -92,4 +92,7 @@ Application monitoring is powered by [Scout APM](https://scoutapm.com). This hel
 
 ## Testing
 
-Run `rails test` to run unit/integration tests.
+* Run `rails test` to run unit/integration tests.
+* Run `rails test:system` to run system tests, using `headless_chrome`.
+* Run `HEADFUL=1 rails test:system` to run system tests, using `headful_chrome`.
+
```
joemasilotti, line 98: Please remove this blank line.

test/application_system_test_case.rb
```diff
@@ -0,0 +1,19 @@
+require "test_helper"
+require "webdrivers"
+
+Capybara.register_driver :headless_chrome do |app|
+  options = ::Selenium::WebDriver::Chrome::Options.new
+  options.add_argument("--headless") unless ENV["HEADFUL"].in?(%w[y t 1 yes true])
```
joemasilotti, line 6: Can we use `ActiveModel::Type::Boolean.new.cast(ENV["HEADFUL"])` here?

test/application_system_test_case.rb
```diff
@@ -0,0 +1,19 @@
...
+
+  client = Selenium::WebDriver::Remote::Http::Default.new
+  client.read_timeout = 240
+
+  Capybara::Selenium::Driver.new(app, browser: :chrome, capabilities: [options], http_client: client)
+end
+
+Capybara.javascript_driver = :headless_chrome
+
+class ApplicationSystemTestCase < ActionDispatch::SystemTestCase
```
joemasilotti, line 17: Let's move this to its own file. We can import _this_ file from the new one, which means we shouldn't have to import _this_ file from new test cases. Right?

joemasilotti (changes requested):
Looks great! I made one suggestion/question and one change request.

Outside of that, two more things:

1. Change `bin/check` to run all tests, including system tests.
2. Update CI to run system tests after unit/integration.

test/application_system_test_case.rb
salochara: Yep. Done!

test/application_system_test_case.rb
salochara: Hmm. Not sure what you mean exactly! 🤔 
I guess it'd be better if you make this change!

salochara:
Hey 👋 
Thanks for your comments!

- Change bin/check to run all tests, including system tests. **-> Done!**
- Update CI to run system tests after unit/integration. **-> Done!**

Regarding the javascript disabled question.. I haven't found the way to use a [capybara driver ](https://github.com/teamcapybara/capybara#drivers) with javascript disabled...
From what I see by running the tests in HEADFUL and with the screenshots, the menu is indeed visible
failures_test_users_can_log_in copy


Let me know whay you think!

test/application_system_test_case.rb
```diff
@@ -0,0 +1,19 @@
+require "test_helper"
+require "webdrivers"
+
+Capybara.register_driver :headless_chrome do |app|
+  options = ::Selenium::WebDriver::Chrome::Options.new
+  options.add_argument("--headless") unless ActiveModel::Type::Boolean.new.cast(ENV["HEADFUL"]).in?(%w[y t 1 yes true])
```
joemasilotti, line 6:
```suggestion
  options.add_argument("--headless") unless ActiveModel::Type::Boolean.new.cast(ENV["HEADFUL"])
```

This returns `true` or `false`, so we don't need to check if it is included in the given options anymore.

test/application_system_test_case.rb
joemasilotti: Ah, never mind actually. I didn't realize this was the [standard practice with minitest](https://guides.rubyonrails.org/testing.html#system-testing)!

joemasilotti (changes requested):
Looks great outside of one tiny change request!

test/application_system_test_case.rb
salochara: Yep. Although not so sure how would we run tests locally with the HEADFUL option set to true. 
I added the options for running the tests like this`HEADFUL=1 rails test test/system/login_test.rb`
If I only run it with  `HEADFUL rails test test/system/login_test.rb` it returns an error. 
Is there a way for running them? What piece am I missing here haha? 🤔

joemasilotti: If I apply my suggested change (above) I can run the tests "headful" (viewing Chrome) with `HEADFUL=1 rails test:system`.  `HEADFUL=true rails test:system`, among other "truthful" options, also work. `rails test:system`, `HEADFUL=false rails test:system`, and `HEADFUL=0 rails test:system` all run the tests headless.

joemasilotti:
Thanks for the contribution - merging in now!

Issue #203, opened 2022-01-02
Basic developer filtering


joemasilotti:
This request comes from [two](https://github.com/joemasilotti/railsdevs.com/discussions/61) [discussions](https://github.com/joemasilotti/railsdevs.com/discussions/174#discussioncomment-1850542) and an email request from a business.

To start, filtering by availability and sorting by when the profile was added should be good enough. Then I'd like to move into compensation, location, available hours per week, and more.

joemasilotti:
[hirethePIVOT.com](https://hirethepivot.com), a fork of railsdevs, is [implementing this](https://github.com/bencgreenberg/hirethepivot.com/pull/42#pullrequestreview-845237180) based on my [initial PR](#177). Here's a screenshot that we could use to model ours after. I don't think we need/want search (yet) but the filters look good to me!

![67DjTWwr jpg-large](https://user-images.githubusercontent.com/2092156/148320733-720f36fb-840a-4d7d-87e1-b15646ca3fd4.jpeg)

joemasilotti:
GitHub's Trending page is also a great example of some good looking filtering.

<img width="1490" alt="image" src="https://user-images.githubusercontent.com/2092156/148432762-8e9d0ebc-3dae-4bba-a74c-1d9796d6d7ac.png">

PR #222, opened 2022-01-06
handle disabled state in form buttons

imella:
This PR adds the `SaveButtonComponent` which handles the rendering of a button to be used on forms. It displays the text of the button and the element to be shown when the button is disabled, which is handled automatically by Turbo when submitting the form.

The change was made in the message, developer and business forms. All devise forms were left out because they still don't play nicely with Turbo.

Screen Shot 2022-01-06 at 20 48 45

### Pull request checklist


jacobdaddario:
Yay! Thank you for handling this.

app/assets/stylesheets/application.tailwind.css
```diff
@@ -4,3 +4,17 @@
 @import "tailwindcss/overrides";
 
 @import "components/pagy";
+
+button .show-when-disabled {
```
jacobdaddario, line 8: I think we might be able to handle this just with Tailwind psuedo-classes. For instance `disabled:show`. In fact, we might be able to eliminate the conditional from the component too. [Sean gives a nice example in his PR for disabling in Turbo.](https://github.com/hotwired/turbo/pull/386#issue-993156012)

jacobdaddario (commented):
Good catch upgrading `turbo-rails`! That includes the new behavior that disables form submitters by default. Speaking of Sean's disable PR, I noticed one thing that might be worth taking into consideration.

app/assets/stylesheets/application.tailwind.css
joemasilotti: +1 on giving Tailwind pseudo-classes a go, if possible!

app/views/businesses/_form.html.erb
```diff
@@ -60,7 +60,9 @@
     </div>
 
     <div class="flex justify-end mt-8 mb-16 mr-4 sm:mr-6 lg:mr-0">
-      <%= form.submit t(".save"), class: "ml-3 inline-flex justify-center py-2 px-4 border border-transparent shadow-sm text-sm font-medium rounded-md text-white bg-gray-600 hover:bg-gray-700 focus:outline-none focus:ring-2 focus:ring-offset-2 focus:ring-gray-500" %>
+      <%= form.button class: "ml-3 inline-flex justify-center align-center py-2 px-4 border border-transparent shadow-sm text-sm font-medium rounded-md text-white bg-gray-600 hover:bg-gray-700 focus:outline-none focus:ring-2 focus:ring-offset-2 focus:ring-gray-500" do %>
+        <%= render SaveButtonComponent.new(text: t(".save"), disable_with: t(".processing")) %>
+      <% end %>
```
joemasilotti, line 65: This diff is making me wonder if we are making the wrong abstraction here. Should the component instead be the _entire_ button, not just the stuff inside? It could take in the form to chain the helper. That would make these really nice to use because _all_ of the CSS would be in there. If you need to add `ml-3` or something very specific we could pass in an optional parameter.

config/locales/en.yml
```diff
@@ -108,6 +109,7 @@ en:
       preferred_compensation:
         help_text: What is your preferred annual salary and/or hourly rate?
         title: Preferred compensation
+      processing: Processing
```
joemasilotti, line 112: I'm not a huge fan of processing, it's a tad too technical, in my opinion. I'd love for it to say `Creating...` or `Updating...` depending on if we are creating or updating the record. If that's not possible or not worth the effort, `Submitting...` or `Saving...` could work!

app/components/save_button_component.rb
```diff
@@ -0,0 +1,8 @@
+class SaveButtonComponent < ApplicationComponent
```
joemasilotti, line 1: How about `SubmitButtonComponent`?

joemasilotti:
Loving this PR; nice work and thanks for the contribution! I made a couple of comments/questions in the code, let me know what you think.

app/views/businesses/_form.html.erb
jacobdaddario: Unfortunately a big known issue with View Component right now is that it doesn't play nice with Rails' form builder. We'd have to make a custom form builder to achieve this.

joemasilotti: What do you mean? We can't pass in the `form` object?

jacobdaddario: I'm not entirely sure about the specifics. I'm just familiar with some issues I've read on the repo and [this note on the package's website](https://viewcomponent.org/known_issues.html#form_for-compatibility).

joemasilotti: Ah, I see. https://github.com/github/view_component/discussions/420

The main issue is passing in the `form` couples the component with the form builder object. On the surface that seems fine, but if we ever need to storybook or use them outside of a form context it starts to fall apart.

I'm not sure what the best approach is, here, to be honest. Maybe it is to keep what we have and not abstract any father?

joemasilotti: OK, messing around with some code for #203, turns out we don't need a reference to the form at all. We can use `submit_tag` or `button_tag type: :submit` and it will submit the `<form>` that it is nested under.

It doesn't solve the whole "ViewComponent + FormBuilder" but we can punt on that for now!

app/components/save_button_component.rb
imella: I like it, it covers more cases than just saving a record.

config/locales/en.yml
imella: The first option sounds way better, so I'll go for that

app/views/businesses/_form.html.erb
imella: > This diff is making me wonder if we are making the wrong abstraction here. Should the component instead be the _entire_ button, not just the stuff inside? It could take in the form to chain the helper. That would make these really nice to use because _all_ of the CSS would be in there. If you need to add `ml-3` or something very specific we could pass in an optional parameter.

I had the same question while doing it, but I just went for an approach focused more on the content and not coupled with the form itself. But using the `button_tag` helper sounds good and keeps the component clean.

imella:
> Loving this PR; nice work and thanks for the contribution! I made a couple of comments/questions in the code, let me know what you think.

Happy to contribute!

imella:
All changes made, should be good for review!

app/components/submit_button_component.rb
```diff
@@ -0,0 +1,15 @@
+class SubmitButtonComponent < ApplicationComponent
+  attr_reader :text, :disable_with, :button_clasess
+
+  def initialize(text:, disable_with: nil, button_classes: nil)
+    @text = text
+    @disable_with = disable_with
+    @button_clasess = button_clasess || default_button_clasess
```
joemasilotti, line 7: I think `button_classes` should be additive, not destructive. Said another way, it should append `button_classes` to `default_button_classes`. I'm also not sure I want `ml-3` as the default. But you were in the code – was every submit button using that class?

app/components/submit_button_component.rb
```diff
@@ -0,0 +1,15 @@
...
+  end
+
+  def default_button_clasess
+    "ml-3 inline-flex justify-center align-center py-2 px-4 border border-transparent"\
+    "shadow-sm text-sm font-medium rounded-md text-white bg-gray-600 cursor-pointer"\
+    "hover:bg-gray-700 focus:outline-none focus:ring-2 focus:ring-offset-2 focus:ring-gray-500"
```
joemasilotti, line 13: If we move to `button_classes:` being additive then this can live in the HTML file!

app/views/businesses/_form.html.erb
```diff
@@ -60,7 +60,7 @@
     </div>
 
     <div class="flex justify-end mt-8 mb-16 mr-4 sm:mr-6 lg:mr-0">
-      <%= form.submit t(".save"), class: "ml-3 inline-flex justify-center py-2 px-4 border border-transparent shadow-sm text-sm font-medium rounded-md text-white bg-gray-600 hover:bg-gray-700 focus:outline-none focus:ring-2 focus:ring-offset-2 focus:ring-gray-500" %>
+      <%= render SubmitButtonComponent.new(text: t(".save"), disable_with: business.persisted? ? t(".updating") : t(".creating")) %>
```
joemasilotti, line 63: Ohhh, nice!

app/views/messages/_form.html.erb
```diff
@@ -13,8 +13,6 @@
 
 <div class="absolute bottom-0 inset-x-0 pl-3 pr-2 py-2 flex justify-end">
   <div class="shrink-0">
-    <button type="submit" class="inline-flex items-center px-4 py-2 border border-transparent text-sm font-medium rounded-md shadow-sm text-white bg-gray-600 hover:bg-gray-700 focus:outline-none focus:ring-2 focus:ring-offset-2 focus:ring-gray-500">
```
joemasilotti, line 16: Ah, see, this one doesn't use `ml-3`. Even though it might not make a visual difference here, I'd rather pull it out and have the others pass it in.

app/views/businesses/_form.html.erb
```diff
@@ -5,7 +5,7 @@
     <%= t(".your_business_profile") %>
   </h1>
 
-  <%= form_with model: business, html: {data: {turbo: false}} do |form| %>
```
joemasilotti, line 8: Oh, good catch!

app/components/submit_button_component.html.erb
```diff
@@ -0,0 +1,9 @@
+<%= button_tag type: :submit, class: "group #{button_clasess}" do %>
+  <span class="show group-disabled:hidden"><%= text %></span>
```
joemasilotti, line 2: `group-disabled` – 👌

joemasilotti (changes requested):
Awesome work here – I love that we were able to get rid of the custom CSS! I made a few requests and a few comments.

app/views/messages/_form.html.erb
imella: Good catch! I didn't noticed that.

app/components/submit_button_component.rb
imella: Got it!

imella:
Suggestions added! Just noticed that `standardrb` automatically made changes for the new 3.1 hash syntax, nice!

joemasilotti (commented):
Looks great – thanks again for the contribution!

PR #223, opened 2022-01-07
Developer validation

soc-man:
### Pull request checklist


db/migrate/20211203172235_add_timezone_to_developers.rb
```diff
@@ -1,5 +1,5 @@
 class AddTimezoneToDevelopers < ActiveRecord::Migration[7.0]
   def change
-    add_column :developers, :time_zone, :string
+    add_column :developers, :primary_time_zone, :string
```
joemasilotti, line 3: We can't change an existing migration because it has already been run. Instead, we would need to create a new one that renames the column.

However, I think we can change the label in the form instead of the database column.

app/models/concerns/avatarable.rb
```diff
@@ -6,5 +6,15 @@ module Avatarable
 
     validates :avatar, content_type: ["image/png", "image/jpeg", "image/jpg"],
       max_file_size: 2.megabytes
+
+    validate :avatar_validation, on: :create
```
joemasilotti, line 10: Interesting approach! Is there not a built in validator to do this?

If not, we should create a custom validator. Check out two lines above which uses `ContentTypeValidator`.

joemasilotti (changes requested):
Thanks for the contribution! I requested two changes then I can help you with the tests.

db/migrate/20211203172235_add_timezone_to_developers.rb
soc-man: Oh I see, that would have been disastrous! Done a bit of reading up and it's been a good learning experience(thankfully not expensive). Okay let me work on that :)

app/models/concerns/avatarable.rb
soc-man: Looked this up and this is what I found.However, I can do more research to see if there's a built-in validator and how to make custom validations. Its been a fun experience!

db/migrate/20211203172235_add_timezone_to_developers.rb
soc-man: Btw just for clarification 
I was thinking of generating a new migration to change the column name, would you prefer that or to change the label in the form?

joemasilotti: Let's change the label in the form, for now.

app/models/concerns/avatarable.rb
soc-man: Hey, added two new commits. On the console they seem to be passing the validation but fails the test when i try to push. Kindly advise

db/migrate/20211203172235_add_timezone_to_developers.rb
joemasilotti: It looks like you added a new migration – can you remove that and instead change the label in the form?

This will make sure your database is in a good state.

```bash
rails db:rollback
rm db/migrate/20211203172235_add_timezone_to_developers.rb
```

app/views/developers/_form.html.erb
```diff
@@ -40,9 +40,9 @@
             </div>
 
             <div class="sm:col-span-6">
-              <%= form.label :time_zone, class: "block text-sm font-medium text-gray-700" %>
+              <%= form.label :primary_time_zone, class: "block text-sm font-medium text-gray-700" %>
               <div class="mt-1">
-                <%= form.time_zone_select(:time_zone, ActiveSupport::TimeZone.us_zones, {include_blank: true}, {class: field_classes(form, :time_zone)}) %>
+                <%= form.primary_time_zone_select(:primary_time_zone, ActiveSupport::TimeZone.us_zones, {include_blank: true}, {class: field_classes(form, :primary_time_zone)}) %>
```
joemasilotti, line 45: I think this is the issue. `#primary_time_zone_select` isn't a method on the form builder.

app/models/concerns/avatarable.rb
joemasilotti:
```suggestion
    validates :avatar, presence: true, on: :create
```

Does this work?

joemasilotti:
I just left two comments in the code and one in a thread– let me know if those work!

app/models/concerns/avatarable.rb
soc-man: No but i'm working on a fix. Learning about custom validators. I realized just now that max_file_size and content_type were custom and are in the validators folder lol

joemasilotti: Nice! Looking forward to it. @ me in a comment when it's ready for review again :)

soc-man: @joemasilotti check out the new code. 

new failed test - the cause :
The following data requires normalization:
Run `i18n-tasks normalize` to fix
config/locales/en.yml

Should I proceed and normalize the data?

joemasilotti: Yes please! That will only touch the locale file.

soc-man: @joemasilotti done

soc-man: new errors below :

Error:
DeveloperTest#test_successful_profile_creation_sends_a_notification_to_the_admins:
ActiveRecord::RecordInvalid: Validation failed: Avatar is not attached, Search status is not included in the list, Time zone can't be blank
    test/models/developer_test.rb:105:in `block (2 levels) in <class:DeveloperTest>'
    test/models/developer_test.rb:104:in `block in <class:DeveloperTest>'

joemasilotti: The params you are sending in the test need to now also include the ones you made non-optional.

soc-man: right, time to learn again :) Let me get on that. Thanks again for the patience

soc-man: just to specify, having trouble attaching the avatar on the test

joemasilotti:
@soc-man, I pushed one commit which should get tie everything up.

It moves the developer avatar validation to be only on create, instead of chained on to the existing one. It also removes the validation for search status (for now). Finally, it fixes the tests!

If this looks good to you then we can merge it in. Thanks again for the contribution.

soc-man:
@joemasilotti sure! This was actually my first gig. I learned so much so thank you for that as well. If there's more stuff to be worked on, i'm down. It's been a great learning experience !:)

joemasilotti:
Welcome to OSS! 🎉 Thanks for your contribution.

There are a few more issues tagged as "good first issue" – feel free to pick anything up that isn't assigned.

PR #233, opened 2022-01-11
feat: Notifications dropdown

### LINKED ISSUE #220: View notifications as a lazy-loaded dropdown
I'd love for the notification bell to link to a lazy-loaded dropdown powered by Turbo. This aligns more closely with modern web apps (instead of linking to a new page).

- [ ] Add a desktop and mobile dropdown when clicking the bell icon
- [ ] Render the content via a lazy-loaded Turbo frame (meaning, only load it after someone clicks it) with a spinner
- [ ] Display the 5 most recent unread notifications
- [ ] Link to "All notifications" at the bottom
- [ ] Have an empty state if there are no unread notifications and link to "Read notifications"

jacobdaddario:
Resolves #220. There's a few things I'd like to talk about that I've opened code reviews for. Additionally, I wasn't sure if the system test configuration has been merged yet, so I can add in a system test if it has.

This PR adds a turboframe that loads in the notifications into a dropdown on the main UI. I made two changes from the original spec. 
1. The history link exists at the bottom of the menu instead of in the empty state.
2. All notifications are loaded. This is so that we could decompose the `notifications#index` page into a turboframe instead of requiring a completely new route.

### Pull request checklist


app/assets/images/icons/loading.svg
```diff
@@ -1,4 +1,4 @@
-<svg class="animate-spin -ml-1 mr-3 h-5 w-5" xmlns="http://www.w3.org/2000/svg" fill="none" viewBox="0 0 24 24">
```
jacobdaddario, line 1: This prevented the spin animation from being general use, but I'm not sure where and why it was there. Is it acceptable for this to be removed? I'm OOTL on where the component came from.

app/components/user_menu/signed_in_component.html.erb
```diff
@@ -1,11 +1,12 @@
 <div class="absolute inset-y-0 right-0 flex items-center pr-2 sm:static sm:inset-auto sm:ml-6 sm:pr-0">
-  <div class="text-white px-2" style="position:relative">
```
jacobdaddario, line 2: Is it okay to change this from an inline style to a Tailwind style? I couldn't tell if the original was intentional.

app/views/notifications/index.html.erb
```diff
@@ -1,19 +1,23 @@
...
-    <%= render PagyPaginatorComponent.new(id: "notifications", pagy: @pagy, url_array: [:notifications]) do |component| %>
-      <% component.loading_icon do %>
-        <%= render LoadingComponent.new %>
-      <% end %>
-      <%= render @notifications %>
+  <%= turbo_frame_tag "notifications" do %>
+    <% if @notifications.any? %>
+      <div class="my-6">
+        <h1 class="text-center text-3xl font-extrabold mb-5"><%= t(".title") %></h1>
+        <%= render PagyPaginatorComponent.new(id: "notifications", pagy: @pagy, url_array: [:notifications]) do |component| %>
```
jacobdaddario, line 6: The Pagy nav at the bottom of this component adds a bunch of spacing to the bottom of this set of notifications. It makes it difficult to re-use this section of the index as a turboframe. What's the reason for it, and why is it hidden? Can we remove it? Can we re-arrange how pagy is setup so that we don't have this issue with decomposing the view?

jacobdaddario (commented):
Here's a few questions that I want to clear up with you.

app/components/user_menu/signed_in_component.html.erb
joemasilotti: Huh, me neither. I don't see why not – go for it.

app/assets/images/icons/loading.svg
joemasilotti: I thought it was being used in the infinite scroll / Pagy component.

app/views/notifications/index.html.erb
joemasilotti: It's hidden because it should only show up when JavaScript is disabled. The Stimulus controller hides it on render. 

More than happy to decompose this if it makes sense. Same goes for the extra spacing - maybe that goes into the caller's code instead of the component.

joemasilotti:
Answered!

I'm a bit worried about loading all notifications. They kick of N+1 queries because each notification has to fetch the associated record(s) and can't be preloaded. Unless I'm missing something with Noticed!

joemasilotti:
Oh, and once #6 gets merged in we will have a system test suite!

jacobdaddario:
> Answered!
> 
> I'm a bit worried about loading all notifications. They kick of N+1 queries because each notification has to fetch the associated record(s) and can't be preloaded. Unless I'm missing something with Noticed!

You definitely know better. I have a very surface-level understanding of noticed and didn't realize it was an N+1. I assumed it was redirecting at the controller level of the show action of the notification.

Would you be opposed to adding a query param to the notifications index that acts as a `limit` value?

joemasilotti:
> Would you be opposed to adding a query param to the notifications index that acts as a `limit` value?

Taking a step back, do we want infinite scroll in a tiny little window? I'm wondering now if it might be better to always limit this to the 10 most recent unread notifications and link to `/notifications` if there are more unread available. Then the limit or pagination could be handled in the `respond_to` block.

jacobdaddario:
I don't think `respond_to` will catch turboframes, just turbostreams. If we added a limit method to the notifications chain, I think we wouldn't have to worry about pagy. It'd just load 5 notifications, and there wouldn't be anything to load on scroll.

joemasilotti:
Ah, good point. I've never used limit and pagy together, so that would require some experimentation to make sure everything works as we expect. Or are you saying limit only with a query param?

jacobdaddario:
It could just be a binary URL param passed in. If `params[:frame] = true` then we tack on `limit(5)` to the AR query. I can take a look after work today. It's been busy and I'm about to get back into things.

joemasilotti:
Sounds great to me. And as always, no rush :)

FYI, #221 has been merged in so system tests are a go!

jacobdaddario:
Just a lunchtime update from me.

Worked on this yesterday and had some issue getting the limit query to work the way I expected. I think maybe Pagy is overriding the limit value on the query. Need more time to investigate. Hoping this doesn't require a new controller.

joemasilotti:
I'm fairly confident that pagy does, indeed, override the limit parameter. You will need to use one or the other based on the param.

I'm not opposed to a new controller if we need it. It could be similar to how `ReadNotificationsController` works.

jacobdaddario:
Was thinking about it more after work. Really didn't want a separate controller because it didn't feel very Turbo-y, but I think that's the easiest solution. Sometimes the best tool in your box is the one that works. 🤷🏻‍♂️

joemasilotti:
@jacobdaddario, we might be able to use `#turbo_frame_request?` in the controller to handle the logic.

https://github.com/hotwired/turbo-rails/blob/main/app/controllers/turbo/frames/frame_request.rb

jacobdaddario:
Perfect timing. I was about to sit down and look at this. I'll have a peak.

jacobdaddario:
@joemasilotti I said screw it and went with the new controller approach. Everything is done now.

joemasilotti (commented):
I made a few little requests but I think we should rethink the architecture a bit. I think we are using Turbo Frames a little wrong here. `target: _top` is forcing us to render the _entire_ dropdown view. Instead, we could use a `<turbo-frame>` tag on `/notifications/index` - that would enable plopping it in a specific spot if we want to only render, for example, the notifications (and not the header and such).

Here's a diff that makes that change in a very broad sense. It would also simplify the code quite a bit. @DavidColby has a great [in-depth article](https://www.colby.so/posts/turbo-frames-on-rails) on how this all wires together.

One caveat is that the entire view is rendered, so `@pagy` would still need to be available. That means we could skip the whole `#turbo_frame_request?` thing and just paginate the entire time. Worse case? We could paginate (defaults to 10) and not render the pagination component, so it would only display 10 – and we could link to "more".

```diff
diff --git a/app/views/notifications/index.html.erb b/app/views/notifications/index.html.erb
index 9b2ee57..025aee7 100644
--- a/app/views/notifications/index.html.erb
+++ b/app/views/notifications/index.html.erb
@@ -1,4 +1,4 @@
-<div class="min-h-full flex flex-col justify-center py-12 sm:px-6 lg:px-8">
+<turbo-frame id="notifications" class="min-h-full flex flex-col justify-center py-12 sm:px-6 lg:px-8">
   <% if @notifications.any? %>
     <h1 class="mt-6 text-center text-3xl font-extrabold mb-5"><%= t(".title") %></h1>
     <%= render PagyPaginatorComponent.new(id: "notifications", pagy: @pagy, url_array: [:notifications]) do |component| %>
@@ -21,4 +21,4 @@
       <% end %>
     <% end %>
   </div>
-</div>
+</turbo-frame>
diff --git a/app/views/shared/_header.html.erb b/app/views/shared/_header.html.erb
index 5ede7ab..29d3a8f 100644
--- a/app/views/shared/_header.html.erb
+++ b/app/views/shared/_header.html.erb
@@ -34,7 +34,7 @@
               </button>
             </div>
             <div class="max-h-[40vh] px-2 pt-1 bg-gray-100 overflow-y-auto shadow-inner">
-              <%= turbo_frame_tag "notifications", loading: :lazy, src: notification_dropdown_path, target: "_top" do %>
+              <%= turbo_frame_tag "notifications", loading: :lazy, src: notifications_path do %>
                 <%= render LoadingComponent.new %>
               <% end %>
             </div>
```

I know that's a lot, let me know if anything doesn't make sense or I you'd like me to take a crack at it.

jacobdaddario:
@joemasilotti 

Okay, I want to break this down, because I'm actually a bit confused about a few of the asks there.

1. Regarding `target="_top"`: I'm confused by this ask here, and I think that's because there's a breakdown between our understandings of what that attribute does. To my knowledge, it creates a full page navigation, but not in the way that you're describing. As I understand it, it makes it so that it loads only the frame fragment from the target URL it's pulling from on first load, thanks to the `src` attribute being present. Then, after that loads, any navigation in the frame breaks things out to a full-page navigation. That seems right to me, and it's the behavior I'm seeing in my browser. Clicking a notification fulling navigates me to the resource for which it's representing instead of of trying to render a frame from the resource's destination. I think we don't have an agreement on that attribute does.
2. Is the second ask about decomposing `notifications#index` instead of a new controller just asking to do what we were originally doing? Pagy and all? That was what I originally did, so I could just reset the branch back to `fb27559a852d287db67e264e63d562ac58a80a29` and then cherry-pick the commit that added the system testing. I want to confirm this before I do it though, because that will re-write the git history and I won't be able to get my progress on the current solution back.

joemasilotti:
1. You are 100% right. Disregard my comment on `_top`, I was mixing two things up!
2. I think that is what I'm saying, yes. Apologies for the roundabout way of getting there. To be doubly safe, you could push this code to a new branch as a backup then come back and do the resetting and cherry-picking.

```
$ git branch -m notifications-dropdown-bk # rename the current branch
$ git push origin head -u # back up the backup to your fork
$ git checkout -b notifications-dropdown # recreate the branch as it was named before
$ git reset --hard ... # do your thang
```

app/views/shared/_header.html.erb
```diff
@@ -9,7 +9,7 @@
         </button>
       </div>
 
-      <div class="flex-1 flex items-center items-stretch justify-start ml-2 md:ml-0">
+      <div class="flex-1 flex items-center justify-start ml-2 md:ml-0">
```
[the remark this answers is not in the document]
jacobdaddario: This has to do with the way that the dropdown for notifications works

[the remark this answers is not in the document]
jacobdaddario: In order to get it to be responsive, the class has to switch. The navbar still seems to be responsive as far as I can tell.

jacobdaddario (commented):
Responding to your comments.

jacobdaddario:
@joemasilotti Really happy with how this came out. Originally, I didn't go with this, but I really wanted to keep the header being in the pre-rendered dropdown, as did you, and it made me realize I could have two turboframes in a view as long as they're in different branches of a conditional. Came out great.

joemasilotti (changes requested):
Outside of the requested code change I want to call out a minor issue.

If you have more than 2 pages worth of notifications you will see multiple spinners when opening the notification drawer. I think this is because the "bottom of page" is being shown right away because the view is small.

I'm not too concerned about this, but it does kick off both network requests "at the same time" which could cause issues down the road. At best it's a performance issue we don't need to worry about yet but at worst it's race condition that puts the notifications out of order.

I don't want to rip out pagination (again) but I'm not sure of a good workaround. Do you have any ideas?

Also, there are a few design changes I'd like to make. I didn't give you any guidance on this so this is on me. Would you prefer I commit directly to your PR or merge in then make my changes separately?

app/views/shared/_notifications_dropdown.html.erb
```diff
@@ -0,0 +1,20 @@
+<div data-toggle-target="element" class="hidden absolute z-10 top-0 inset-x-0 p-2 md:w-screen md:max-w-sm md:right-9 md:top-11 md:left-auto">
```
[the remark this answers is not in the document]
jacobdaddario: Will do

jacobdaddario:
@joemasilotti You can always commit to my PRs 😄.  It's your repository after all.

I'll make the change to the file really quickly, and then I assume this is out of my hands? Just want to make sure before I take my focus away from this project for a few weeks.

Additionally, could you get a screen capture of the behavior you're describing? I'm genuinely surprised to hear that you're running into this problem. I've got my monitor set to 1440p resolution, and I can't fit 10 notifications in the dropdown without scrolling. Additionally, I don't think the auto-pagination controller should mount before the notifications are rendered... I can't conceptualize what's happening, and hopefully if you can get a screen capture of it, it'll make it clearer.

app/views/shared/_notifications_dropdown.html.erb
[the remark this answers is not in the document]
jacobdaddario: Done

joemasilotti (approved):

joemasilotti:
Ha, fair!

I approved the PR. Consider your responsibility relinquished :)

I'm going to make some design changes this weekend (hopefully) and will merge soon. Thanks again for your PR!

joemasilotti:
Ah, so the issue occurs when you are on `/notifications` and then open the drawer. Because we are binding to `#notifications` for the Turbo Frame, the dropdown is getting picked up first. So the pagination on the main frame is getting "eaten" by the dropdown and never appearing.

Not sure how I'm going to handle this just yet, but wanted to post an update.

jacobdaddario:
So are you saying that being at the notifications URL prevents the pagination object from instantiating correctly? That seems weird given that it's two different request cycles... Maybe I'm missing something here, or maybe it's a bug in Turbo?

joemasilotti:
Not exactly. Pagination works. But there are now two elements with the id "notifications". So when one pagination triggers it only loads the content into the first element - the drop down.

jacobdaddario:
Ahhhh, that's tricky

jacobdaddario:
By the way, I still do have a backup of the unique controller approach if you want to go that route to circumvent the non-unique IDs problem.

joemasilotti:
Hey again @jacobdaddario, sorry for the delay on this one. I'm going to close this PR for now.

The more I use the site the more I realize I don't "miss" having the dropdown. The new page works well enough for now. And if we can avoid extra complexity I'd like to keep it that way.

Nothing against your (two!) implementations, I just don't want to be responsible for maintaining this when there are so many other things going on.

Issue #219, opened 2022-01-06
Translations?


joemasilotti:
Most of the site's copy has been moved to [`locales/en.yml`](https://github.com/joemasilotti/railsdevs.com/blob/main/config/locales/en.yml) thanks to @fetijashari. I'd love if we could start taking advantage of this and being translating to other languages besides English!

If you are interested in helping out, let me know by replying below. I'm not fluent in anything other than English.

---

Ideally, non-English locales should fail safely. Meaning, if we are missing a Spanish translation it should fallback to the English copy for now. This ensures we can continue to update the site without being held back by translating into other languages after the initial bulk of the work is done.

Also, if we can add a rake task that shows which translations are missing for which language that would be awesome. I'm pretty sure the `i18n-tasks` gem, which is already configured, can help us there.

salochara:
Hey! 👋 

I'll happily help with this. 
Spanish is my native tounge, so I can definitely work on this!

I'll work on this during the weekend 😄

joemasilotti:
Amazing – thank you so much! I'm sure there will be some places where translations are missing. Feel free to extract those or ping me and I can take care of it.

hungmi:
HI!
It's so cool that railsdevs.com is going to be an international service.
I can and I want to do translations to Chinese Traditional/Simplified.
And I think this can be extended to Japanese in the future.
I am not a professional designer, but [some design changes](https://medium.com/dropbox-design/design-for-internationalization-24c12ea6b38f) may be needed in the future.
What is your thought about this? 😊

salochara:
Hey! 👋 

Just a quick update from my side.. I didn't have much time during the weekend to work on this. 
I'll be working on this hopefully during the week (or most probably during the weekend). 
I've been quite busy at work 😅

joemasilotti:
@hungmi,

> It's so cool that railsdevs.com is going to be an international service.

I'm excited too! This will be my first app I've worked on with translations, so I'm going to be learning a lot here.

> I can and I want to do translations to Chinese Traditional/Simplified. And I think this can be extended to Japanese in the future.

❤️ Thank you!

> I am not a professional designer, but some design changes may be needed in the future. What is your thought about this? 😊

If we have to make changes that's fine by me. I vote we apply the translations first then see what looks weird before doing anything upfront.

joemasilotti:
@salochara, no worries! Whenever you get to it is fine :)

freebian:
Hi @joemasilotti, 

[...]

PR #240, opened 2022-01-15
Add Spanish translations

salochara:
Hey there! 👋 

- I've added the Spanish translations of the whole application! 
- On how to handle the switching locales logic, there are a few [options](https://guides.rubyonrails.org/i18n.html#managing-the-locale-across-requests) on how to do this. For now, I've added an around_action and have been passing the `:es` locale as a param. Let me know how you want to handle this! 
- On a rake task that checks which translations are missing... By executing `bundle exec i18n-tasks health` when running `bin/check`, this task will return if there are any missing translations in any locale! 
- As @hungmi posted in this [issue](https://github.com/joemasilotti/railsdevs.com/issues/219), there are indeed some design changes that need to be done. For example, in this view👇
CleanShot 2022-01-14 at 20 52 32@2x

Let me know what you think! 

### Pull request checklist

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -3,6 +3,7 @@ class ApplicationController < ActionController::Base
   include StoredLocation
 
   rescue_from Pundit::NotAuthorizedError, with: :user_not_authorized
+  around_action :switch_locale
 
   def after_sign_in_path_for(user)
     if (stored_location = stored_location_for(:user)).present?
@@ -15,6 +16,11 @@ class ApplicationController < ActionController::Base
   end
 
   private
+
+  def switch_locale(&action)
+    locale = params[:locale] || I18n.default_locale
+    I18n.with_locale(locale, &action)
+  end
 
   def user_not_authorized
     flash[:alert] = I18n.t("pundit.errors.unauthorized")
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -3,7 +3,6 @@ class ApplicationController < ActionController::Base
   include StoredLocation
 
   rescue_from Pundit::NotAuthorizedError, with: :user_not_authorized
-  around_action :switch_locale
 
   def after_sign_in_path_for(user)
     if (stored_location = stored_location_for(:user)).present?
@@ -16,11 +15,6 @@ class ApplicationController < ActionController::Base
   end
 
   private
-
-  def switch_locale(&action)
-    locale = params[:locale] || I18n.default_locale
-    I18n.with_locale(locale, &action)
-  end
 
   def user_not_authorized
     flash[:alert] = I18n.t("pundit.errors.unauthorized")
--- a/app/views/businesses/_form.html.erb
+++ b/app/views/businesses/_form.html.erb
@@ -67,7 +67,7 @@
 
         <div class="mt-5 md:mt-0 md:col-span-2">
           <div>
-            <%= form.label :developer_notifications, class: "text-base font-medium text-gray-900" %>
+            <%= form.label t(".developer_notifications.legend"), class: "text-base font-medium text-gray-900" %>
             <p class="text-sm leading-5 text-gray-500"><%= t(".developer_notifications.help_text") %></p>
             <div class="mt-4">
               <%= render UpgradeAccountComponent.new(current_user) do %>
--- a/bin/check
+++ b/bin/check
@@ -10,7 +10,7 @@ bundle exec standardrb --fix
 bundle exec erblint --lint-all
 
 # Check i18n health
-bundle exec i18n-tasks health
+bundle exec i18n-tasks health en
 
 # Run the test suite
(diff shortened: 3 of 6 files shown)


hungmi:
Hi~ @salochara 
I just started to do some Chinese translations based on your work. 👍
You may want to add something like
```ruby
def default_url_options
  { locale: I18n.locale }
end
```
to `application_controller`. It helps me when I clicking around.

BTW, I want to know what is the preferred way to set the current locale, too.
I think this has something to do with the url structure.
Ex, the site I work for is using urls like `https://example.com/es/...`, `https://example.com/zh-TW/...`.

joemasilotti:
This is amazing - thank you so much for the contribution! Before we merge it in I want to make sure a few things are addressed.

First, I want to ensure a missing Spanish (or other language) translation does _not_ break the build. Development on the site is fast, and I don't want to be held up because we don't have translations ready for a new feature. If missing, lets fall back to the English translation. Ideally, `bin/check` only checks English but we have other tasks (probably already in the i18n gem) that can check other languages on their own.

Second, Devise provides translations for other languages via [devise-i18n](https://github.com/tigrish/devise-i18n). Let's use that instead of the manual translations since they will work if/when we use more of Devise.

Finally, since this is my first localized application, I have some questions on how this works in practice. What is standard practice for someone changing locales? I see you can add the `?locale=es` param, but that resets on each request and there's no way to discover it. Also, what are the SEO implications of our decisions? I know we can use a subdomain, a different domain entirely, and a top-level path, but would love to hear your experience here. I'm a bit concerned of doing a (sub)domain because the developer profiles will never be translated.

joemasilotti:
After reading [the Rails i18n guide](https://guides.rubyonrails.org/i18n.html) (again) I think I'm leaning towards `railsdevs.com/es/developers` where `railsdevs.com/developers` defaults to English. Then, a dropdown or links in the footer can link to different locales. This approach preserves the locale when sharing a URL with someone.

There are a few caveats with this approach that I'm not yet sure how to handle.

1. For the sitemap, are translated pages no longer canonical? How do we reflect this when sending to Google and such?
2. How does one see the Spanish version of the homepage? `railsdevs.com/es` would try to route to the `/es` path!
3. The language should be specified in `<head>`.

salochara:
Hey! 👋 

This makes sense to me. I believe this is a good approach.
``
After reading the Rails i18n guide (again) I think I'm leaning towards railsdevs.com/es/developers where railsdevs.com/developers defaults to English. Then, a dropdown or links in the footer can link to different locales. This approach preserves the locale when sharing a URL with someone.
``

This is also the first time I work with a localized application, so no expierence from this side either 🙃 haha. 

As for the devise translations, I've taken taken the trasnlations we're currently using direclty from [devise-i18n](https://github.com/tigrish/devise-i18n/blob/master/rails/locales/es.yml)


Also, after reading a bit on this, I believe we can implement the translate-missing keys using Google Translate, DeepL or Yandex as described [here](https://github.com/glebm/i18n-tasks#google-translate-missing-keys). I feel this will come in handy for this project

Let me know what do you think about all of this and what steps to take to move forward! 

Best!

joemasilotti:
> As for the devise translations, I've taken taken the trasnlations we're currently using direclty from devise-i18n

Let's use that gem directly instead of copy-pasting the file in. That ensures us that if any updates to Devise occur we will get the changes here, too.

> Also, after reading a bit on this, I believe we can implement the translate-missing keys using Google Translate, DeepL or Yandex as described here. I feel this will come in handy for this project

For now, I'd prefer not to automate any translations and have them all done manually. Which is why we still need to fallback to English if a translation is missing. Can you try [this approach](https://stackoverflow.com/a/7886246/384110) and see if it works as I'm describing?

---

Once these are taken care of I will revisit #256 to hopefully merge them in together.

salochara:
> Let's use that gem directly instead of copy-pasting the file in. That ensures us that if any updates to Devise occur we will get the changes here, too.

Hmm. I don't see the way of doing that with the gem. As far as I undesrtand, with the [customize view](https://github.com/tigrish/devise-i18n#customizing-views) part, it will regenerate the devise views. I might need some help from you on this one! 

> For now, I'd prefer not to automate any translations and have them all done manually. Which is why we still need to fallback to English if a translation is missing. Can you try this approach and see if it works as I'm describing?

As for this, with this line in `application.rb`
```
# Use default language as fallback if translation is missing
config.i18n.fallbacks = true    
```
I've tested with a case where there is a missing translation in the Spanish version, ask for the Spanish version via /?locale=es in the params, and will indeed fallback to the English translation. In short, I believe we're covered!

As for the logic for changing params, I see there's some work done on that in this PR https://github.com/joemasilotti/railsdevs.com/pull/256
I have removed the logic I implemented for it and will leave these on my PR:
- Translation (whole app and devise)
- Check only for i18n **english** health

joemasilotti:
Merged into #264.

PR #241, opened 2022-01-15
Add income report page [Fixes #234]

### LINKED ISSUE #234: Move public revenue and expenses in-house
Currently, the public revenue and expenses [live on Notion](https://same-legend-a38.notion.site/railsdevs-revenue-and-expenses-c28706a1b3a846d4a058e4fef135aa67). I'd love to move these three tables and rendering into the codebase.

image

Ideally, we don't have to calculate anything manually. So the few SUM columns should be done in Ruby (or the database). Quickly sketching this out, we could have a few YAML files that get read in and spit out into nice looking HTML styled with Tailwind UI. The [Simple striped](https://tailwindui.com/components/application-ui/lists/tables#component-d60e8c748260b622747ec1568ba9c509) table looks perfect for this use case.

image

If you want to want to work on this but don't have access to Tailwind UI then feel free to open a PR with an un-styled table and I can finish it up.

---

This is a great first issue for someone to pick up because you don't need to know a lot about the existing codebase to get started.

vluna:
### Description
- Copy table from [Notion site](https://same-legend-a38.notion.site/railsdevs-income-report-c28706a1b3a846d4a058e4fef135aa67) to /income-report within application

### Pull request checklist

### Related Issue
#234 

### To-do Ideas
- Pull data from API and store it in database
- Render tables from database


vluna:
@joemasilotti Note that I normalized and ran the command from the failing test in local and it passes. Let me know if there is an issue on my side and what could it be. Thanks!

app/controllers/expenses_controller.rb
```diff
@@ -0,0 +1,14 @@
...
+  private
+
+  def renderer
+    Redcarpet::Markdown.new(Redcarpet::Render::HTML)
+  end
+
+  def file(name)
+    Rails.root.join("app/views/expenses/show.#{name}.md")
+  end
+end
```
joemasilotti, line 14: Is any of this code needed?

config/locales/en.yml
```diff
@@ -171,6 +171,128 @@ en:
       attached: is not attached
       content_type_invalid: image format not supported
       max_file_size_invalid: image over %{size}
+  expenses:
```
joemasilotti, line 174: I don't think this data should be in a locale file. If we translate to another language (see #240) then we would have to duplicate all of this to that locale file and keep them in sync.

Instead, this should live in `data/monies.yml` or similar and be parsed as YAML then spit out.

Also, let's calculate the totals in Ruby. I don't want to have to worry about updating three fields every time I add an expense or revenue line item.

config/locales/en.yml
```diff
@@ -171,6 +171,128 @@ en:
...
+    show:
+      description_html: "<p class='mb-8'>This page documents all monetary transactions associated with railsdevs.com to further instill value #2, <a href='/about' class='underline'>doing everything in public</a>.</p>"
```
joemasilotti, line 176: Let's move the wrapping `<p>` tag out of the locale file.

config/locales/en.yml
```diff
@@ -171,6 +171,128 @@ en:
...
+            description: Heroku bill (prorated)
+            total: 6.59
+          - date: 12-12-2021
+            description: Fathom analytics
+            total: 14
+          - date: 01-01-2022
+            description: Heroku bill
+            total: 12.09
+          - date: 13-01-2022
+            description: Fathom analytics
```
joemasilotti, line 199: We lost the link here, right?

config/routes.rb
```diff
@@ -5,6 +5,7 @@
 
   resource :home, only: :show
   resource :about, only: :show, controller: :about
+  resource :expenses, only: :show, path: "income-report"
```
joemasilotti, line 8: If the route is `/income-report` then the controller should be named the same or similar.

joemasilotti: I'm happy with both being "financials".

joemasilotti (changes requested):
Thanks for the PR! Let me know if you have any questions on the review or if something doesn't make sense.

config/locales/en.yml
vluna: I think for this one, it will be better if we have access to the API. We can pull the info with a worker, do all the necessary calculations and then storing it in the DB.  Other than that, we can still use the YAML or a CSV if we just import and export of the current data. If so, which directory do you want the `monies.yml` or maybe `financials.yml`?

joemasilotti: My goal is to remove the external dependency entirely and make the entire thing static, outside of calculating the totals in Ruby.

vluna: I moved the data for the table around. I am sure there is a better way to do this but for now this is what I came up with.

app/controllers/financials_controller.rb
```diff
@@ -0,0 +1,22 @@
...
+  def overview_data
+    @expenses["data"].group_by { |e| e["date"].to_date.strftime("%B, %Y") }.merge(@revenue["data"].group_by { |e| e["date"].to_date.strftime("%B, %Y") }).map { |e, i|
+      {month: e, expense_total: (i.sum { |y| (y["description"] ? y["total"] : 0) }), revenue_total: (i.sum { |y| (y["source"] ? y["total"] : 0) })}
+    }.map { |e| {month: e[:month], expense_total: e[:expense_total], revenue_total: e[:revenue_total], total: (e[:revenue_total] - e[:expense_total])} }
+  end
+
+  def get_table(table)
+    YAML.safe_load(File.read("#{Rails.root}/data/financials.yml"))["financials"][table]
+  end
+end
```
joemasilotti, line 22: Let's move all of this to a model so the controller doesn't have to do any work.

```
class IncomeReport
  def expenses
     # ...
  end

  def total_expenses
    # ...
  end
end
```

Then, in the controller, `@report = IncomeReport.new`. And the view can `@report.expenses`.

I'd also love to move all magic strings and symbols to that new model. I don't like that the view/controller knows about the structure of the YAML files. This should all be abstracted away in the `IncomeReport` model.

Finally, once all the logic is in a PORO we can add tests!

joemasilotti (changes requested):

vluna:
Updates have been applied.

joemasilotti:
Thanks for making the changes! Can you please add tests? Ideally, all logic should be tested in the model and a single integration test that renders `/financials` and ensures nothing breaks.

joemasilotti:
Thanks @vluna! I have some design cleanup I'd like to take care of. But other than that I think you are all set.

image

vluna:
@joemasilotti tried my best to match the screenshot, let me know if that works.

joemasilotti:
Hey, sorry if I wasn't clear. I'm already working on this so no need to try and match it yourself. Thanks!

joemasilotti:
Thanks again for this PR! I ended up taking a step back and realizing I wanted something way more automated, as in 100% hands off. So I did exactly what I told you not to do and integrated Stripe :)

No hard feelings, but I think my direction changed on this. And I used your PR as a guide to get mine going.

Closed via #270.

PR #242, opened 2022-01-16
Fix seeds with avatar and time_zone to fix validation failures

jespr:
The seeds took a nose dive after validations for time_zone and avatar were added. This fixes that. Thoughts on avatar files for the `seeds.rb` file to be placed in `test/files`

### Pull request checklist


jespr:
Let me add a few more things to the seed before we merge :)

db/seeds.rb
```diff
@@ -7,49 +7,73 @@ def create_user!(name)
   )
 end
 
+def create_developer!(attributes, avatar_name:)
```
joemasilotti, line 10: If `avatar_name:` is required we can pass it with `attributes` so we don't have to use two hashes when calling this method.

```
def create_developer!(attributes)
  avatar_name = attributes.delete(:avatar_name)
  Developer.create!(attributes) do |developer|
    developer.avatar.attach(io: File.open(Rails.root.join("test/files/#{avatar_name}")), filename: avatar_name)
  end
end
```

db/seeds.rb
```diff
@@ -7,49 +7,73 @@ def create_user!(name)
...
+  developer = Developer.new(attributes)
+  developer.avatar.attach(io: File.open(Rails.root.join("test/files/#{avatar_name}")), filename: avatar_name)
```
joemasilotti, line 12: We already have some images in `test/fixtures/files/` – I'd rather use those instead of adding new ones.

joemasilotti (changes requested):
Thanks for the PR! I didn't even realize the seeds were broken.

db/seeds.rb
jespr: So just `lovelace.jpg` for all the devs, and basecamp for the company? or?

jespr: Added the changes :)

joemasilotti (changes requested):
I'd like to still use the images you uploaded – can you move them to `test/support/fixtures/`? Sorry if that was confusing in my first review.

Also, I think we can extract another helper for attaching an image.

```ruby
def attach(filename, to:)
  to.attach(io: File.open(Rails.root.join("test/fixtures/files", filename), filename: filename)
end

attach("basecamp.png", to: business.avatar)
```

jespr:
Doh! I completely misunderstood! This makes more sense! Let me add that

jespr:
@joemasilotti I think that's it

joemasilotti:
I made one small commit to optimize the images (why not?) and remove duplicates.

Thanks again for the contribution! I'll merge this in as soon as CI passes.

PR #244, opened 2022-01-17
Issue#213: More robust database seeds

xmarlox:
**Issue#213: More robust database seeds**

- Add 30 developers
- Notifications 30+
- Plus notification for initial conversation of first business to first developer
- Add `TRUNCATE_ALL` env var as argument to `bin/setup` or `rails db:seed` 
- Add seeds to `bin/setup`

### Pull request checklist


db/seeds.rb
```diff
@@ -1,79 +1,125 @@
+if Rails.env.production?
+  raise "You're in Production! Seeding data is prohibited!"
+end
```
joemasilotti, line 3:
```suggestion
raise "You're in Production! Seeding data is prohibited!" if Rails.env.production?
```

db/seeds.rb
```diff
@@ -1,79 +1,125 @@
...
-    developer.avatar.attach(io: File.open(Rails.root.join("test/fixtures/files/#{avatar_name}")), filename: avatar_name)
+    attach(avatar_name, to: developer.avatar)
   end
 end
 
-def attach(filename, to:)
-  to.attach(io: File.open(Rails.root.join("test/fixtures/files", filename)), filename:)
-end
+puts "Starting to seed data..."
+ApplicationRecord.transaction requires_new: true do
```
joemasilotti, line 37: If this is in a transaction then the entire seed should probably in one.

db/seeds.rb
```diff
@@ -1,79 +1,125 @@
...
+  time_zones = ["Eastern Time (US & Canada)", "Pacific Time (US & Canada)"]
+  developer = nil
+  puts " Creating developers..."
+  10.times do |num|
```
joemasilotti, line 41: I _really_ want to avoid "random" data in the seeds. I know this isn't actually random, but there's little value in "Dennis9 Ritchie" IMHO. But I also understand the maintenance upkeep of manually creating individual records for 10+ developers. Let me think on this a bit more and get back to you.

db/seeds.rb
xmarlox: @joemasilotti 
technically, the entire functionality of the seed is inside the transaction since the code above `ApplicationRecord.transaction requires_new: true do` are methods called inside the transaction block.

joemasilotti: Ah, right! Thanks for pointing that out.

joemasilotti:
Hey @xmarlox, do you still have appetite to get this across the finish line? If not, I'm going to close the PR for now.

joemasilotti:
Thanks for the contribution, @xmarlox! I've gone back and forth a bit on this but I'm going to close this for now. I booted up the app from scratch (new machine) today and didn't feel like I was missing much.

If we do try this again in the future I'd like something that strikes a balance between what we have now and something randomized. But I don't know what that looks like just yet – and it's not fair to you to keep this open for so long.

Thanks again and no hard feelings!

Issue #245, opened 2022-01-20
Unconfirmed User cannot re-send confirmation instruction


freebian:
**Describe the bug**
Unconfirmed Users cannot re-send their confirmation instruction. 
System is not recognising the email?

**To Reproduce**
0. Preliminaries: Registered User - not confirmed in the database
1. Go to https://railsdevs.com/users/confirmation/new
2. Type in an email from registered user with confirmed_at empty
3. Click "Resend confirmation instruction"
4. See error "Invalid E-mail or password."

**Additional context**
This error message also appears if you try to resend confirmation instructions to an already confirmed user

joemasilotti:
Thanks for reporting this! I'm not sure why it could be happening as I don't think we have any custom Devise configuration. 

Either way, do you have the appetite to try and fix it?

freebian:
Will check into this later. Documenting my finds here in the meantime:

It seems that something is interfering with the default devise logic, because:
- if you put in an already confirmed email the error is still the same. Default devise error would be "email already confirmed"
- if you put in a total random email the error is still the same. Default devise would be "email not found"

Somehow it creates a devise failure: not found in database.

freebian:
See PR: #246

joemasilotti:
Closed via #246.

PR #246, opened 2022-01-20
Bugfix #245

freebian:
Change form URL on devise confirmation > new

session_path was present. confirmation_path should be present


joemasilotti:
Thanks for the bug report and quick PR!

Issue #247, opened 2022-01-20
Developers are adding full social media links


joemasilotti:
Some developers are adding their full URL for social links, like `https://github.com/joemasilotti` instead of `joemasilotti`. This messes up the link because it assumes only the path should be set.

I'm not sure how to handle this best just yet. But some ideas include making it more obvious in the form, some sort of validation, and backfilling/"fixing" existing records.

ghiculescu:
I think making it clearer in the UI is a good idea but I also thought of a simple way to make it work in more cases: https://github.com/joemasilotti/railsdevs.com/pull/248

freebian:
I usually do what is named in Tailwind as "Input with add-on" and write the complete URL/Domain in it.
https://tailwindui.com/components/application-ui/forms/input-groups

```
<div>
  <label for="company-website" class="block text-sm font-medium text-gray-700">
    Company Website
  </label>
  <div class="mt-1 flex rounded-md shadow-sm">
    <span class="inline-flex items-center px-3 rounded-l-md border border-r-0 border-gray-300 bg-gray-50 text-gray-500 sm:text-sm">
      https://github.com/
    </span>
    <input type="text" name="company-website" id="company-website" class="flex-1 min-w-0 block w-full px-3 py-2 rounded-none rounded-r-md focus:ring-indigo-500 focus:border-indigo-500 sm:text-sm border-gray-300" placeholder="www.example.com">
  </div>
</div>
```

joemasilotti:
@freebian, I think we are already doing that, no?

<img width="1080" alt="image" src="https://user-images.githubusercontent.com/2092156/150648971-c3cc6565-c436-40d2-b117-9315896c367a.png">

thomasvanholder:
As an easy fix, you could make adding a url or account more explicit:

<img width="568" alt="CleanShot 2022-01-25 at 13 28 02@2x" src="https://user-images.githubusercontent.com/36309895/150985947-93b868cb-9eb7-4c8f-9132-905be8744e64.png">

dhairyagabha:
@joemasilotti some platforms basically cleanup the link in the text field on paste/value change. What about some javascript that handles this validation and cleans up the values on paste/value change?

Since the root domain of these platforms will remain the same, should be a simple search and replace.

joemasilotti:
There is active work on this issue in #248. Please comment there if you have ideas on a better approach then what is currently proposed.

PR #248, opened 2022-01-20
Handle social links with or without the link prefix

ghiculescu:
This should make more links work even if they were copy pasted rather than the user entering just their handle.

Fixes https://github.com/joemasilotti/railsdevs.com/issues/247

### Pull request checklist


joemasilotti:
Thanks for the PR! I like this as a temporary approach, but it doesn't address the underlying issue of folks not entering data the way the system expects. If you have the appetite, I'd prefer a different take.

### Validations

If we validate the content we should be able to enforce someone only typing in their username for each field. From some quick googling, platforms have the following restrictions:

* GitHub - alphanumeric with a single hyphen
* Twitter - alphanumeric with underscores
* LinkedIn - alphanumeric

From that, I think we can be liberal and create a `SocialUsernameValidator` that only accepts letters, numbers, hyphens, and underscores and use it for all 3. Then, the validation error can say "Your [GitHub|Twitter|LinkedIn] username can only contain letters, numbers, hyphens, and underscores."

I don't care about validating the single hyphen or that Twitter doesn't support hyphens, etc. Using the same validator for all 3 is fine. Especially because I don't want to have to update this if/when their policies change.

### Existing developer profiles

The tricky part is existing developer profiles. We have two options there:

1. Only run this validation if the previous value was blank. For example, this will trigger when creating a new profile and when adding a social username to an existing profile for the first time.
2. Writing a migration that converts all existing usernames to valid ones. This might be as simple as dropping the prefixes you listed above, but I can't say for sure without diving in to the data.

---

Let me know what you think and if you'd like to run with this!

ghiculescu:
@joemasilotti thanks for the feedback. I think I respectfully disagree though. My view is that you should always be liberal in what input you accept and try to handle as many different inputs as possible if they can all lead to a reasonable output. (There's sometimes an exception if you want to report on the data and thus really need it to be uniform.)

So in this specific case, it seems like the root cause is some developers type in their handle "correctly" and some copy and paste the URL from their profile page. In both use cases, we can work out what outcome is expected by the user. Therefore I think we should support either method of input, rather than making the second user group jump through hoops to suit our preferred data format.

Basically, rather than saying this:

> it doesn't address the underlying issue of folks not entering data the way the system expects.

I think we should be asking, how can we make the system handle data entered the way the user expects it to work?

joemasilotti:
Thanks for pushing back! After sleeping on this, I agree that we should accept as much as possible. However, I don't think we should store ill-formatted data in the database and rely on the presentation layer to clean it up.

My vote is to accept both usernames and full links in the form field. Then, using a version of your display logic, strip out the "prefix" and only store the username in the database. If extracted, we can use this to normalize all of the existing data, too.

I'm leaning towards an Active Record callback to trigger the normalization.

joemasilotti:
Hey @ghiculescu, do you still have appetite to take this over the finish line? If not, no worries, I can pick up where you left off.

joemasilotti:
All good! Thanks for the heads up. I'll take care of it.

Issue #235, opened 2022-01-11
Add "role type" filter


joemasilotti:
A PR addressing this issue should add a new filter to `/developers` so folks can filter profiles by what type of role they are interested in. Most of the code can be modeled after the time zone filter with the exception of this being a join model on `RoleType`.

themudassarhassan:
Hey @joemasilotti 
I would like to work on this one if not already taken.

joemasilotti:
That would be great, thank you! All yours.

joemasilotti:
Closed via #251.

PR #251, opened 2022-01-21
add role type filter to developers page

themudassarhassan:
This adds role type filter to developers page.
Related [Issue](https://github.com/joemasilotti/railsdevs.com/issues/235)
### Pull request checklist

### DIFF
--- a/app/components/developer_query_component.rb
+++ b/app/components/developer_query_component.rb
@@ -18,6 +18,15 @@ class DeveloperQueryComponent < ApplicationComponent
       .map { |offset| [offset, "#{offset} #{t("developer_query_component.gmt")}"] }
   end
 
+  def role_types
+    %i[part_time_contract full_time_contract full_time_employment]
+      .map { |role| [role, RoleType.human_attribute_name(role)] }
+  end
+
+  def role_selected?(role_pair)
+    query.role_types.include?(role_pair.first)
+  end
+
   private
 
   def formatted_time_zone(time_zone)
--- a/app/components/developer_query_component.rb
+++ b/app/components/developer_query_component.rb
@@ -18,12 +18,12 @@ class DeveloperQueryComponent < ApplicationComponent
       .map { |offset| [offset, "#{offset} #{t("developer_query_component.gmt")}"] }
   end
 
-  def role_types
-    RoleType::TYPES.map { |role| [role, RoleType.human_attribute_name(role)] }
-  end
-
   def role_selected?(role_pair)
     query.role_types.include?(role_pair.first)
+  end
+
+  def role_types
+    RoleType::TYPES.map { |role| [role, RoleType.human_attribute_name(role)] }
   end
 
   private
--- a/app/components/developer_query_component.rb
+++ b/app/components/developer_query_component.rb
@@ -7,7 +7,7 @@ class DeveloperQueryComponent < ApplicationComponent
     @query = query
   end
 
-  def selected?(time_zone_pair)
+  def time_zone_selected?(time_zone_pair)
     query.time_zones.include?(time_zone_pair.first)
   end
 
@@ -19,8 +19,7 @@ class DeveloperQueryComponent < ApplicationComponent
   end
 
   def role_types
-    %i[part_time_contract full_time_contract full_time_employment]
-      .map { |role| [role, RoleType.human_attribute_name(role)] }
+    RoleType::TYPES.map { |role| [role, RoleType.human_attribute_name(role)] }
   end
 
   def role_selected?(role_pair)
--- a/app/models/developer.rb
+++ b/app/models/developer.rb
@@ -25,7 +25,11 @@ class Developer < ApplicationRecord
     max_file_size: 10.megabytes
 
   scope :filter_by_utc_offset, ->(utc_offset) { where(utc_offset:) }
-  scope :filter_by_role_types, ->(role_types) { joins(:role_type).where("part_time_contract=? or full_time_contract=? or full_time_employment=?", role_types.include?(:part_time_contract), role_types.include?(:full_time_contract), role_types.include?(:full_time_employment)) }
+  scope :filter_by_role_types, ->(role_types) do
+    RoleType::TYPES.filter_map { |type|
+      where(role_type: {type => true}) if role_types.include?(type)
+    }.reduce(:or).joins(:role_type)
+  end
 
   scope :available, -> { where(available_on: ..Time.current.to_date) }
   scope :newest_first, -> { order(created_at: :desc) }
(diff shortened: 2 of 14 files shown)


app/components/developer_query_component.rb
```diff
@@ -18,6 +18,15 @@ def time_zones
       .map { |offset| [offset, "#{offset} #{t("developer_query_component.gmt")}"] }
   end
 
+  def role_types
+    %i[part_time_contract full_time_contract full_time_employment]
```
joemasilotti, line 22: This is now the third or fourth time the codebase is iterating over these keys. I'm worried if we update these we won't catch everywhere we use them.

Can you move this to a constant on `RoleType` and refactor the others?

```ruby
class RoleType < ApplicationRecord
  TYPES = %i[part_time_contract full_time_contract full_time_employment].freeze

  # ...
end

puts RoleType::TYPES
```

app/components/developer_query_component.rb
```diff
@@ -18,6 +18,15 @@ def time_zones
...
+      .map { |role| [role, RoleType.human_attribute_name(role)] }
+  end
+
+  def role_selected?(role_pair)
```
joemasilotti, line 26: If we are calling this `#role_selected?` we should change the name of line 10 to `#time_zone_selected?`.

app/queries/developer_query.rb
```diff
@@ -27,15 +28,24 @@ def time_zones
...
 
   def initialize_pagy
     @_records = Developer.includes(:role_type).with_attached_avatar
     sort_records
     time_zone_filter_records
+    role_type_filter_records
     @pagy, @records = build_pagy(@_records)
   end
 
+  def role_type_filter_records
```
joemasilotti, line 45: Let's move this below `#time_zone_filter_records`.

app/models/developer.rb
```diff
@@ -25,6 +25,7 @@ class Developer < ApplicationRecord
     max_file_size: 10.megabytes
 
   scope :filter_by_utc_offset, ->(utc_offset) { where(utc_offset:) }
+  scope :filter_by_role_types, ->(role_types) { joins(:role_type).where("part_time_contract=? or full_time_contract=? or full_time_employment=?", role_types.include?(:part_time_contract), role_types.include?(:full_time_contract), role_types.include?(:full_time_employment)) }
```
joemasilotti, line 28: I don't think this is querying for what we want.

If I don't select `full_time_contract` then this query will filter out developers who _aren't_ looking for a full-time contract. Instead, it should be including developers who both are and aren't looking. Does that make sense?

Said another way, each of these three statements should be added conditionally if the value is selected.

joemasilotti (changes requested):
Thanks for the PR! My big feedback is around the query. I don't think it's doing what we want. Also, please add tests that cover the changes to the component, the model, and the integration as a whole.

I also made some minor code change requests. If anyone of this is confusing do let me know and I can help tackle it with you.

app/components/developer_query_component.rb
themudassarhassan: Sure that makes sense.

app/components/developer_query_component.rb
themudassarhassan: sure will do that.

app/models/developer.rb
themudassarhassan: I have a confusion regarding your comment. Here's how the current implementation will work:

- If I select `full_time_contract` then this query will output developers who are looking for a full-time contact only.
- If I select `full_time_contract` and `part_time_contract` then this query will output developers who are either looking for a full-time contract or part-time contract, and same with any other combination.
- and if I select all three options then this query will output developers who are looking for a full-time contract or a part-time contract or a full-time employment.

Can you please tell which developers to output in each of the above scenario? So we are on the same page. Thanks.

joemasilotti: * If I select full-time then I should only see developers that have full-time checked.
* If I select full-time and part-time I should only see developers that have one or the other checked.
* If I select all three then I should see developers with any of the three options checked.
* If I select none then the filter isn't applied and all developers are shown.

The last one is working, so we can ignore that. But if I output the SQL from the first one here is the query.

```ruby
[12] pry(main)> Developer.filter_by_role_types([:full_time_contract]).to_sql
=> "SELECT \"developers\".* FROM \"developers\" INNER JOIN \"role_types\" ON \"role_types\".\"developer_id\" = \"developers\".\"id\" WHERE (part_time_contract=FALSE or full_time_contract=TRUE or full_time_employment=FALSE)"
```

That means we are lumping in folks who set the other two to `false`. **This is including folks who have `false` set for all three values in all queries.***

Compare that to this query, which only sets one of the values:

```ruby
[28] pry(main)> Developer.joins(:role_type).where(role_type: {part_time_contract: true}).to_sql
=> "SELECT \"developers\".* FROM \"developers\" INNER JOIN \"role_types\" \"role_type\" ON \"role_type\".\"developer_id\" = \"developers\".\"id\" WHERE \"role_type\".\"part_time_contract\" = TRUE"
```

Building on that, if someone filtered by more than one value:

```ruby
[36] pry(main)> Developer.joins(:role_type).where(role_type: {part_time_contract: true}).or(Developer.joins(:role_type).where(role_type: {full_time_contract: true})).to_sql
=> "SELECT \"developers\".* FROM \"developers\" INNER JOIN \"role_types\" \"role_type\" ON \"role_type\".\"developer_id\" = \"developers\".\"id\" WHERE (\"role_type\".\"part_time_contract\" = TRUE OR \"role_type\".\"full_time_contract\" = TRUE)"
```

---

Does that make sense?

joemasilotti: I think that this should get us what we want.

```suggestion
  scope :filter_by_role_types, ->(role_types) do
    RoleType::TYPES.filter_map { |type|
      where(role_type: {type => true}) if role_types.include?(type)
    }.reduce(:or).joins(:role_type)
  end
```

```ruby
[64] pry(main)> Developer.filter_by_role_types([:part_time_contract, :full_time_contract]).to_sql
=> "SELECT \"developers\".* FROM \"developers\" INNER JOIN \"role_types\" \"role_type\" ON \"role_type\".\"developer_id\" = \"developers\".\"id\" WHERE (\"role_type\".\"part_time_contract\" = TRUE OR \"role_type\".\"full_time_contract\" = TRUE)"
```

themudassarhassan: Yeah that makes sense. Thanks. I'll update the PR.

themudassarhassan:
@joemasilotti I have pushed changes. Please review when you get some time.

joemasilotti (approved):
Looks great! Thanks for the contribution. I made a few small copy/design tweaks. I'll merge this in as soon as CI passes.

PR #256, opened 2022-01-22
Add i18n basics

hungmi:
Hey everyone!

Based on https://github.com/joemasilotti/railsdevs.com/pull/240, I have made some changes to make things work with different locales.
Check points from https://github.com/joemasilotti/railsdevs.com/pull/240#issuecomment-1013928980 are added in the checklist below.
There are still a lot to do. 😄

According to [Google](https://developers.google.com/search/docs/advanced/crawling/localized-versions), there are three methods we can tell Google which pages are localized variations of the same content.
I followed the `HTML` [option](https://developers.google.com/search/docs/advanced/crawling/localized-versions#html).

About tests, I have solved all errors ~except those component tests~.
I didn't have a chance to use view_component, I really need some help from you.

And about the UI/UX, those footer links I add really need your help. 😅

### Pull request checklist

- [x] We can visit urls like `railsdevs.com/developers`, `railsdevs.com/zh-TW/developers`
- [x] `railsdevs.com/developers` defaults to English
- [x] Links in the footer can link to different locales
- [x] `railsdevs.com/es` will show the Spanish version of the homepage.
- [x] The language should be specified in `<head>`.

Edit:
All tests pass.


joemasilotti:
Thanks for this! A few questions and notes:

1. What's the point of passing `id:` explicitly to some URL helpers? That parameter is unpermitted (check the logs).
2. We will need to also iterate over translations in the sitemap. There might be more configuration needed to write the language to the sitemap, too.
3. Is the language supposed to be a dropdown or links to each in the footer?
4. Are the changes in the new language there or was it copy-pasted as an example to switch languages?
5. Stripe should not be nested under the locale, not `/admin`.

hungmi:
> Thanks for this! A few questions and notes:
> 
> 1. What's the point of passing `id:` explicitly to some URL helpers? That parameter is unpermitted (check the logs).
I will check again.
Because the `new_developer_message_url(developer)` will result in a `missing required keys: [:developer_id]` error.
```
Error:
BusinessSubscriptionCheckoutTest#test_redirects_to_sending_a_developer_message_if_a_developer_is_given:
RuntimeError: Wrapped undumpable exception for: ActionController::UrlGenerationError: No route matches {:action=>"new", :controller=>"cold_messages", :locale=>#<Developer id: 630170759, name: "Developer One", email: nil, ..., availability_status: "now">}, missing required keys: [:developer_id]
```

> 2. We will need to also iterate over translations in the sitemap. There might be more configuration needed to write the language to the sitemap, too.
According to the Google search documents I provide, sitemap is one of three ways to provide localized versions of pages.
Isn't it?

> 3. Is the language supposed to be a dropdown or links to each in the footer?
I believe this depends on how many languages will need to be shown. A Dropdown may be better when more than 5 languages.

> 4. Are the changes in the new language there or was it copy-pasted as an example to switch languages?
Yes, it's an example for test purposes. I am happy to use other complete translation files instead of the Chinese version now.

> 5. Stripe should not be nested under the locale, not `/admin`.
Ok!

joemasilotti:
1. Your error snippet looks like it's trying to pass the developer to the `locale` parameter. This is a little worrying, because I thought that's what the param merging you added to `ApplicationController `should be doing.
2. Ah, nice! I hadn't read the Google link, thanks for calling it out. Yes, going with the HTML approach looks correct and we should not have to change the sitemap.
3. Fair enough. I think we will launch this with 3 languages, maybe 4. I can take care of the design on that.
4. Gotcha, sounds good! I'll take care of merging in these change with the Spanish translations in a new PR.

config/locales/en.yml
```diff
@@ -266,6 +266,7 @@ en:
       email: Email
       github: GitHub
       income_report: Income report
+      # Please use the language name of this locale as the translation.
```
joemasilotti, line 269: FYI this comment will be removed via `bundle exec i18n-tasks normalize`.

app/views/shared/_alternate_page_links.html.erb
```diff
@@ -0,0 +1,3 @@
+<% I18n.available_locales.each do |locale| %>
```
joemasilotti, line 1: Should this skip over the current locale?

app/controllers/application_controller.rb
```diff
@@ -16,6 +17,18 @@ def after_sign_in_path_for(user)
...
+    I18n.locale = if params[:locale].to_s.to_sym.in?(I18n.available_locales)
+      params[:locale]
+    else
+      I18n.default_locale
+    end
+  end
+
+  def default_url_options(options = {})
+    options.merge({locale: I18n.locale})
+  end
```
joemasilotti, line 30: This method is making me a little nervous. 

1. `developer_url(Developer.first)` is now adding the locale `en` to the URL. If the current locale is English then we shouldn't add a locale at all. For example, clicking from the homepage to a developer adds in `/en/*`.
2. This looks to be the cause of the broken tests and adding needing to set the `id:` param for some path generation.

config/environments/test.rb
```diff
@@ -51,10 +51,10 @@
   # config.action_view.annotate_rendered_view_with_filenames = true
 
   # Devise mailer.
-  config.action_mailer.default_url_options = {host: "localhost", port: 3000}
+  config.action_mailer.default_url_options = {host: "localhost", port: 3000, locale: I18n.locale}
 
   # Configure host for URL helpers.
-  Rails.application.routes.default_url_options = {host: "localhost", port: 3000}
+  Rails.application.routes.default_url_options = {host: "localhost", port: 3000, locale: I18n.locale}
```
joemasilotti, line 57: Why are these only added for test but not for any other environment? And what is the purpose of them?

When in a configuration file `I18n.locale` will always be English, so we should be able to omit this, right?

config/locales/zh-TW.yml
```diff
@@ -0,0 +1,298 @@
+---
```
joemasilotti, line 1: Let's set [the fallback langauge](https://stackoverflow.com/a/7886246/384110) and only include the translated lines in this file.

test/integration/alternate_page_links_test.rb
```diff
@@ -0,0 +1,13 @@
+require "test_helper"
+
+class AlternatePageLinksTest < ActionDispatch::IntegrationTest
+  test "All available locales are shown in <head> with correct hreflang" do
```
joemasilotti, line 4:
```suggestion
  test "all available locales are shown in <head> with correct hreflang" do
```

app/views/shared/_alternate_page_links.html.erb
hungmi: No, [the examele here](https://developers.google.com/search/docs/advanced/crawling/localized-versions#example) shows the current locale should be present with a `hreflang="x-default"`.

config/locales/en.yml
hungmi: Yes, do you have any advice about how to remind other people about the purpose of this `language` key?

config/locales/zh-TW.yml
hungmi: Ok, we have `config.i18n.fallbacks = true` in `application.rb`. I have removed duplications in the zh-TW file.

config/locales/en.yml
joemasilotti: We could change the key name to make it more obvious. `language_in_native_tongue` or something like that.

test/integration/alternate_page_links_test.rb
hungmi: Edited.

app/controllers/application_controller.rb
hungmi: In order to avoid set locale to all the `*_url`, `*_path` in whole application, I take this from [Rails guide](https://guides.rubyonrails.org/i18n.html#setting-the-locale-from-url-params).
I have made some changes to avoid adding the locale en to the URL.
Please review the newer version.👍

config/routes.rb
```diff
@@ -1,22 +1,24 @@
...
-  resources :businesses, except: :destroy
-  resources :read_notifications, only: :index, path: "/notifications/read"
-  resources :notifications, only: %i[index show]
-  resources :conversations, only: %i[index show] do
-    resources :messages, only: :create
-    resource :block, only: %i[new create]
-  end
-  resources :developers, except: :destroy do
-    resources :messages, only: %i[new create], controller: :cold_messages
+  scope "(:locale)", locale: /#{I18n.available_locales.join("|")}/ do
```
joemasilotti, line 4:
```suggestion
  scope "(:locale)", locale: /#{I18n.available_locales.join("|")}/, defaults: {locale: nil} do
```

This change should fix the weirdness with the missing optional parameters when using route helpers. I think we can also get rid of some of that `default_url` configuration, too, but I haven't tested that.

hungmi: This is pretty neat! 👍
I have redo related changes and force pushed.

config/environments/test.rb
hungmi: This is not needed after set `locale nil` in `routes.rb`.

app/controllers/application_controller.rb
hungmi: This also not needed after set locale nil in routes.rb.

config/locales/en.yml
hungmi: I am not sure about this, but I changed the key name to `language_name_of_locale`.
How do you think? 😀

joemasilotti:
This is looking great! I'm going to give it one more manual test run and will provide a final review.

hungmi:
No, I just found I am wrong. [This](https://github.com/joemasilotti/railsdevs.com/pull/256#discussion_r790225824) seems to override `locale` to nil for all url helpers. 😱

joemasilotti:
I think we need to bring back the default URL options along with this change.

config/routes.rb
hungmi: I have removed this setting because all url helpers in View will get `locale nil`.
Instead, the `resolve_locale` in `ApplicationController` will help us to set the locale to nil when it's the default locale.

config/environments/test.rb
hungmi: You are right! I also need to set this for development environment to make notifications & mailers work.
[Here](https://github.com/excid3/noticed/issues/49) is a reference.
And instead of `I18n.locale`, I use `{locale: nil}`. Because currently there is no mechanism to set locale param to nil when it's the default locale.

hungmi:
Hi! @joemasilotti 
I just pushed a new version to show my concepts.
In short, we need to set `default_url_options` for `url_for` manually, because it is ignored for unknown reasons.
```ruby
class BusinessSubscriptionCheckout
  include Rails.application.routes.url_helpers
  self.default_url_options = Rails.application.routes.default_url_options
  ...
end
```

## Details below:
I used `pry` to check the `default_url_options` settings in `development.rb` or `test.rb` is working or not.

Take this as an example:
https://github.com/joemasilotti/railsdevs.com/blob/22fa712703d34059ea698260dbc99c4a3f1627f1/app/models/business_subscription_checkout.rb#L30-L38
I add a breakpoint on Line 31, and run this test to get me into rails console: https://github.com/joemasilotti/railsdevs.com/blob/61105c0363596d885ccd6b2ffa18d10b359c88ec/test/models/business_subscription_checkout_test.rb#L18-L25
Since we have `include Rails.application.routes.url_helpers` on Line 2 of `business_subscription_checkout.rb`, we can call `self.default_url_options`:
```ruby
    31: def success_url
 => 32:   binding.pry
    33:   if developer.present?
    34:     new_developer_message_url(developer)
    35:   elsif user.business.present?
    36:     conversations_url
    37:   else
    38:     new_business_url
    39:   end
    40: end

[1] pry(#<BusinessSubscriptionCheckout>)> self.default_url_options
=> {}
[2] pry(#<BusinessSubscriptionCheckout>)> Rails.application.routes.default_url_options
=> {:host=>"localhost", :port=>3000, :locale=>nil}
[3] pry(#<BusinessSubscriptionCheckout>)>
```

As you can see, the setting in environment files is ignored.
In fact, I have set all options I can find. All of them are ignored:
```ruby
# config/environments/development.rb
Rails.application.default_url_options = {host: "localhost", port: 3000, locale: nil}
Rails.application.config.action_mailer.default_url_options = {host: "localhost", port: 3000, locale: nil}
Rails.application.config.action_controller.default_url_options = {host: "localhost", port: 3000, locale: nil}
Rails.application.routes.default_url_options = {host: "localhost", port: 3000, locale: nil}
Railsdevs::Application.default_url_options = {host: "localhost", port: 3000, locale: nil}
```
---

And for `controllers`, what we need is the `default_url_options` in `ApplicationController`.

---
If you agree with these changes, I believe we can move on to the refactoring part. 👍

joemasilotti:
Excellent find! I'm not sure why it wouldn't get picked up, but your refactor looks good to me. Please continue with the refactor and let me know when it's ready for review :)

Also, in case it's not on your radar, it looks like `#resolve_locale` might be duplicated. Thanks for the time put into this one!

joemasilotti:
Can you add `en` to the the i18n-tasks step in CI? That's why it is failing right now.

app/lib/url_helpers_with_default_url_options.rb
```diff
@@ -0,0 +1,6 @@
+module UrlHelpersWithDefaultUrlOptions
+  include Rails.application.routes.url_helpers
+  def default_url_options
```
joemasilotti, line 3:
```suggestion

  def default_url_options
```

app/controllers/application_controller.rb
```diff
@@ -3,6 +3,8 @@ class ApplicationController < ActionController::Base
   include StoredLocation
 
   rescue_from Pundit::NotAuthorizedError, with: :user_not_authorized
+  before_action :set_locale
```
joemasilotti, line 6:
```suggestion

  before_action :set_locale
```

app/notifications/new_message_notification.rb
```diff
@@ -1,4 +1,5 @@
 class NewMessageNotification < Noticed::Base
+  include UrlHelpersWithDefaultUrlOptions
```
joemasilotti, line 2:
```suggestion
  include UrlHelpersWithDefaultUrlOptions
  
```

config/environments/development.rb
```diff
@@ -65,18 +65,15 @@
   # config.action_cable.disable_request_forgery_protection = true
 
   # Devise mailer.
-  config.action_mailer.default_url_options = {host: "localhost", port: 3000}
+  config.action_mailer.default_url_options = {host: "localhost", port: 3000, locale: nil}
 
   # Configure host for URL helpers.
-  Rails.application.routes.default_url_options = {host: "localhost", port: 3000}
+  Rails.application.routes.default_url_options = {host: "localhost", port: 3000, locale: nil}
```
joemasilotti, line 71: Do we need these for production, too?

test/integration/switch_locale_test.rb
```diff
@@ -0,0 +1,36 @@
+require "test_helper"
+
+class SwitchLocaleTest < ActionDispatch::IntegrationTest
+  test "I18n.locale will depends on the locale of urls" do
+    [:en, :"zh-TW"].each do |locale|
```
joemasilotti, line 5:
```suggestion
    I18n.available_locales.each do |locale|
```

app/controllers/application_controller.rb
```diff
@@ -16,6 +18,22 @@ def after_sign_in_path_for(user)
...
+    end
+  end
+
+  def default_url_options(options = {})
+    options.merge({locale: resolve_locale})
+  end
+
+  def resolve_locale(locale = I18n.locale)
+    locale == I18n.default_locale ? nil : locale
+  end
```
joemasilotti, line 35: If this is a helper method, should it be private? Genuine question, not implying anything.

joemasilotti (changes requested):
Almost there! Just a few small code formatting tweaks and a question. Other than that we are ready to merge!

app/controllers/application_controller.rb
hungmi: Oh, I never thought about this.
But I believe it's good to left only actions as public in a controller?
[This](http://lifeonubuntu.com/why-your-controller-methods-should-be-private/) explains better than me.

test/integration/switch_locale_test.rb
hungmi: I think a combination of the default locale and another one is enough to bring our attentions when things break.
How do you think?

config/environments/development.rb
hungmi: Yes, we need.
The current setting in `production.rb` is `Rails.application.routes.default_url_options[:host] = ENV["HOST"]`.
Can it be changed to the hash format like this?

test/integration/switch_locale_test.rb
joemasilotti: Good point.

config/environments/development.rb
joemasilotti: Yeah, go for it. We won't need to set `port`, so just `host` and `locale`.

hungmi: Done!

joemasilotti:
Merged into #264.

PR #257, opened 2022-01-23
Make links in messages clickable [1/2]

djfpaagman:
The other day I got a message with a link in it, which I had to copy paste into my browser. I think it makes sense to make links clickable, also called "autolinking". For this I've added the [rinku](https://github.com/vmg/rinku) gem which looks a bit stale at first glance but I think it still does the job alright. Better than messing with regexes at least.

I've decided to just pass the string through `rinku` and then through `simple_format` with the rest of the arguments.

See screenshots below for what it looks like. Note that links are not really standing out right now, is that something to improve as well? I noticed that none of the links have underlines or anything so feedback there is welcome.

image
image

### Pull request checklist

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -52,4 +52,4 @@ gem "sidekiq", "~> 6.3"
 gem "sitemap_generator", "~> 6.1"
 gem "stripe", ">= 2.8", "< 6.0"
 gem "tzinfo-data", platforms: [:mingw, :mswin, :x64_mingw, :jruby]
-gem "rinku"
+gem "auto_html"
--- a/Gemfile
+++ b/Gemfile
@@ -36,6 +36,7 @@ group :test do
   gem "webdrivers"
 end
 
+gem "auto_html", "~> 2.1"
 gem "aws-sdk-s3", "~> 1", require: false
 gem "classy-yaml", "~> 0.7"
 gem "devise", "~> 4.8.1"
@@ -52,4 +53,3 @@ gem "sidekiq", "~> 6.3"
 gem "sitemap_generator", "~> 6.1"
 gem "stripe", ">= 2.8", "< 6.0"
 gem "tzinfo-data", platforms: [:mingw, :mswin, :x64_mingw, :jruby]
-gem "auto_html"
--- a/Gemfile
+++ b/Gemfile
@@ -52,3 +52,4 @@ gem "sidekiq", "~> 6.3"
 gem "sitemap_generator", "~> 6.1"
 gem "stripe", ">= 2.8", "< 6.0"
 gem "tzinfo-data", platforms: [:mingw, :mswin, :x64_mingw, :jruby]
+gem "rinku"
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -82,6 +82,10 @@ GEM
     addressable (2.8.0)
       public_suffix (>= 2.0.2, < 5.0)
     ast (2.4.2)
+    auto_html (2.1.0)
+      gemoji (~> 4.0.0.rc2)
+      redcarpet (~> 3.5)
+      rinku (~> 2.0)
     aws-eventstream (1.2.0)
     aws-partitions (1.519.0)
     aws-sdk-core (3.121.3)
@@ -142,6 +146,7 @@ GEM
     ffi-compiler (1.0.1)
       ffi (>= 1.0.0)
       rake
+    gemoji (4.0.0.rc3)
     globalid (1.0.0)
       activesupport (>= 5.0)
     highline (2.0.3)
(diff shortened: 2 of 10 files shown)


joemasilotti:
I love the idea of linking to URLs and email addresses. But I'm not 100% sure we want to add a dependency that hasn't been updated in a few years. Also, if it's written in C is there a native binary that needs to be installed?

I'm wondering if we want to explore https://github.com/dejan/auto_html instead. It was last updated 2 months ago and seems to do what we want. 

Dependencies aside, in the long term messages should probably be formatted by the writer via ActionText.

djfpaagman:
@joemasilotti I've replaced `rinku` with `auto_html` in the latest commit :+1:, it does look like a better gem indeed, I didn't know of it's existence yet, so it's nice to discover it this way as well :smile:

Gemfile
```diff
@@ -52,3 +52,4 @@ gem "sidekiq", "~> 6.3"
 gem "sitemap_generator", "~> 6.1"
 gem "stripe", ">= 2.8", "< 6.0"
 gem "tzinfo-data", platforms: [:mingw, :mswin, :x64_mingw, :jruby]
+gem "auto_html"
```
joemasilotti, line 55: Please lock this to a minor version via "~>" and sort this "group" of gems.

app/helpers/message_helper.rb
```diff
@@ -0,0 +1,8 @@
+module MessageHelper
+  def smart_format(string, **options)
+    link_filter = AutoHtml::Link.new
+    with_links = link_filter.call(string)
+
+    simple_format(with_links, options)
```
joemasilotti, line 6: auto_html has it's own `AutoHtml::SimpleFormat` pipeline. Let's use that instead via a `AutoHtml::Pipeline`.

joemasilotti (changes requested):
Looking good! I made two small requests.

Outside of those, the README recommends [storing the formatted text in the database](https://github.com/dejan/auto_html#using-autohtml-with-activerecord). I think we should handle that now since messages are not yet paginated and some conversations have 10+ messages. This also gets us around needing a helper because we can call `message.body_html` directly.

That said, it will require a small data migration to update existing records. A rake task that I can run in production will be perfect.

djfpaagman:
@joemasilotti implemented your suggestions, I've added a migration for the new field and a rake task to fill it. To deploy this now requires the following steps:

1. **Merge the PR** :wink: 
1. **Run the migration**
2. **Run the Rake task** `rails messages:migrate_body_html`

Note that it does not use the new field yet, because it will be empty until the migration has ran. To have a nice deployment I've decided to split it out. So that means the following needs to happen in a new PR/deploy:

1. **Start using the new field.** A small thing still to tackle is the use of the `mt-4` class for paragraphs now (see code snippet below). SimpleFormat does not support that. A possible solution could be to wrap the message in a `prose` class from tailwind typography, which handles margins for paragraphs automatically. What do you think?
https://github.com/joemasilotti/railsdevs.com/blob/61105c0363596d885ccd6b2ffa18d10b359c88ec/app/views/messages/_message.html.erb#L14
2. **Remove the `skip` in `test/integration/messages_test.rb:101`.** Since the formatting is now also tested in the model this could also be wrapped up in the "messages are formatted" test I guess.
2. (optional) Add `null: false` to the `body_html` column
1. (optional) Remove the Rake task

Gemfile.lock
```diff
@@ -305,6 +310,7 @@ GEM
       mime-types (>= 1.16, < 4.0)
       netrc (~> 0.8)
     rexml (3.2.5)
+    rinku (2.0.6)
```
joemasilotti, line 313: Does auto_html depend on rinku or did this sneak in from your first pass?

joemasilotti (approved):
Outside of one minor comment this looks great!

Can you also submit the follow up PR that takes care of removing the rake task, making sure the column is non-null, and uses the new column? Once that is approved I will kick off the flow.

Thanks for the contribution!

Gemfile.lock
djfpaagman: heh yeah interestingly it's also a dependency from `auto_html` :smile:

djfpaagman:
Part 2 is up! #262 .

I left a comment there about the migration, I think it would be nice to tackle that before merging either of the PRs.

joemasilotti:
OK, all merged in! Thanks so much for the contribution.

`auto_html` did take down the site for a few minutes, but I [opened a PR](https://github.com/dejan/auto_html/pull/187) so it doesn't happen to anyone else in the future.

image

djfpaagman:
Oof weird that that didn't show up in development 🤔  great that it's merged and live though 🎉

joemasilotti:
I think because Standard depends on it so we already had the gem installed.

PR #262, opened 2022-01-27
Make links in messages clickable [2/2]

djfpaagman:
Follow up of https://github.com/joemasilotti/railsdevs.com/pull/257.

**:warning: Note: only merge after #257 has been merged and the data migration task has been run!**


db/migrate/20220127141658_set_body_html_null_false.rb
```diff
@@ -0,0 +1,7 @@
+class SetBodyHtmlNullFalse < ActiveRecord::Migration[7.0]
+  def change
+    execute "UPDATE messages SET body_html = '' WHERE body_html IS NULL"
```
djfpaagman, line 3: I've added this here to make sure the database is migratable for local development environments, else it will complain about `body_html` having null values and will stop the migration. The downside of this is that people who only run this migration won't have any message bodies, because they're just set to an empty string here.

How do you feel about doing the data migration that's in the Rake task in the database migration? I'm not against it for simple stuff. The big upside is that it makes life a bit easier for other developers working on the app. The downside is that it can get stale and break eventually. WDYT?

djfpaagman: The alternative is that people just reload the seeds from this point on, could be viable as well. Not sure how to properly communicate that though :thinking:

joemasilotti: Let's remove it. If it is only for the development environment we can communicate that to folks.

Also, there's a single message in the seeds right now. So I don't think folks are relying to seeing a lot of them appear.

joemasilotti (changes requested):

joemasilotti (approved):

PR #264, opened 2022-01-29
Translations

### LINKED ISSUE #219: Translations?

joemasilotti:
This PR combines the work done in #256 with the Spanish translations in #240. I had to rework some of the Devise translations and made a few design changes. Everything seems to be working fine, including CI.

The problem is that tests seem to be failing sporadicly. I can't nail down what is going on but I'm assuming it has to do with setting the locale. If anyone has any ideas of what's going on (CC @hungmi and @salochara) I'd love some help!

Closes #219, #256, and #240.

### Pull request checklist

### DIFF
--- a/app/views/shared/_alternate_page_links.html.erb
+++ b/app/views/shared/_alternate_page_links.html.erb
@@ -1,3 +1,0 @@
-<% I18n.available_locales.each do |locale| %>
-  <link rel="alternate" href="<%= request.base_url + url_for(locale: resolve_locale(locale)) %>" hreflang="<%= locale == I18n.locale ? "x-default" : locale %>">
-<% end %>
--- a/app/views/shared/_alternate_page_links.html.erb
+++ b/app/views/shared/_alternate_page_links.html.erb
@@ -1,0 +1,3 @@
+<% I18n.available_locales.each do |locale| %>
+  <link rel="alternate" href="<%= request.base_url + url_for(locale: resolve_locale(locale)) %>" hreflang="<%= locale == I18n.locale ? "x-default" : locale %>">
+<% end %>
--- a/test/integration/switch_locale_test.rb
+++ b/test/integration/switch_locale_test.rb
@@ -14,8 +14,8 @@ class SwitchLocaleTest < ActionDispatch::IntegrationTest
     end
   end
 
-  test "links to all locales will show, except the current locale" do
-    current_locale = :"zh-TW"
+  test "links to all locales will show" do
+    current_locale = :es
     get root_path(locale: current_locale)
     assert I18n.locale == current_locale
 
@@ -27,7 +27,6 @@ class SwitchLocaleTest < ActionDispatch::IntegrationTest
         assert_select "a[href='/#{locale}']", text: language_name_of(locale)
       end
     end
-    assert_select "a[href='/#{current_locale}']", count: 0, text: language_name_of(current_locale)
   end
 
   def language_name_of(locale)
--- a/test/integration/switch_locale_test.rb
+++ b/test/integration/switch_locale_test.rb
@@ -20,7 +20,7 @@ class SwitchLocaleTest < ActionDispatch::IntegrationTest
     [:en, :"zh-TW"].each do |locale|
       I18n.with_locale(locale) do
         get root_path
-        
+
         assert_select "a[href='/']", text: language_name_of(I18n.default_locale)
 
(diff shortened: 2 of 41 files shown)


joemasilotti:
Ah, perfect! CI failed and it looks like it's because the locale is still set to Spanish for some reason. I think the `before_action` is messing things up – maybe it needs to be reset under test?

test/integration/switch_locale_test.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  test "Visit urls with an unavailable locale will return 404" do
+    assert_raises ActionController::RoutingError do
+      get root_path(locale: "Klingon")
+    end
+  end
+
+  test "links to all locales will show" do
+    current_locale = :es
```
gmmcal, line 18: You may have issues with this in the future. it is better to use `I18n.with_locale` when testing locale switches

From https://guides.rubyonrails.org/i18n.html

```
2.2 Managing the Locale across Requests

A localized application will likely need to provide support for multiple locales. To accomplish this, the locale should be set at the beginning of each request so that all strings are translated using the desired locale during the lifetime of that request.

The default locale is used for all translations unless I18n.locale= or I18n.with_locale is used.

I18n.locale can leak into subsequent requests served by the same thread/process if it is not consistently set in every controller. For example executing I18n.locale = :es in one POST requests will have effects for all later requests to controllers that don't set the locale, but only in that particular thread/process. For that reason, instead of I18n.locale = you can use I18n.with_locale which does not have this leak issue.
```

That can explain why an unrelated test is failing on CI

hungmi: @gmmcal @joemasilotti 
I have pushed [new commits](https://github.com/joemasilotti/railsdevs.com/pull/256#commits-pushed-5087736) to use `I18n.with_locale` instead.
Can you take a look?

gmmcal: @hungmi these changes on #256 lgtm

app/views/shared/_alternate_page_links.html.erb
```diff
@@ -0,0 +1,3 @@
+<% I18n.available_locales.each do |locale| %>
+  <link rel="alternate" href="<%= request.base_url + url_for(locale: resolve_locale(locale)) %>" hreflang="<%= locale == I18n.locale ? "x-default" : locale %>">
```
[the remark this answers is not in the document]
joemasilotti: Good call. Done!

test/integration/switch_locale_test.rb
```diff
@@ -0,0 +1,35 @@
+require "test_helper"
+
+class SwitchLocaleTest < ActionDispatch::IntegrationTest
+  test "I18n.locale will depends on the locale of urls" do
+    [:en, :"zh-TW"].each do |locale|
```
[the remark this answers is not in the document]
joemasilotti: Ended up going with `I18n.available_locales`.

test/integration/switch_locale_test.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  test "links to all locales will show" do
+    current_locale = :es
+    get root_path(locale: current_locale)
+    assert I18n.locale == current_locale
+
+    other_locales = I18n.available_locales - [current_locale]
+    other_locales.each do |locale|
+      if locale == I18n.default_locale
```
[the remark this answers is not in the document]
joemasilotti: I agree. And @hungmi already ripped it out!

test/integration/switch_locale_test.rb
joemasilotti: Thanks for the updates! I cherry-picked your commits onto this branch.

joemasilotti:
Thanks for the review and help @gmmcal and @hungmi!

PR #265, opened 2022-01-29
Profile images do not keep attached on form error

### LINKED ISSUE #249: Profile images do not keep attached on form error
**Describe the bug**
When creating (and probably updating) profile, if some text field is invalid and user uploads an image, preview of image is shown on error page, but when user tried to save again, a new error pops saying image was not uploaded.

**To Reproduce**
1. Go to 'https://railsdevs.com/developers/:id/edit' on newly created profile
2. Remove data from one of mandatory fields
3. Upload a new image
4. Click on `Save`
5. Error of mandatory field should show on page, as well as image preview
6. Add content to missing field
7. Try to save again
8. Error of image not uploaded should show up

**Expected behavior**
Profile should be saved correctly after mandatory field is filled

gmmcal:
Fixes #249

When an image is uploaded on developer profile and some other field fails validation, the original image is not persisted across requests and fails do be properly set on second request.

This causes user to have to re-upload image and creates a set of phantom images on database (and potentially on storage as well).

### Pull request checklist

### DIFF
--- a/app/views/businesses/_form.html.erb
+++ b/app/views/businesses/_form.html.erb
@@ -46,8 +46,8 @@
                     <%= inline_svg_tag "icons/outline/refresh.svg", class: "w-8 h-8 text-gray-900 animate-reverse-spin" %>
                   </div>
                 </div>
-                <%= form.file_field :avatar, direct_upload: true, accept: "image/png, image/jpg, image/jpeg", data: {action: "file-upload#select direct-upload:initialize->file-upload#start direct-upload:error->file-upload#error"}, class: "hidden" %>
                 <%= form.hidden_field :avatar, value: form.object.avatar.signed_id, id: "business_avatar_hidden" if form.object.avatar.attached? %>
+                <%= form.file_field :avatar, direct_upload: true, accept: "image/png, image/jpg, image/jpeg", data: {action: "file-upload#select direct-upload:initialize->file-upload#start direct-upload:error->file-upload#error"}, class: "hidden" %>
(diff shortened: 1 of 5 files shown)


gmmcal:
@joemasilotti build is broken due to ruby download. can you restart action?

joemasilotti:
Nice! Can you apply this to the business form, too?

gmmcal:
Any other form as well?

joemasilotti:
Nope, those are the only two that use an avatar right now. Also note, the header image might have the same issue.

joemasilotti:
Thanks for the quick turnaround! Two quick requests before I merge:

1. Can you add a test to cover the new code? Even if it's looking for the hidden field via a CSS selector. Just so we don't drop it in the future. 
2. Is this the recommended approach for active storage uploads? I've never see this (or used signed_id before) and want to make sure we aren't introducing a hack. 

Thanks again.

gmmcal:
1. check
2. hard to find official documentation about this, but here is a good explanation https://stackoverflow.com/questions/50360307/active-storage-best-practice-to-retain-cache-uploaded-file-when-form-redisplays. I feel like (not confirmed) that using simple-form instead of manually writing html form would have solved this from start, but changing it now might take a while

joemasilotti:
I think there might be bug in this. To reproduce:

1. Create a developer profile
2. Edit the profile
3. Add a new avatar
4. Click Save

I expect to see the new avatar. Instead, I see the old one.

gmmcal:
Odd... will check. Thanks for testing

gmmcal:
This happens to be much harder than anticipated. Haven't found a good solution yet

joemasilotti:
Got it, thanks for the update! Let me know if you make any progress or want a second pair of eyes.

joemasilotti:
Hey again @gmmcal, do you still have appetite to fix the bug that this PR introduced? If not let me know and I can take a pass or close the PR.

gmmcal:
I do. I'm only lacking time =\

joemasilotti:
Moving the hidden field above the file field fixes the issue! I think this is because the upload will override the hidden one's parameters when it changes.

Thanks again for the PR. I'm going to merge this in when CI passes.

PR #273, opened 2022-02-03
Add relative availability

boardfish:
Developers may be immediately available after a given notice period (a fixed period of time). This PR aims to store that in days and use it instead of `available_on` if it's specified, meaning that developers won't have to revisit the platform daily to update the date on which they're available. Their availability is instead marked by a window of days.

Two fields, marked "Available to start on" (with a Clear button beside it) and "Available in (days)". "Available to start on" is a date field, and "Available in (days)" is a number field.

Each field clears the other one if it is edited, so that both fields can't be set at the same time. If they are somehow both set, `available_on` takes priority. The `Clear` button allows a developer to set their availability to the `unspecified` state.

TODO:

- [x] Add to form
- [x] Make it easy to submit either a fixed date or a relative time period

Seems like right now, the default availability date on the form is the current date. It's probably better to make that fixed date `nil` if relative availability is set.

### Pull request checklist

### DIFF
--- a/app/views/developers/_form.html.erb
+++ b/app/views/developers/_form.html.erb
@@ -161,8 +161,24 @@
                 <div class="grid grid-cols-1 sm:grid-cols-3 sm:col-span-3 md:gap-6">
                   <div class="sm:col-span-1">
                     <%= form.label :available_on, class: "block text-sm font-medium text-gray-700" %>
-                    <div class="mt-1">
-                      <%= form.date_field :available_on, value: developer.available_on_in_database, class: "shadow-sm focus:ring-gray-500 focus:border-gray-500 block w-full sm:text-sm border-gray-300 rounded-md text-center" %>
+                    <div class="mt-1 flex items-center w-full">
+                      <%= form.date_field :available_on, value: developer.available_on_in_database || '', class: "shadow-sm focus:ring-gray-500 focus:border-gray-500 block w-full sm:text-sm border-gray-300 rounded-md text-center",
+                          data: {
+                            controller: 'clear-field',
+                            clear_field_target_value: 'input[name="developer[available_in_days]"]',
+                            action: 'input->clear-field#clear'
+                          }
+                      %>
+                      <%= button_tag 'Clear',
+                          name: 'clear_available_on',
+                          type: :button,
+                          class: "inline-flex justify-center align-center py-2 px-4 border border-transparent shadow-sm text-sm font-medium rounded-md text-white bg-gray-600 cursor-pointer hover:bg-gray-700 focus:outline-none focus:ring-2 focus:ring-offset-2 focus:ring-gray-500 ml-3",
+                          data: {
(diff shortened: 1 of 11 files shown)


boardfish:
I should add that while working on the Stimulus controller, I experienced issues with getting the most recent JS bundle to load, even with a force reload. I don't remember exactly what fixed it, but it might've been getting rid of `app/assets/builds` and manually rerunning `yarn build`.

joemasilotti:
> Developers may be immediately available after a given notice period (a fixed period of time).

Is that common? I'm not aware of anyone who is in that scenario unless that are trying to convey that they can start work after giving their two-weeks notice.

I'm hesitant to add something like this because it adds overhead to a form section that is already a little complicated.

boardfish:
> unless they are trying to convey that they can start work after giving their two-weeks notice.

Yeah, that's my intent with this feature. A fixed date might be good for someone who has resigned and is taking a fixed period of time off, or has to act out their notice period for a role from which they've just resigned, before searching for a new role. This allows folks who are currently employed to make their availability clear by stating that it's after a fixed notice period which they haven't yet begun.

If I'm reading you right, your concerns seem to be more with how this is presented to the user than the backend complexity. I agree that it'd be ideal to revise how this looks, potentially by only making one of those options available at a time. Here's one idea for how we could display that:
Screen Shot 2022-02-08 at 10 36 43

boardfish:
Perhaps not enough to validate this personally. We could get that kind of validation from our audiences on Twitter or in Discussions here, perhaps. You're welcome to close this if that doesn't seem like a worthwhile time investment.

joemasilotti:
I do like the idea, I just want us to understand if it is worth it. If you'd like, feel free to open a discussion or post about it on Twitter! I'll keep this open for a little while until we have some data either way.

app/views/developers/_form.html.erb
```diff
@@ -161,8 +161,43 @@
...
-                    <div class="mt-1">
-                      <%= form.date_field :available_on, class: "shadow-sm focus:ring-gray-500 focus:border-gray-500 block w-full sm:text-sm border-gray-300 rounded-md text-center" %>
+                    <div class="mt-1 flex items-center w-full">
+                      <%= form.date_field :available_on, value: developer.available_on_in_database || "", class: "shadow-sm focus:ring-gray-500 focus:border-gray-500 block w-full sm:text-sm border-gray-300 rounded-md text-center",
+                          data: {
+                            controller: "clear-field",
+                            clear_field_target_value: 'input[name="developer[available_in_days]"]',
+                            action: "input->clear-field#clear"
+                          } %>
+                      <%= button_tag "Clear",
```
gmmcal, line 171: Missing i18n

joemasilotti:
Hey again @boardfish, I'm going to close this PR for now. Feel free to open a discussion or Twitter thread to gauge interest and we can revisit.

PR #275, opened 2022-02-03
Show up-to-date developer's count number in developers index

salochara:
👋  Hey! 

**Changes in this PR:**
- Show the updated developers count number in the developers index, right next to the "Add my Profile" call to action
Developers count

**Motivation?** 

I think that, since there are now more than 200+ devs (from what I see from the developers/:id url), we should be showing that in the site. Also, it's right next to a CTA. From someone who's thinking about registering into the site, I think this makes it more attractive. 

The only concern I has while doing this, was performance on the count. I've done some tests and I believe it's benefitial, since for the next query under the index action, Pagy pulls it from cache as shown in the image below

CleanShot 2022-02-02 at 16 13 04

Let me know what you think!

### Pull request checklist


app/helpers/application_helper.rb
```diff
@@ -2,4 +2,11 @@ module ApplicationHelper
   def hotwire_livereload_tags
     super if Rails.env.development?
   end
+
+  def round_to_lower_multiple_of_ten(number)
+    return number - 1 if number <= 10
+
+    modulo = number % 10
+    number - modulo
+  end
```
joemasilotti, line 11: We can use `Developer.count.round(-1)` instead of implementing this ourselves.

joemasilotti (changes requested):
Oh, good call! I totally forgot about that part of the copy 🤦‍♂️.

app/helpers/application_helper.rb
salochara: Yeah, I also played around with it but didn't want to use it, because of these cases:
```
[5] pry(main)> 126.round(-1)
=> 130
```
It might be too much haha but that's the reason I didn't use it. 

Up to you though!

joemasilotti: I'm fine with that. If we have 195 developers and say "200+" that's a difference of 2.5% which I'm OK with. Thanks for checking though!

salochara: Done! 👍

Issue #218, opened 2022-01-06
Message email should link to sign in


coder2000:
**Describe the bug**
Currently the message email links to the sign up page. If someone is receiving a message it is safe to assume they are already signed up and a step can be skipped by linking to sign in.

**To Reproduce**
Click link in email saying you received a message.

**Expected behavior**
Link directly to the message if session is still active, otherwise redirect to sign in.

joemasilotti:
This [wiki from Devise](https://github.com/heartcombo/devise/wiki/Redirect-to-new-registration-(sign-up)-path-if-unauthenticated) is how I originally customized the redirection. I'm not sure how we can have a conditional redirection, though. Ideally, `authenticate_user!` in `NotificationsController` has a way to customize the route.

coder2000:
```
  def authenticate_user_with_sign_up!
    unless user_signed_in?
      store_location_for(:user, request.fullpath)
      redirect_to find_store_register_path, alert: "You must have an active account to access this page."
    end
  end
```
This is how I redirect for all my controllers, could adapt something like this as a before action.

joemasilotti:
Hmm, I feel like that kind of completely diverts away from Devise's internal plumbing. I'd rather continue to use Devise's method which takes care of this already (and some more).

dhairyagabha:
I would like to take a look at this as my first issue contribution, if that is okay?

joemasilotti:
Sure thing, all yours @dhairyagabha!

dhairyagabha:
Hey @joemasilotti 

I haven't been able to get to this. Please feel free to open it up. If I find a fix, I will request and open a PR.

kaushikhande:
@joemasilotti I have attached a PR for this issue. Please review.

PR #282, opened 2022-02-14
 Message email should link to sign in #218

kaushikhande:
Message email should link to sign in #218

### Pull request checklist


joemasilotti (changes requested):
This is great! I love the idea of overriding that single method in the controller.

Can you please pull out the code to a controller concern? Then it will be even more explicit (by naming it) and we can use it in other places.

joemasilotti:
Looks great, thanks for the update! I renamed the module and make a small design tweak.

PR #288, opened 2022-02-20
added invisible field to search_status enum, added visible scope to d…

GALTdea:
First PR Request! Work in progress but I wanted to submit a draft because I might need a bit of help. 
-Goal:  give developers the option to hide their profile . 
so far, i've  added 'invisible' option to the search_status and created a visible scope. I've never seen a developer_query class so I am a bit confused with it. 

…eveloper.rb

### Pull request checklist


app/models/developer.rb
```diff
@@ -23,15 +24,19 @@ class Developer < ApplicationRecord
...
+
+ 
   scope :filter_by_utc_offset, ->(utc_offset) { where(utc_offset:) }
   scope :filter_by_role_types, ->(role_types) do
     RoleType::TYPES.filter_map { |type|
       where(role_type: {type => true}) if role_types.include?(type)
     }.reduce(:or).joins(:role_type)
   end
 
+  scope :visible, -> { where(search_status:  [:open, :actively_looking, :not_interested, nil] ) }
```
joemasilotti, line 39:
```suggestion
  scope :visible, -> { where.not(search_status:  [:invisible] ) }
```

db/schema.rb
```diff
@@ -19,7 +19,7 @@
     t.string "record_type", null: false
     t.bigint "record_id", null: false
     t.bigint "blob_id", null: false
-    t.datetime "created_at", precision: 6, null: false
+    t.datetime "created_at", null: false
```
joemasilotti, line 22: I think this entire file shouldn't be committed. There's something weird going on with these precision calls (the same thing happens to me!).

app/models/developer.rb
```diff
@@ -23,15 +24,19 @@ class Developer < ApplicationRecord
   validates :bio, presence: true
   validates :time_zone, presence: true, on: :create
   validates :cover_image, content_type: ["image/png", "image/jpeg", "image/jpg"],
-    max_file_size: 10.megabytes
+  max_file_size: 10.megabytes
 
+
+
+
```
joemasilotti, line 31: This chunk can be discarded!

joemasilotti (commented):
I left a few comments and you are off to a great start!

This new field should appear in the form without any more changes. You might need to add a line to the English locale file, though (maybe).

Anywhere developers are shown we want to make sure we now use the `.visible` scope, too.

app/models/developer.rb
GALTdea: Awesome, Thanks for the feedback. Like I said, I'm not familiar with the way you set up the developers_query class, and Pagy for that matter.  I added the scope to the initialize_pagy_and_developers class. let me know what you think... I added the English and Spanish Locale.
One more thing I might need to do is deny access to developers show view when not authorize developer & invisible

app/models/developer.rb
```diff
@@ -23,7 +24,7 @@ class Developer < ApplicationRecord
   validates :bio, presence: true
   validates :time_zone, presence: true, on: :create
   validates :cover_image, content_type: ["image/png", "image/jpeg", "image/jpg"],
-    max_file_size: 10.megabytes
+  max_file_size: 10.megabytes
```
joemasilotti, line 27: Please remove this change - running `bundle exec standardrb --fix` will fix any formatting issues.

config/locales/en.yml
```diff
@@ -162,6 +162,7 @@ en:
       save: Save
       search_status:
         actively_looking_help_text: Your profile can get featured on the homepage.
+        invisible_help_text: Your profile will remain invisible to anyone else on the platform.
```
joemasilotti, line 165:
```suggestion
        invisible_help_text: Your profile is hidden and can only be seen by yourself.
```

db/schema.rb
joemasilotti: Please discard all changes to this file.

test/models/developer_test.rb
```diff
@@ -79,6 +79,11 @@ class DeveloperTest < ActiveSupport::TestCase
     refute_includes developers, developers(:unavailable)
   end
 
+  test "visible scope filters developers with 'visible' search_status" do
+    developers = Developer.visible
+    assert_includes developers, developers(:invisible)
+  end
+
```
joemasilotti, line 86: This test feels backwards, shouldn't the assertion check if the invisible developer is _not_ included in the query?

joemasilotti (changes requested):
This is shaping up nicely! Outside of the review comments there are 3 more things we need to take care of:

1. The homepage, `HomeController`, needs the new `.visible` scope.
2. Invisible developer profiles should not be added to the sitemap. `config/sitemap.rb:22` needs the new `.visible` scope.
3. Visiting a invisible profile directly via the URL should 404 _unless_ you are that user (so it is your profile).

app/controllers/developers_controller.rb
```diff
@@ -51,6 +52,15 @@ def require_new_developer!
...
   end
 
+  def invisible?
+    @developer = Developer.find(params[:id])
+    if @developer.search_status == "invisible"
+      unless user_signed_in? && current_user == @developer.user
+        redirect_to root_path, notice: t(".invisible")
+      end
+    end
+  end
```
joemasilotti, line 62: This logic can be moved to `DeveloperPolicy`, which is powered by Pundit. Check out `MessagingPolicy` for an example on something very similar.

app/views/developers/show.html.erb
```diff
@@ -1,7 +1,7 @@
 <% content_for :open_graph_tags do %>
   <%= render DeveloperOpenGraphTagsComponent.new(developer: @developer) %>
 <% end %>
-
+ <%#= current_user.developer? %>
```
joemasilotti, line 4: Please remove this.

joemasilotti (changes requested):

GALTdea:
Struggling a bit with i18n-tasks. my code passes bin/check locally, but won't pass ci/test on github.  When I run i18n-tasks health I get 167 missing translations. I'm not sure that the main branch has those translations 

locale-warning

joemasilotti:
We only care about the English translations (for now). Run this instead:

```
bundle exec i18n-tasks health en
``

**Comment by GALTdea** [comment]
Here's a rookie question. Is it ok to make some of these changes on the github branch? I've been making changes locally and re committing. also,  I'm guessing I can commit your suggestions here in github, right? 
Screen Shot 2022-02-25 at 11 57 09 AM

**Comment by joemasilotti** [comment]
Feel free to click the commit button on GitHub if it's easier for you! I don't have a preference either way.

File: test/fixtures/developers.yml [hunk 9]
```diff
@@ -45,6 +45,13 @@ with_full_time_employment:
...
   hero: A full-time employee
   bio: I am a developer available for full-time employment.
+
+invisible:
+  user: with_invisible_profile
+  name: Invisible Developer
+  hero: I'd rather stay invisible for now
+  bio: Soon to be visible
+  search_status: "invisible"
   search_status: open
```
joemasilotti, line 55: Unless I'm reading this diff wrong, it looks like the second `search_status` will override the first one.

I think `open` should be moved above (where it was) and `invisible` should be here. Also, you can drop the quotes.

test/integration/developers_test.rb
```diff
@@ -13,6 +13,17 @@ class DevelopersTest < ActionDispatch::IntegrationTest
...
 
+  test "can't view developer with invisible profile" do
+    one = developers :available
+    two = developers :unavailable
+    three = developers :invisible
+
+    get developers_path
+    assert_select "h2", text: three.hero, count: 0
+    assert_select "h2", one.hero
+    assert_select "h2", two.hero
```
joemasilotti, line 24:
```suggestion
    visible = developers :available
    invisible = developers :invisible

    get developers_path

    assert_select "h2", one.hero
    assert_select "h2", text: invisible.hero, count: 0
```

test/models/developer_test.rb
```diff
@@ -78,7 +78,6 @@ class DeveloperTest < ActiveSupport::TestCase
     assert_includes developers, developers(:available)
     refute_includes developers, developers(:unavailable)
   end
-
```
joemasilotti, line 81: Please add back this blank line.

joemasilotti (changes requested):
Looking good! I made a bunch of code requests but they are mostly small. In summary:

1. Move the invisible check in `DevelopersController` to a policy.
1. Remove all changes to the non-English locale files.
1. Add a test for visiting an invisible developer profile and it 404s then visit as the developer and it 200s.
1. A few small comments inlined.

joemasilotti (commented):
I'm making my way through the changes now. It was a lot of small changes so I decided to do them on my own. I made each change very small in its own commit so you can see the flow, if you're interested.

Bigger picture, can you add tests for `DeveloperPolicy`? I'm also stumped on why the integration test is failing, but I'm hoping to dive in later.

Other than those 2 things this looks great!

joemasilotti:
I had some time so I ended up finishing this up. Thanks again for the contribution! I'm going to give it another review this weekend before merging.

GALTdea:
Thank you! A big part of my struggle was my lack of experience with Github in collaborative settings. A 30 minutes tutorial would have saved me a lot of headaches. 
The next PR will go much smoother.

PR #291, opened 2022-02-22
Fix menu flashing on full page reload

### LINKED ISSUE #284: Menu flashing on full page load
**When doing a full page refresh on `/developers`, the sort and filter menus momentarily flash**

**Reproduce**
1. Go to `/developers`
2. Hit refresh
3. Notice that -for a brief second- all filtering menus are fully open 

**Expected behavior**
The menus should be hidden and only open up if the user clicks on them

**Video/Proof**
Turned it down to "Slow 3G" for demonstration. Just wait for it.

https://user-images.githubusercontent.com/1859806/154079769-dfa7fcf8-4484-40a3-a30b-67e973747100.mp4

**Additional context**
The problem seems to be located somewhere [along these lines]( https://github.com/joemasilotti/railsdevs.com/blob/61105c0363596d885ccd6b2ffa18d10b359c88ec/app/components/developer_query_component.html.erb#L15). What's effectively happening here is that we load the page without the `hidden` class present and only after stimulus connects, `#toggle` is called and `hidden` is added. That's why we see the flash: It's actually the interval before JS kicks in. It should be the other way around though: `hidden` should be set in the DOM on page load and only by clicking on the menu stimulus should remove the class. 

I think this also explains the fact that during page navigation you don't see any issues (stimulus is already loaded in memory and `#toggle` is instant). 

If you're interested @joemasilotti, I could prepare a PR for it (at least for the `/developers` page).

kstratis:
Fixes #284 

### Pull request checklist

- [x] Your code contains tests relevant for code you modified - No tests needed here

### DIFF
--- a/app/components/developer_query_component.html.erb
+++ b/app/components/developer_query_component.html.erb
@@ -12,7 +12,7 @@
             </button>
           </div>
 
-          <div data-toggle-target="element" class="origin-top-left absolute left-0 z-10 mt-2 w-40 rounded-md shadow-2xl bg-white ring-1 ring-black ring-opacity-5 focus:outline-none" role="menu" aria-orientation="vertical" aria-labelledby="mobile-menu-button" tabindex="-1">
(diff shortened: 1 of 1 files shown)


joemasilotti:
Looks great, thanks for the contribution!

PR #295, opened 2022-02-24
Anonymize avatar filenames

joemasilotti:
This PR fixes a small security issue where attachment filenames could leak personal information about the developer or business. If, for example, I upload my avatar with the filename `joemasilotti.jpg` then someone could see that filename (and therefore my name) in their browser's DOM or network inspector.

The code changes rename every avatar upload to `avatar.jpg`, where `.jpg` is the original filename extension. It also includes a backfill to fix existing records.

### Pull request checklist

### DIFF
--- a/app/models/concerns/avatarable.rb
+++ b/app/models/concerns/avatarable.rb
@@ -13,7 +13,9 @@ module Avatarable
     private
 
     def anonymize_avatar_filename
-      avatar.blob.filename = "avatar#{avatar.filename.extension_with_delimiter}"
+      if avatar.attached?
+        avatar.blob.filename = "avatar#{avatar.filename.extension_with_delimiter}"
+      end
     end
   end
 end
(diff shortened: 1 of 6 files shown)


jacobdaddario:
Ah! What a great catch. I wonder if this could be made as a PR to ActiveStorage in Rails?

joemasilotti:
Good question! I could see it being an option on the `has_attached` helper.

jacobdaddario:
Looking at the code for that, I don't think it would be too hard to add it as a side-effect to blob creation. I'll earmark this and maybe dive into it one day. Lemme know if you get into it yourself, I'd be happy to do code review.

joemasilotti:
That sounds way better than what I've written! Are side effects of blobs a built in feature / API? I'd love to learn more about it.

jacobdaddario:
Eating some breakfast and saw this notification. Just wanted to shoot back a quick response.

I might have misused the word side effect. `Blob` is just a Rails model, so I was trying to say that we could add a `before_create` action on the model there. The `filename` column lives on `Blob` as seen [here](https://github.com/rails/rails/blob/d6e6686c5f8b21ea806eb27984e86f7e0fa7bba2/activestorage/db/migrate/20170806125915_create_active_storage_tables.rb#L8). 

The tough part of this PR would be how we pass a configuration option from `has_*_attached` to a blob record, especially since variant records are also backed by blobs.

PR #298, opened 2022-02-25
Clean up Open Graph meta tags

joemasilotti:
This PR cleans up how/when Open Graph tags are rendered and tweaks the copy on a few key pages.

Current failing tests are because I can't figure out how to test `content_for` in a View Component.

### Pull request checklist

### DIFF
--- a/app/helpers/meta_helper.rb
+++ b/app/helpers/meta_helper.rb
@@ -1,5 +1,6 @@
 module MetaHelper
   def open_graph_tags
-    content_for :open_graph_tags
+    render(OpenGraphTagsComponent.new) unless content_for?(:open_graph_tags)
+    content_for(:open_graph_tags)
   end
 end
--- a/app/helpers/meta_helper.rb
+++ b/app/helpers/meta_helper.rb
@@ -1,6 +1,0 @@
-module MetaHelper
-  def open_graph_tags
-    render(OpenGraphTagsComponent.new) unless content_for?(:open_graph_tags)
-    content_for(:open_graph_tags)
-  end
-end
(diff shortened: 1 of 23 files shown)


app/helpers/meta_helper.rb
```diff
@@ -1,5 +1,6 @@
 module MetaHelper
   def open_graph_tags
-    content_for :open_graph_tags
+    render(OpenGraphTagsComponent.new) unless content_for?(:open_graph_tags)
+    content_for(:open_graph_tags)
```
radar, line 4: Following from your Twitter thread here: https://twitter.com/joemasilotti/status/1497341331315650560?s=20&t=nc7y6ITf4HnEDPGLhai9LA.

How about:

```
render(OpenGraphTagsComponent.new, tags: content_for(:open_graph_tags)
```

Then in the component itself, return the tags if they're present, otherwise render the component's view.

joemasilotti: Maybe I'm missing something, but doesn't this mean I have to add back in the following block every time I render the component?

```
<% content_for(:open_graph_tags) do %>
  <%= render OpenGraphComponent.new(...) %>
<% end %>
```

joemasilotti:
For those following along, I went with something a bit different. Instead of rendering the `content_for` in the component I've moved that logic to a helper that then renders the component. The helper passes all the arguments down to the component and has an option to render a more specific Open Graph component.

```
def open_graph_tags(options = {})
  content_for(:open_graph_tags) do
    component = options.delete(:component) || OpenGraphTagsComponent
    render component.new(**options)
  end
end
```

PR #314, opened 2022-03-08
Pricing v2

joemasilotti:
This PR introduces a second pricing tier to railsdevs. The goal is two-fold:

1. Better incentive for me to help businesses hire devs fast
2. Keep railsdevs accessible to bootstrappers and solopreneurs.

Pricing v2

### Tier 1: Hire freelancers and contractors

* $99/month

This is largely unchanged from the current plan, but you can no longer hire full-time employees (only freelancers and contractors).

### Tier 2: Hire full-time employees

* $299/month
* 10% of hired dev's first year salary

This plan will also include a concierge onboarding service, priority video call support, and hand-picked developers from me every month.

### Why?

Because there's currently a weird dynamic with a monthly subscription. To increase revenue I'm incentivized to keep you on the platform as long as possible. But I want to help businesses hire devs as fast as possible! Which goes direct against that.

An additional payout on full-time hires realigns our goals: hire their next developer, fast. I'm encouraged to help as much as possible without worrying about cutting into revenue.

I'm still dedicated to being the best place for bootstrappers and solopreneurs find freelance Rails developers. Keeping that plan at $99/month ensures it remains approachable for one-off help and using the service only when needed.

Existing customers will be legacy'd in to the original plan. They can continue to pay $99/mo and hire whoever they want without paying the 10% as long as they keep their subscription active and paid. 

This is an experiment. Will it crash and burn? Maybe! But if it works then railsdevs becomes one step closer to becoming a successful open source business. 💪

### To-dos

- [x] Add two tiers to `/pricing`
- [x] Revamp design of `/pricing`
- [x] Add new plans to Stripe (test mode)
- [x] Copy over plans to Stripe (production mode)
- [x] Add plans to credentials (production)
- [x] Only allow purchases from `/pricing` (revamp the checkout flow)
- [x] "Hire me" button click -> checkout -> cold message form (ensure flow still works)
- [x] Feature flag v2 `/pricing`
- [x] Banner for existing customers

### Pull request checklist


test/integration/stripe_checkout_test.rb
```diff
@@ -0,0 +1,34 @@
...
+
+  test "redirects to the stored location on success" do
+    success_path = new_developer_message_path(Developer.first)
+    get success_path
+
+    stub_pay(@user) do
+      post stripe_checkout_path
+      assert_equal Analytics::Event.last.url, success_path
+    end
+  end
```
joemasilotti, line 33: This test is failing and I can't figure out why. The `stub_pay` (from `test/support/helpers/pay_helper.rb`) isn't picking up on the `user` stub. I _think_ because we stub `@user` but the controller fetches via `current_user` so they aren't the same instance (although they are the same record).

I also can't set `session` directly until a request is made. Which might be a red herring now that I type it out.

PR #334, opened 2022-03-17
Revamp messaging with Hotwire

### LINKED ISSUE #333: Hotwire-ify messaging
The messaging flow is a little rough around the edges and could be improved with a little Hotwire.

This idea isn't super concrete, yet. But here's what I'm thinking.

1. When sending a message, render a Turbo Stream to append the message to the conversation and re-render the form.
2. Handle errors with a flash message as if it wasn't submitted via Turbo.
3. Ensure if JavaScript is disabled that everything still works.
4. Stretch goal: Show new messages via Action Cable on the receiving end.

joemasilotti:
This PR closes #333. It uses a Turbo Stream to send messages and append to the list.

### Pull request checklist

### DIFF
--- a/app/controllers/messages_controller.rb
+++ b/app/controllers/messages_controller.rb
@@ -3,7 +3,6 @@ class MessagesController < ApplicationController
   before_action :require_active_subscription!
 
   def create
-    @conversation = conversation
     @message = Message.new(message_params.merge(conversation:, sender:))
     authorize @message, policy_class: MessagingPolicy
 
@@ -26,7 +25,7 @@ class MessagesController < ApplicationController
   end
 
   def conversation
-    Conversation.visible.find(params[:conversation_id])
+    @conversation ||= Conversation.visible.find(params[:conversation_id])
   end
 
   def developer
--- a/app/controllers/messages_controller.rb
+++ b/app/controllers/messages_controller.rb
@@ -3,13 +3,16 @@ class MessagesController < ApplicationController
   before_action :require_active_subscription!
 
   def create
+    @conversation = conversation
     @message = Message.new(message_params.merge(conversation:, sender:))
     authorize @message, policy_class: MessagingPolicy
 
     if @message.save
-      redirect_to conversation
+      respond_to do |format|
+        format.turbo_stream { @new_message = conversation.messages.build }
+        format.html { redirect_to conversation }
+      end
     else
-      @conversation = conversation
       render "conversations/show", status: :unprocessable_entity
     end
   end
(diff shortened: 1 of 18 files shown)


app/controllers/messages_controller.rb
```diff
@@ -3,13 +3,16 @@ class MessagesController < ApplicationController
   before_action :require_active_subscription!
 
   def create
+    @conversation = conversation
```
ParamagicDev, line 6: You could also memoize this in your method call:

```rb
def conversation
  @conversation ||= conversation
end
```

And then you dont need to do this top level assignment!

app/views/conversations/show.html.erb
```diff
@@ -20,15 +20,13 @@
     <% end %>
   </div>
 
-  <ul role="list" class="mt-8 space-y-2">
+  <ul id="messages" role="list" class="mt-8 space-y-2">
```
ParamagicDev, line 23: `<ul>` actually comes with an implicit `role` of `"list"` so you dont need to specify it again! (it can sometimes even cause unexpected behavior when you add a role to an element that has the same implicit role)

https://developer.mozilla.org/en-US/docs/Web/HTML/Element/ul
https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Roles/list_role

app/controllers/messages_controller.rb
```diff
@@ -3,13 +3,16 @@ class MessagesController < ApplicationController
...
     @message = Message.new(message_params.merge(conversation:, sender:))
     authorize @message, policy_class: MessagingPolicy
 
     if @message.save
-      redirect_to conversation
+      respond_to do |format|
+        format.turbo_stream { @new_message = conversation.messages.build }
+        format.html { redirect_to conversation }
```
ParamagicDev, line 13: I know this is a `create` request so POST requests dont fall victim to it, but I've been making it a habit of specifying `status: :see_other` or `status: 303` just in case! (this is merely personal preference)

ParamagicDev (approved):
This looks great! Streams look good! Nice work Joe!

app/controllers/messages_controller.rb
joemasilotti: Ah, good call! I forgot that the next line calls into that method anyway.

app/views/conversations/show.html.erb
joemasilotti: Huh! TIL, thank you!

joemasilotti:
Thank you! I appreciate the review - merging in your changes after CI passes. :D

Issue #327, opened 2022-03-14
Enable features for admins


joemasilotti:

kaushikhande:
Can I work on this?

joemasilotti:
You got it!

kaushikhande:
@joemasilotti Please review

joemasilotti:
Closed via #327.

PR #336, opened 2022-03-18
Issue#327 Enable features for admins

kaushikhande:
[Issue-327](https://github.com/joemasilotti/railsdevs.com/issues/327) Enable features for admins #327 
### Pull request checklist

### DIFF
--- a/app/models/feature.rb
+++ b/app/models/feature.rb
@@ -1,5 +1,5 @@
 class Feature
-  def self.enabled?(feature_name:, user: nil)
+  def self.enabled?(feature_name, user: nil)
     case feature_name.to_sym
     when :new_developer_fields_banner
       true
--- a/app/models/feature.rb
+++ b/app/models/feature.rb
@@ -1,12 +1,12 @@
 class Feature
-  def self.enabled?(feature_name)
+  def self.enabled?(feature_name:, user: nil)
     case feature_name.to_sym
     when :new_developer_fields_banner
       true
     when :pricing_v2
       true
     when :role_level_filters
-      !Rails.env.production?
+      !Rails.env.production? || user&.admin?
     end
   end
 end
--- a/app/models/feature.rb
+++ b/app/models/feature.rb
@@ -1,12 +1,12 @@
 class Feature
-  def self.enabled?(feature_name, user: nil)
+  def self.enabled?(feature_name, user:)
     case feature_name.to_sym
     when :new_developer_fields_banner
       true
     when :pricing_v2
       true
     when :role_level_filters
-      !Rails.env.production? || user&.admin?
+      !Rails.env.production? || user.admin?
     end
   end
 end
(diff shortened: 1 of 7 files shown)


app/models/feature.rb
```diff
@@ -1,12 +1,12 @@
 class Feature
-  def self.enabled?(feature_name)
+  def self.enabled?(feature_name:, user: nil)
```
joemasilotti, line 2:
```suggestion
  def self.enabled?(feature_name, user: nil)
```

I don't think we need to name the first variable since it _is_ the feature we are switching on.

test/components/developer_query_component_test.rb
```diff
@@ -47,9 +52,37 @@ class DeveloperQueryComponentTest < ViewComponent::TestCase
...
+      render_inline DeveloperQueryComponent.new(query:, user:)
+    end
+
+    assert_text RoleLevel.human_attribute_name("junior")
+    assert_text RoleLevel.human_attribute_name("senior")
+    assert_text RoleLevel.human_attribute_name("principal")
+    assert_text RoleLevel.human_attribute_name("mid")
+    assert_text RoleLevel.human_attribute_name("c_level")
+  end
+
```
joemasilotti, line 82: While I appreciate these tests I don't think we want them. I'd rather have the `Feature` class not require test changes when updated.

joemasilotti (changes requested):
Two small requests but other than that this looks good to me!

app/models/feature.rb
kaushikhande: done

test/components/developer_query_component_test.rb
kaushikhande: done

kaushikhande:
@joemasilotti Please review. Added changes.

app/models/feature.rb
```diff
@@ -1,12 +1,12 @@
 class Feature
-  def self.enabled?(feature_name)
+  def self.enabled?(feature_name, user: nil)
```
joemasilotti, line 2: Please remove the defaulted parameter here. It needs to be required so it doesn't slip through. All components using `Feature` will need to pass in a user.

app/models/feature.rb
```diff
@@ -1,12 +1,12 @@
...
     case feature_name.to_sym
     when :new_developer_fields_banner
       true
     when :pricing_v2
       true
     when :role_level_filters
-      !Rails.env.production?
+      !Rails.env.production? || user&.admin?
```
joemasilotti, line 9:
```suggestion
      !Rails.env.production? || user.admin?
```

After requiring `user`, from the comment above.

joemasilotti (changes requested):
Not sure why you changed the defaulted parameter, but once that gets changed back we can merge!

kaushikhande:
Done

joemasilotti (approved):

joemasilotti:
Looks great, thank you for the contribution!

PR #340, opened 2022-03-20
improve system tests for filters by matching against regex instead of string

themudassarhassan:
This PR is aiming to improve the assertions of filters system test by matching the `current_path` against a regex of expected query param instead of whole string. More information why this change is needed can be found [here](https://github.com/joemasilotti/railsdevs.com/discussions/338).

### Pull request checklist


joemasilotti:
Love it, great idea using a regex to match. Thank you!

PR #342, opened 2022-03-21
Conversation scroll

jasonplatts:
Pull request adds scrolling to the message form when the conversation page loads.

### Pull request checklist

### DIFF
--- a/app/javascript/controllers/index.js
+++ b/app/javascript/controllers/index.js
@@ -9,5 +9,8 @@ application.register("analytics--events", Analytics__EventsController)
 import FileUploadController from "./file_upload_controller.js"
 application.register("file-upload", FileUploadController)
 
+import PageScrollController from "./page_scroll_controller.js"
+application.register("page-scroll", PageScrollController)
+
 import ToggleController from "./toggle_controller.js"
 application.register("toggle", ToggleController)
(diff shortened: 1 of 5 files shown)


joemasilotti:
Thanks for this! I love the idea of scrolling to the bottom when loading a conversation.

But I'm wondering if this is possible without JavaScript. For example, can we use an anchor to have the browser automatically scroll to the message form at the bottom?

jasonplatts:
No problem. Thank you for reviewing the pull request! You had a great suggestion. I tried to implement it, but ran into a couple of issues.

First, Turbo seems to be interfering with navigating to the anchor. It will initially jump to the correct location, then reload and jump to the top of the page again. The screen below shows the behavior when clicking on the conversation link generated by this code, `link_to conversation_path(conversation, anchor: "message_body")`

anchor-example

The second issue is that it doesn't seem to be possible to include the anchor with the submission of the message form without using JavaScript.

Do you have any thoughts or suggestions?

joemasilotti:
Huh, good to know! Do you know if this is a documented bug or something with the configuration/rendering of railsdevs?

jasonplatts:
Hi Joe. I sent you a message on this PR. When the form was hotwired, the scrolling stopped working on the initial page load. I'm trying to get some help figuring out the problem.

jasonplatts:
Sorry Joe. This is not a problem with the turbo frame, but rather turbo in general I think.

See below:
2022-03-30 - railsdev com scroll issue

joemasilotti:
Shout out to @ParamagicDev for helping me get this one over the finish line!

PR #348, opened 2022-03-23
Redesign `/developers` with new filters

joemasilotti:
The filters on `/developers` are getting a little crowded and we are quickly running out of horizontal screen space. This draft PR moves the filters to a second column along the left (mobile stays in a "drawer" that slides in).

image

Some considerations/problems/ideas:

1. How do we submit the form? This design doesn't work well with submit buttons. Perhaps autosubmit on change?
2. If we go that route it could be worth it to covert this to a Turbo Stream and re-render the list of developers, only.
3. If not, we will need to remember the state of the opened/closed filtering panels when the page is redirected after the form submission.
4. We could leave the important categories "open" on initial page load (Work preference and Role level) to encourage using them.
5. Search (from #341) will fit in nicely right above the first filter. Same "Submit problem" as above, but that could be debounced for 2s while someone is typing before submitting.

### Pull request checklist

- [ ] You have added significant changes and product updates to the [changelog](CHANGELOG.md)


joemasilotti:
Closed in favor of #410.

Issue #215, opened 2022-01-05
Brand color


joemasilotti:
Gray looks great (ha), but I think it's time for a real brand color. I've toyed with the idea of red to match Ruby/Rails, but am open to other ideas.

Also, ideally this would be one of the 17 non-gray [Tailwind CSS 3.0 colors](https://tailwindcss.com/docs/customizing-colors) so we can use all the 10 variants out of the box. If not, we would need to manually create 10 variants that work with Tailwind UI.

joemasilotti:
On second thought, I'm going to take GitHub's lead and make black/gray the "brand" color. This leaves open opportunities to use green, red, and yellow as accent colors for positive, negative, and neutral context, respectively. We could even add in blue for links!

PR #349, opened 2022-03-24
Change colors to buttons and links from gray to dark blue

anjolovic:
I've created this PR for brand colors only. I swapped colors to all buttons from gray to dark blue, and did the same for links, in some places but not all. Thoughts?

### DIFF
--- a/app/components/developer_profile_action_component.html.erb
+++ b/app/components/developer_profile_action_component.html.erb
@@ -1,11 +1,11 @@
 <div class="mt-4 flex space-x-3 md:mt-0">
   <% if update_profile? %>
-    <%= link_to edit_developer_path(user.developer), class: "inline-flex items-center px-4 py-2 border border-transparent shadow-sm text-sm font-medium rounded-md text-white bg-gray-600 hover:bg-gray-700 focus:outline-none focus:ring-2 focus:ring-offset-2 focus:ring-gray-500" do %>
+    <%= link_to edit_developer_path(user.developer), class: "inline-flex items-center px-4 py-2 border border-transparent shadow-sm text-sm font-medium rounded-md text-white bg-blue-900 hover:bg-blue-800 focus:outline-none focus:ring-2 focus:ring-offset-2 focus:ring-gray-500" do %>
       <%= inline_svg_tag "icons/solid/pencil.svg", class: "-ml-1 mr-2 h-5 w-5" %>
       <%= t(".edit_profile") %>
     <% end %>
   <% else %>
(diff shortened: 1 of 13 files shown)


joemasilotti:
Thanks for kicking this off. I've been thinking about adding an accent color or two for a long time. #215 was created in the beginning of January!

From you recommendation, I dove through the Color Palettes part of the Refactoring UI book again. Of the 14 palettes, these are the ones that I'm excited about.

### Palette 1
Palette 1

I love the dark banner background and how it bleeds into the navigation bar. That is something I think I want to keep with a redesign. I think this might be the most "boring" of them all, though. The blue is very uninspiring.

### Palette 4
Palette 4

The biggest curve ball of the bunch: green! Green has kind of become my brand color (see: my avatar, my website, my house, etc.) so I would love to keep that going with railsdevs. The downside is that green doesn't feel very Ruby/Rails nor job board-y. But maybe that's a good thing to differentiate.

### Palette 6
Palette 6

The opposite of 4: this is the most Ruby/Rails. It's red. But it's also hard to make red look good unless you are very careful. I'm also not a huge fan of the yellow but I can see why it works. Probably my least favorite of the four.

### Palette 9
Palette 9

Feels like a better version of 1 to me. But, it is more gray/boring which is what I'm trying to move _slightly_ away from. I love the little pink accent colors, too.

---

Typing this all out makes me think I'm deciding between 4 (green) and 9 (dark/cyan). Unless the community feels strongly about going with 6 (the red one)!

I'd love to hear your thoughts, @anjolovic. And anyone else who would like to comment!

joemasilotti:
> The way I see these layouts and colors are two folds. ... All of the palettes above are application UI, which is the feature-rich layout.

Great point! I'll definitely want to mess around with some marketing page examples from Tailwind UI using the color palette we end up choosing.

> I like 1 and 4, which is super elegant and simple. Not so crazy about 6. But definitely love 9, which combines 1 and 4 with bold colors, like pink, but settle enough to invite 👍

Looking at this again after a night of sleep... I'm really liking 9. I think that using green might be tough in the long run so sticking to blues/grays with some splashed in pink could work really well.

anjolovic:
Should I create another PR pertaining only to Marketing/Landing Page UI, and another for Application UI?

jacobdaddario:
Yo!! Where did you get those mock-ups? They're amazing. Are those proposed future views for the home page?

joemasilotti:
> Yo!! Where did you get those mock-ups? They're amazing. Are those proposed future views for the home page?

They are the screenshots from the Color Palettes PDF in the  [Refactoring UI](https://www.refactoringui.com) Complete Package.

jacobdaddario:
Ohhh! I misunderstood what you were saying. I didn't realize that was part of the full package for Refactoring.

joemasilotti:
Thanks again for diving into this, Alex. I'm going to close the PR for now as it's 4+ months old and gotten quite stale.

I have some design ideas in the works and will be sharing publicly soon. Stay tuned!

Issue #358, opened 2022-04-01
Show warning when part-time plan reaches out to full-time only developer


joemasilotti:
Businesses that have signed up for the $99/month part-time plan are not allowed to hire full-time developers.

When reaching out to a developer, if the business is on the part-time plan and the developer is not looking for part-time contract roles...

- [ ] Show a warning/info message and disable the form on the cold message page and conversation page 
- [ ] Enforce logic in `MessagingPolicy`

joemasilotti:
All yours!

GALTdea:
got the logic working from the cold_messages_controller but I can't get it to work on the MessagingPolicy. going over the pundit docs to see how to access the developer record since the record variable it's an empty message object. not experienced on pundit.

joemasilotti:
Exciting! Take a look at the other policy classes - there should be an example that shows what you need to get close.

joemasilotti:
Closed via #380.

PR #369, opened 2022-04-09
added logic to messaging_policy with business privileges and added f… #358

GALTdea:
I have a couple of questions regarding your specs: 
1) Show a warning/info message and disable the form on the cold message page and conversation page
- Do you want the part-time business account user to have access to the message view? or can the be redirected upon clicking the "Hire Me" link on the developer show page?  

…ull_time method to user.rb

### Pull request checklist

- [ ] You have added significant changes and product updates to the [changelog](CHANGELOG.md)


GALTdea:
I don't think i was supposed to open a new (draft) PR. IN any case, I just wanted to open up a draft because I had  questions regarding your specs. "Show a warning/info message and disable the form on the cold message page and conversation page"
Do you want the part-time business account user to have access to have access to the message(s) pages? or can they be redirected upon clicking the "Hire Me" link on the developer show page?

joemasilotti:
> Do you want the part-time business account user to have access to have access to the message(s) pages? or can they be redirected upon clicking the "Hire Me" link on the developer show page?

The part-time business should have access to the cold message view but the form should be disabled with the warning message being shown. The `Messaging` policy should also enforce _creating_ a message for that business to that developer in case someone gets clever and POSTs directly to the endpoint bypassing the form.

GALTdea:
Ok, I should be getting closer but I'd like for you to take a look at my approach to disabling the form. basically i created a variable '@disabled' in the controller to keeps track on its given Policy.   Currently trying to figure out how not to redirect from the new view and and show the warning message instead when businesses are denied. 
looks like pundit recognizes it from the en.yml if i use the policy/method name as a key? i'll give it a try in the morning. That would solve for redirect to since I would not use the new? method.

joemasilotti:
I took a pass at reworking some of your code with a new approach. I'm trying a new policy, `SubscriptionPolicy`, that isn't called via `authorize`. Instead, the method is called directly. Yes, it kind of defeats the purpose of Pundit but it does keep all the policy-like code in the same directory.

I'm doing this so we don't raise the Pundit exception but set an instance variable to true/false. Alternatively, we could raise and display the message in an alert back on the developer page. But it might be hard to tap into the message output because we would need HTML.

Anyway, here's what's left to get this deployed:

- [ ] Add test for when @messageable is false in ColdMessagesTest.
- [ ] Add test for when @messageable is false in MessagesTest.
- [ ] Add tests for SubscriptionPolicy.
- [ ] Authorize SubscriptionPolicy in MessagesController#create.

joemasilotti:
Making progress here, only two things left I think:

- [ ] Show error alert in place of form on `Conversations#show`
- [ ] Ensure updating a plan via Stripe Billing Portal is reflected in database

GALTdea:
The new messaging_policy makes a lot more sense than my method on messaging_policy.  I am glad you added the messaging. I spent a couple days trying to figure how to enforce the messaging from the policy, pundit's not very flexible with customizing exceptions

joemasilotti:
OK, I think this is good to go! Everything is working as I expect it to and all the tests are green. I'm going to give the code another review tomorrow (or later this week) and merge it in.

joemasilotti:
Thanks for kicking this off, @GALTdea!

PR #320, opened 2022-03-12
Draft: mark all notifications as read

gathuku:
### Pull request checklist

- [x] Your code contains tests relevant for code you modified
- [ ] You have linted and tested the project with `bin/check`

closes #316 

UI
Screen Shot 2022-03-12 at 12 18 50 PM

PR #374, opened 2022-04-12
Fix OpenStartup::ReportingTest (Timezone issue)

padi:
Likely fixes the same testing issue described in [#320](https://github.com/joemasilotti/railsdevs.com/pull/320#issuecomment-1065902204)

#to_time, by default, uses the `:local` time (varies per machine),
has timezone data. `#to_i` offsets this when converting to integer… 
but the timezone information at this point is lost!
	
`OpenStartup::Reporting#create_or_update_transaction`
doesn't convert it back to the correct timezone	

Result: if you’re living on the right side of UTC (starting from UTC+1),
some transactions appear on january reports instead of february

Solution: remove test dependency to any dev machine's
`:local` time and `use to_time(:utc)` instead

### DIFF
--- a/test/models/open_startup/reporting_test.rb
+++ b/test/models/open_startup/reporting_test.rb
@@ -117,11 +117,11 @@ class OpenStartup::ReportingTest < ActiveSupport::TestCase
   end
 
   def january
-    (today - 1.day).to_time.to_i
+    (today - 1.day).to_time(:utc).to_i
   end
 
   def february
-    today.to_time.to_i
+    today.to_time(:utc).to_i
   end
 
   def today


padi:
If you're on the western side of UTC,
you can simulate this behavior
by changing your `TZ` env variable i.e.

```ruby
# in reporting_test.rb
setup do
  ENV["TZ"] = "Etc/UTC-1"`
end

teardown do
  ENV["TZ"] = nil
end
```

or in `bin/rails c`

```ruby
[2] pry(main)> def today
  Date.new(2022, 2, 1)
end
=> :today
[4] pry(main)> ENV["TZ"] = "Etc/UTC-1"
=> "Etc/UTC-1"
[5] pry(main)> today.to_time(:local)
=> 2022-02-01 00:00:00 +0100
[6] pry(main)> today.to_time(:utc)
=> 2022-02-01 00:00:00 UTC
[7] pry(main)> today.to_time(:local) - today.to_time(:utc)
=> -3600.0 # local is offset by 1 hour compared to utc
```

joemasilotti:
Love it! Thanks for diving in and providing a fix.

Issue #313, opened 2022-03-08
Clean up ignored i18n keys 


joemasilotti:
Currently, the i18n-tasks configuration, `config/i18n-tasks.yml:112`, is ignoring a bunch of keys. These keys are dynamic, so the `health` command can't pick up on them. Instead, we should be marking the dynamic keys as such where they are used.

https://github.com/glebm/i18n-tasks#dynamic-keys

A PR addressing this issue should:

1. Remove all `ignore_unused` keys from the configuration
2. Ensure `bundle exec i18n-tasks health en` still passes

padi:
@joemasilotti looks like a good first issue to familiarize myself with the codebase.

Just tried removing ignored_unused and there's a total of 140 unused/undetected keys.

Are you fine with one big PR on this one?

joemasilotti:
Awesome, thanks for kicking this off! One PR is fine as I _think_ it will only require a bunch of special comments in the codebase.

joemasilotti:
Hey @padi, do you still have appetite to get this over the finish line? If not, let me know and I can pass it on to someone else or take it over myself.

padi:
I was able to remove some of the ignored files but stopped it midway. I even upgraded to the latest version of i18n-tasks, where one of the recent addition if a new scanner might come in handy.

I can put up a PR in a few hours if you want and then continue from there. Just typing on mobile right now.

joemasilotti:
Sure thing, that would be great! Thank you.

padi:
@joemasilotti yes you can assign this to someone else.

pkayokay:
I'd love to pick this up!

joemasilotti:
Excellent - all yours!

joemasilotti:
For now, now. If they are getting in the way of running the health check then you can remove them. I'm working on revamping translations soon.

pkayokay:
Looks like we're not using devise's lockable feature, is it safe to assume that any translations related to that should be removed?

pkayokay:
Other translations that we don't actually use, here's a few. I can have a PR for a more detailed review.
- `devise.registrations.edit.cancel_my_account` -> `Cancel my account`
- `devise.registrations.edit.leave_blank_if_you_don_t_want_to_change_it` -> `leave blank if you don't want to change it`
- `devise.registrations.edit.unhappy` -> `Unhappy`

joemasilotti:
[...]

PR #385, opened 2022-04-24
Unignore i18n tasks

padi:
Kickstarts #313 

From 140 -> 117 unused keys (if I remove all ignore_unused in i18n-tasks.yml)

Leaves the activerecord and devise-related translations ignored by i18n-tasks.

### DIFF
--- a/app/views/shared/_error_messages.html.erb
+++ b/app/views/shared/_error_messages.html.erb
@@ -2,6 +2,9 @@
   <% content_for :flash do %>
     <%= render ToastMessageComponent.new(icon: "x_circle", color: :red, messages_tag: :ul) do |c| %>
       <%= c.title do %>
+        <%# i18n-tasks-use t('activerecord.models.business') %>
+        <%# i18n-tasks-use t('activerecord.models.developer') %>
+        <%# i18n-tasks-use t('activerecord.models.user') %>
         <%= I18n.t "errors.messages.not_saved", count: resource.errors.count, resource: resource.class.model_name.human.downcase %>
       <% end %>
 
--- a/config/i18n-tasks.yml
+++ b/config/i18n-tasks.yml
@@ -87,6 +87,15 @@ search:
   ## If `strict` is `false`, guess usages such as t("categories.#{category}.title"). The default is `true`.
   # strict: true
 
+  ## Allows adding ast_matchers for finding translations using the AST-scanners
+  ## The available matchers are:
+  ## - RailsModelMatcher
+  ##     Matches ActiveRecord translations like
+  ##     User.human_attribute_name(:email) and User.model_name.human
+  ##
+  ## To implement your own, please see `I18n::Tasks::Scanners::AstMatchers::BaseMatcher`.
+  <%# I18n::Tasks.add_ast_matcher('I18n::Tasks::Scanners::AstMatchers::RailsModelMatcher') %>
+
   ## Multiple scanners can be used. Their results are merged.
   ## The options specified above are passed down to each scanner. Per-scanner options can be specified as well.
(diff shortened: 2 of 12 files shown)


app/views/shared/_error_messages.html.erb
```diff
@@ -2,6 +2,9 @@
   <% content_for :flash do %>
     <%= render ToastMessageComponent.new(icon: "x_circle", color: :red, messages_tag: :ul) do |c| %>
       <%= c.title do %>
+        <%# i18n-tasks-use t('activerecord.models.business') %>
+        <%# i18n-tasks-use t('activerecord.models.developer') %>
+        <%# i18n-tasks-use t('activerecord.models.user') %>
```
joemasilotti, line 7: Is this saying the entire scope is used? I think it should be a bit more specific, like `activerecord.models.business.errors.messages.not_saved`, right? Or maybe I'm misunderstanding how this is working.

config/i18n-tasks.yml
```diff
@@ -87,6 +87,16 @@ search:
...
 
+  ## Allows adding ast_matchers for finding translations using the AST-scanners
+  ## The available matchers are:
+  ## - RailsModelMatcher
+  ##     Matches ActiveRecord translations like
+  ##     User.human_attribute_name(:email) and User.model_name.human
+  ##
+  ## To implement your own, please see `I18n::Tasks::Scanners::AstMatchers::BaseMatcher`.
+  <% require 'i18n/tasks/scanners/ast_matchers/rails_model_matcher' %>
+  <% I18n::Tasks.add_ast_matcher('I18n::Tasks::Scanners::AstMatchers::RailsModelMatcher') %>
```
joemasilotti, line 98: Which translations does this match?

joemasilotti (changes requested):
Thanks for kicking this off! I posted a few small questions in the code.

app/views/shared/_error_messages.html.erb
padi: It's being used in the latter part of the line.

```rb
 <%= I18n.t "errors.messages.not_saved", count: resource.errors.count, resource: resource.class.model_name.human.downcase %>
```

specifically:

```rb
resource.class.model_name.human.downcase
```
[According to Rails Guides (i18n):](https://guides.rubyonrails.org/i18n.html#active-model-methods)

> `model_name.human` and `human_attribute_name` use translations for model names and attribute names if available in the activerecord.models scope


Usage:

```rb
pry(main)> I18n.with_locale(:es) { Developer.model_name.human }
=> "Desarrollador"
pry(main)> I18n.with_locale(:es) { I18n.t "errors.messages.not_saved", count: 2, resource: Developer.model_name.human }
=> "Ocurrieron 2 errores al tratar de guardar Desarrollador:"
```

config/i18n-tasks.yml
padi: I wanted to check if it will match at least some of the unused keys, but it seems that it didn't and forgot to remove comment this back.

joemasilotti: Gotcha, let's remove it then.

app/views/shared/_error_messages.html.erb
joemasilotti: But is your comment saying that the _entirety_ of that resource is being used? Or only the "not saved" error message?

padi: Ah, I see. 

More specifically, it's using `activerecord.models.(model).one`

joemasilotti: Got it, thanks for clarifying!

PR #390, opened 2022-04-29
fix misaligned form field

MichaelArriaga:
### Pull request 
Fix misaligned form field in `app/views/developers/_form.html.erb`

image


joemasilotti:
Thank you!

PR #397, opened 2022-05-01
Added website and contact_role fields to businesses

scottharvey:
### Pull request checklist


scottharvey:
Screen Shot 2022-05-01 at 09 28 08
Screen Shot 2022-05-01 at 09 29 22

Here is my first pass at implementing this feature, here are some things I think we should add:

1. Client side URL validation in admin
2. Make it so the business website opens in a new tab
3. Maybe add a label to the show business page so it's clear what the contact role information is
4. System tests

joemasilotti:
Looking good! Do you have the appetite to add those 4 items? If so, go for it. If not, let me know and we can hand it off.

test/system/business_profile_test.rb
```diff
@@ -0,0 +1,14 @@
+
+require "application_system_test_case"
+
+class BusinessProfileTest < ApplicationSystemTestCase
```
scottharvey, line 4: @joemasilotti When I run this system test for some reason it's running on port 3000 so it's connecting to my dev server instead of the test server.

Any thoughts on why that might be and how to fix it?

scottharvey: This is how I'm running the test

`rails test test/system/business_profile_test.rb HEADFUL=1`

joemasilotti: Hmm, I'm not sure. They are working fine for me.

But honestly, I don't think we need a system test for this. There's no custom JavaScript on this page so it feels like overkill. Adding to the existing integration test should be good enough.

Gemfile
```diff
@@ -64,3 +64,4 @@ gem "sitemap_generator", "~> 6.1"
 gem "stripe", ">= 2.8", "< 6.0"
 gem "timezone_finder", "~> 1.5.7"
 gem "tzinfo-data", platforms: [:mingw, :mswin, :x64_mingw, :jruby]
+gem "validate_url"
```
joemasilotti, line 67: I'm not sold on adding this dependency here since we aren't doing it for the developer's website. What's your reasoning for needing to validate this?

app/models/business.rb
```diff
@@ -12,6 +12,7 @@ class Business < ApplicationRecord
   validates :company, presence: true
   validates :bio, presence: true
   validates :developer_notifications, inclusion: {in: developer_notifications.keys}
+  validates :website_url, url: true
```
joemasilotti, line 15: If we are using the gem I think we also need to allow blanks.

```suggestion
  validates :website_url, url: {allow_blank: true}
```

db/migrate/20220501021745_add_website_url_and_contact_role_to_businesses.rb
```diff
@@ -0,0 +1,6 @@
+class AddWebsiteUrlAndContactRoleToBusinesses < ActiveRecord::Migration[7.0]
+  def change
+    add_column :businesses, :website_url, :string, null: true
+    add_column :businesses, :contact_role, :string, null: true
```
joemasilotti, line 4:
```suggestion
    add_column :businesses, :website, :string, null: true
    add_column :businesses, :role, :string, null: true
```

And applied to the rest of the PR, too!

app/views/businesses/show.html.erb
```diff
@@ -17,6 +17,14 @@
             <div class="ml-4 lg:ml-0">
               <div class="text-base font-medium text-gray-900"><%= @business.company %></div>
               <div class="text-base font-medium text-gray-600"><%= @business.name %></div>
+              <% if @business.contact_role.present? %>
+                <div class="text-base font-medium text-gray-600">
+                  Contact Role: <%= @business.contact_role %>
+                </div>
+              <% end %>
+              <div class="text-base font-medium text-gray-600">
+                <%= render LinkableTextComponent.new(@business.website_url, url: @business.website_url, target: "_blank", class: "underline") %>
```
joemasilotti, line 26: GitHub is reporting a security issue here. Can you address that?

Gemfile
scottharvey: I was on the fence about it as well, I will remove it and just use a regex to validate the field

db/migrate/20220501021745_add_website_url_and_contact_role_to_businesses.rb
scottharvey: I was going to name the field just `website` but then I thought it was better to be more specific in case you wanted to add a `website_name` or `website_icon` in the future.

The column name `role` seemed to vague to me, If I see `business.role` I would expect it to return the role of the business. Naming it `contact_role` seems like a much clearer name in my opinion.

test/system/business_profile_test.rb
scottharvey: > Adding to the existing integration test should be good enough

Which test are you referring to?

Gemfile
joemasilotti: It looks like we can wrap [`Addressable::URI`](https://www.rubydoc.info/gems/addressable/Addressable/URI) in a validator and that would get us pretty far.

db/migrate/20220501021745_add_website_url_and_contact_role_to_businesses.rb
joemasilotti: > I was going to name the field just `website` but then I thought it was better to be more specific in case you wanted to add a `website_name` or `website_icon` in the future.

If/when we add website name or icon we can rename this column.

> The column name `role` seemed to vague to me, If I see `business.role` I would expect it to return the role of the business. Naming it `contact_role` seems like a much clearer name in my opinion.

I buy that! Then we should probably rename `name` to `contact_name`.

scottharvey: I have pushed a commit to rename business `name` to `contact_name`

https://github.com/joemasilotti/railsdevs.com/pull/397/commits/50e5ef99258db620fb71a1a41ebe3a10a39feb5a

test/system/business_profile_test.rb
joemasilotti: `test/integration/businesses_test.rb`

db/migrate/20220501021745_add_website_url_and_contact_role_to_businesses.rb
joemasilotti: I'd prefer to rename this in a separate PR. But if we are going to do it here then let's not alias the method name and write a migration to rename the column.

test/system/business_profile_test.rb
scottharvey: I have updated the existing integration tests and removed my system test.

https://github.com/joemasilotti/railsdevs.com/pull/397/commits/69fa0632cc9c58f7c7f3fb36015e53ed7c11590b

db/migrate/20220501021745_add_website_url_and_contact_role_to_businesses.rb
scottharvey: I updated the migration to rename the database field.

https://github.com/joemasilotti/railsdevs.com/pull/397/files#diff-5b5a85582ac4960aa6ae66f16491a3fb6878af9fea738f5dcd6e65c89bdf5182R5

app/models/business.rb
```diff
@@ -8,13 +8,16 @@ class Business < ApplicationRecord
...
-  validates :name, presence: true
+  validates :contact_name, presence: true
   validates :company, presence: true
   validates :bio, presence: true
   validates :developer_notifications, inclusion: {in: developer_notifications.keys}
+  validates :website, format: URI::DEFAULT_PARSER.make_regexp(%w[http https]), allow_blank: true
 
   after_create_commit :send_admin_notification, :send_welcome_email
 
+  alias_attribute :name, :contact_name
```
scottharvey, line 19: @joemasilotti The reason for the `alias_attribute` is so we can still call `.name` on an object without know if it's a Business or a Developer eg:

`conversation.other_recipient(current_user).name`

joemasilotti: Got it, thanks!

joemasilotti:
Looking good! I have a few tweaks I want to make to the design that I'm hoping to get to soon.

I'm thinking of something like this.

image

scottharvey:
> I'm thinking of something like this.

That looks nicer 😄 

Let me know if you want me to do anything else for this PR 😉

Issue #393, opened 2022-04-30
Notify admin if developer changes status to indicate being hired


joemasilotti:
This is the first check in trying to track when a developer gets hired via the platform.

A PR addressing this issue should:

- [ ] Send a new admin email/notification when a developer's status changes to indicate they may have been hired

Any of the following changes should trigger the notification:

* A developer's search status changes from actively looking or open to not interested or invisible
* An admin marking a developer as invisible (via the Invisibilize button) should _not_ trigger the notification
* A developer removing full-time employment from their selected role types

benngarcia:
happy to jump in on this

joemasilotti:
Sure thing, all yours @benngarcia!

joemasilotti:
Closed via #401.

PR #401, opened 2022-05-01
Add potential hire notifications

benngarcia:
Addresses #393 - Adds notifications.

Some requested considerations were admin spam (new accounts clicking back and forth, with a higher profile update volume, despite likely not being potentially hired). Multiple triggers being hit, etc. Some things DB side could be different, but I'd rather leave that to a larger overhaul (if ever desired) than incremental improvements. Hack the rest in, lol.

One mild concern is I introduce a poorly performing developer method, `notifications_as_subject`, which needs to do a full sequential scan on with notifications table while parsing heavy text + performing postgres functions. Runs in background only at the moment, so it's fine to ship as is imo. 

ALSO: I google translated for translations, sorry. 

### Pull request checklist

- [:white_check_mark: ] My code contains tests covering the code I modified
- [:white_check_mark: ] I linted and tested the project with `bin/check`
- [:white_check_mark: ] I added significant changes and product updates to the [changelog](CHANGELOG.md)


config/locales/es.yml
```diff
@@ -234,6 +234,7 @@ es:
     new_conversation: "%{name} (%{company}) empezó una conversación con %{developer}."
     new_developer_profile: "%{developer} agregó su perfil de desarrollador."
     new_message: "%{sender} te envió un mensaje."
+    potential_hire: "%{developer} puede haber sido contratado."
```
joemasilotti, line 237: Please remove these - we can run a Google Translate sweep of everything later.

config/locales/pt-BR.yml
```diff
@@ -250,6 +250,7 @@ pt-BR:
     new_conversation: "%{name} (%{company}) iniciou uma conversa com %{developer}."
     new_developer_profile: "%{developer} adicionou seu perfil de desenvolvedor."
     new_message: "%{sender} lhe enviou uma mensagem."
+    potential_hire: "%{developer} pode ter sido contratado."
```
joemasilotti, line 253: Please remove these - we can run a Google Translate sweep of everything later.

app/notifications/potential_hire_notification.rb
```diff
@@ -0,0 +1,27 @@
+class PotentialHireNotification < Noticed::Base
+  deliver_by :database, if: :still_needed?
+  deliver_by :email, mailer: "AdminMailer", method: :potential_hire, if: :still_needed?
+
+  param :developer
+
+  def title
+    t "notifications.potential_hire",
+      developer: developer.name
```
joemasilotti, line 9:
```suggestion
    t("notifications.potential_hire", developer: developer.name)
```

app/notifications/potential_hire_notification.rb
```diff
@@ -0,0 +1,27 @@
...
+    developer_path(developer)
+  end
+
+  def still_needed?
+    return false if developer.notifications_as_subject
+      .where(created_at: 7.days.ago..10.seconds.ago,
+        type: "PotentialHireNotification").exists?
+
+    !developer.new_developer_account?
+  end
```
joemasilotti, line 26: Can you dive into why this is needed? It feels like a pre-optimization.

app/views/admin_mailer/potential_hire.html.erb
```diff
@@ -0,0 +1,4 @@
+<p><b><%= link_to @developer.name, "mailto:#{@developer.user.email}" %></b> may have been hired!</p> 
+
+
+<%= link_to "developer profile", notification_url(@notification) %>!</p>
```
joemasilotti, line 4:
```suggestion
<p><%= link_to "View their profile", notification_url(@notification) %></p>
```

test/mailers/previews/admin_mailer_preview.rb
```diff
@@ -13,4 +13,9 @@ def new_conversation
     notification = Notification.where(type: NewConversationNotification.to_s).first
     AdminMailer.with(record: notification, recipient: User.first).new_conversation
   end
+
+  def potential_hire
+    notification = Notification.where(type: PotentialHireNotification.to_s).first
```
joemasilotti, line 18: Is one of these created when running the database seeds? If not, can you update an existing record to make sure one of these notifications is created?

test/models/developer_test.rb
```diff
@@ -91,6 +91,41 @@ class DeveloperTest < ActiveSupport::TestCase
...
+    assert_difference "Notification.count", 0 do
+      developer.update!(search_status: :actively_looking)
+      developer.update!(search_status: :not_interested)
+    end
+
+    assert_equal Notification.last.type, PotentialHireNotification.name
+  end
+
+  test "correctly identifies a new developer account" do
+    assert_equal @developer.new_developer_account?, true
```
joemasilotti, line 126:
```suggestion
    assert @developer.new_developer_account?
```

test/models/developer_test.rb
```diff
@@ -91,6 +91,41 @@ class DeveloperTest < ActiveSupport::TestCase
...
+      developer.update!(search_status: :not_interested, created_at: (Developer::NEW_ACCOUNT_THRESHOLD + 1.day).ago)
+    end
+
+    assert_equal Notification.last.type, PotentialHireNotification.name
+    assert_equal Notification.last.recipient, users(:admin)
+  end
+
+  test "admins do not get alerted to new accounts changing search status" do
+    developer = Developer.create!(developer_attributes.merge(search_status: :actively_looking))
+    assert_difference "Notification.count", 0 do
```
joemasilotti, line 106:
```suggestion
    assert_no_difference "Notification.count"
```

test/models/role_type_test.rb
```diff
@@ -0,0 +1,14 @@
...
+
+  test "after update callback" do
+    developer = Developer.create!(developer_attributes.merge(created_at: 30.days.ago))
+    role_type = RoleType.create!(full_time_employment: true, developer:)
+
+    assert_difference "Notification.count", 1 do
+      role_type.update!(full_time_employment: false)
+    end
+  end
+end
```
joemasilotti, line 14:
```suggestion
end
```

joemasilotti: (fixes the line ending)

test/models/role_type_test.rb
```diff
@@ -0,0 +1,14 @@
+require "test_helper"
+
+class RoleTypeTest < ActiveSupport::TestCase
+  include DevelopersHelper
+
+  test "after update callback" do
```
joemasilotti, line 6:
```suggestion
  test "notifies admins when changes indicate a potential hire" do
```

test/models/role_type_test.rb
```diff
@@ -0,0 +1,14 @@
...
+    developer = Developer.create!(developer_attributes.merge(created_at: 30.days.ago))
+    role_type = RoleType.create!(full_time_employment: true, developer:)
```
joemasilotti, line 8: Any reason to not use an existing fixture and update these two fields?

joemasilotti (changes requested):
This is looking great! Thanks for kicking this off and knocking it out so quickly. I made a bunch of small change requests and posted a few questions.

Outside of those, my big question is if we really need the throttling based on when the change was made. I'd rather get flooded with notifications now and have to optimize later then pre-optimize now and have to maintain this code forever.

Perhaps all we need to get this launched is to ensure that if a developer triggers the notification via changing `RoleType` _and_ their search status that it only sends admins one notification.

Curious to hear what you think!

app/notifications/potential_hire_notification.rb
joemasilotti: Oh, and for future reference, I think we can use the built in querying support from noticed to find these.


https://github.com/excid3/noticed/discussions/220#discussioncomment-2501568

benngarcia:
> This is looking great! Thanks for kicking this off and knocking it out so quickly. I made a bunch of small change requests and posted a few questions.
> 
> Outside of those, my big question is if we really need the throttling based on when the change was made. I'd rather get flooded with notifications now and have to optimize later then pre-optimize now and have to maintain this code forever.
> 
> Perhaps all we need to get this launched is to ensure that if a developer triggers the notification via changing `RoleType` _and_ their search status that it only sends admins one notification.
> 
> Curious to hear what you think!

Oh, totally. I had trouble context switching from work, where I'm currently working on similar funnel optimizations, including admin notifications. I'm used to implementing a throttling mechanism, typically using our own `NotificationRecord` model, which also tracks all email sends. You're absolutely right that this is an unnecessary pre-optimization for this PR.

In terms of changing logical condition operator from OR to AND, I'm trying to remember how the after_update_commit will work when looking at an associated object that may be wrapped in the same transaction, or it maybe isn't wrapped in a transaction. In lieu of that complexity, I've taken a look at the original ask. I think it would be better to only look at the developers `search_status` and ignore changes to `role_type` for now. 

My reasoning is that for majority of use-cases we'd like to be notified of - users who remove their full_time_employment roll type option will be a majority inclusive sub-set of users who change their search status. I think we can reduce complexity, remove some red herrings and, in my mind, sacrificing very little 'capture' by only looking at developer `search_status` for now. 

Thoughts?

app/notifications/potential_hire_notification.rb
benngarcia: Addressed in comments below, I agree :)

joemasilotti:
> I think it would be better to only look at the developers search_status and ignore changes to role_type for now.

Works for me! Thanks for thinking that through. Let's do it.

benngarcia:
> > I think it would be better to only look at the developers search_status and ignore changes to role_type for now.
> 
> Works for me! Thanks for thinking that through. Let's do it.

Sweet, requested changes were made.

app/models/developer.rb
```diff
@@ -90,8 +95,23 @@ def feature!
...
   end
 
+  def new_developer_account?
+    created_at > NEW_ACCOUNT_THRESHOLD.ago
+  end
+
   private
 
+  def changes_indicate_potential_hire?
+    return false unless saved_change_to_search_status? && !new_developer_account?
```
joemasilotti, line 105: Let's drop the second clause. Since we are firing on update, and not on create, I'm fine with getting an extra notification if someone changes their status after signing up.

app/views/admin_mailer/potential_hire.html.erb
```diff
@@ -0,0 +1,4 @@
+<p><b><%= link_to @developer.name, "mailto:#{@developer.user.email}" %></b> may have been hired!</p> 
+
+
+<%= link_to "View their profile", notification_url(@notification) %>!</p>
```
joemasilotti, line 4:
```suggestion
<p><%= link_to "View their profile", notification_url(@notification) %></p>
```

db/seeds/developers.rb
```diff
@@ -54,3 +54,14 @@
   twitter: Faker::Internet.username,
   linkedin: Faker::Internet.username
 })
+
+# No Longer Looking
+developer_not_interested = SeedsHelper.create_developer!("hired", {
```
joemasilotti, line 59:
```suggestion
# Potential hire
developer = SeedsHelper.create_developer!("hired", {
```

joemasilotti (changes requested):
A few tiny tweaks and we are good to go!

benngarcia:
Latest requested changes made :+1:

joemasilotti:
Alright! I had to fix a broken test but other than that this is great.

I also tweaked some things that were personal preference. I'm going to follow up with another PR that addresses the TODO after I merge in #404.

PR #410, opened 2022-05-03
Accounts vs. users

joemasilotti:
@jayeclark, this is the most progress I've made on a PR for integrating accounts vs. users.

Some design goals:

1. `current_account` feels good enough for now.
2. I don't care about downtime if it makes the PR more complicated. Said another way, we can put the app in maintenance mode, deploy, then run all the migrations.
3. Account switching is required to launch.
4. Account creation _via creating a developer or business profile_ is the only way to create an account.
5. I don't care about inviting additional users to accounts just yet.

### DIFF
--- a/app/components/developer_profile_action_component.html.erb
+++ b/app/components/developer_profile_action_component.html.erb
@@ -1,6 +1,6 @@
 <div class="mt-4 flex space-x-3 md:mt-0">
   <% if update_profile? %>
-    <%= link_to edit_developer_path(user.developer), class: "inline-flex items-center px-4 py-2 border border-transparent shadow-sm text-sm font-medium rounded-md text-white bg-gray-600 hover:bg-gray-700 focus:outline-none focus:ring-2 focus:ring-offset-2 focus:ring-gray-500" do %>
+    <%= link_to edit_developer_path(account.developer), class: "inline-flex items-center px-4 py-2 border border-transparent shadow-sm text-sm font-medium rounded-md text-white bg-gray-600 hover:bg-gray-700 focus:outline-none focus:ring-2 focus:ring-offset-2 focus:ring-gray-500" do %>
       <%= inline_svg_tag "icons/solid/pencil.svg", class: "-ml-1 mr-2 h-5 w-5" %>
       <%= t(".edit_profile") %>
     <% end %>
--- a/app/components/developer_profile_action_component.rb
+++ b/app/components/developer_profile_action_component.rb
@@ -1,11 +1,11 @@
 class DeveloperProfileActionComponent < ApplicationComponent
-  attr_reader :user
+  attr_reader :account
 
-  def initialize(user)
-    @user = user
+  def initialize(account)
+    @account = account
   end
 
   def update_profile?
-    user&.developer
+    account&.developer
   end
 end
--- a/app/components/new_developer_fields_component.rb
+++ b/app/components/new_developer_fields_component.rb
@@ -1,10 +1,10 @@
 class NewDeveloperFieldsComponent < ApplicationComponent
   include ComponentWithIcon
 
-  attr_reader :user
+  attr_reader :account
 
-  def initialize(user, enabled: true)
-    @user = user
+  def initialize(account, enabled: true)
+    @account = account
     @enabled = enabled
   end
 
@@ -19,10 +19,10 @@ class NewDeveloperFieldsComponent < ApplicationComponent
   end
 
   def editing?
-    user&.developer&.persisted?
+    account&.developer&.persisted?
   end
 
   def missing_fields?
-    user.developer.missing_fields?
+    account.developer.missing_fields?
   end
 end
--- a/app/components/user_menu/base_component.rb
+++ b/app/components/user_menu/base_component.rb
@@ -1,14 +1,15 @@
 module UserMenu
   class BaseComponent < ApplicationComponent
-    attr_reader :user
+    attr_reader :user, :account
 
-    def initialize(user)
+    def initialize(user, account)
       @user = user
+      @account = account
     end
 
     def component
       if user.present?
-        UserMenu::SignedInComponent.new(user)
+        UserMenu::SignedInComponent.new(user, account)
       else
         UserMenu::SignedOutComponent.new
       end
--- a/app/components/user_menu/signed_in_component.html.erb
+++ b/app/components/user_menu/signed_in_component.html.erb
@@ -15,7 +15,7 @@
     <div>
       <button type="button" data-action="toggle#toggle" class="bg-gray-800 flex text-sm rounded-full focus:outline-none focus:ring-2 focus:ring-offset-2 focus:ring-offset-gray-800 focus:ring-white" id="user-menu-button" aria-expanded="false" aria-haspopup="true">
         <span class="sr-only"><%= t(".open_user_menu") %></span>
-        <%= render AvatarComponent.new(avatarable: user.business || user.developer, classes: "h-8 w-8") %>
+        <%= render AvatarComponent.new(avatarable: account.business || account.developer, classes: "h-8 w-8") %>
       </button>
     </div>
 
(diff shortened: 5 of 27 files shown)


jayeclark:
Hey @joemasilotti, let's make sure I understand the overall current vs eventual desired model before I really dive in because I'm not sure I'm understanding the terminology correctly (and my original intuition on approach is a bit different from/simpler than this PR):

| Current | Future |
|---------|--------|
| A **user** is an individual real person who accesses RailsDevs to try to get a job (developer), try to find talent (business), or both | No change |
| A user can own only one developer profile. | No change 
| A user can own only one business profile. | No change |
| A user associated with a business profile has different access to recruit developers on the board based on the subscription plan of the business | No change |
| Business profiles exist independently from the users who own them | No change |
| A subscription (and the evaluation of what the business has access to based its subscription) is tied to the business profile, not the user | No change |
| A user cannot be associated with more than one business profile | A user _can_ be associated with more than one business profile, as determined by the business profile owner.  The user may have different permissions with that business based on their role (ie an admin role where a user can invite and manage other users to the business vs a standard role that can contact devs on behalf of the business |
| A user cannot transfer ownership of their business profile | A user _can_ transfer ownership of their business profile (Use Case: I am a hiring manager at Company A and am leaving the company. I also have an active Developer profile. I need to transfer my ownership of Company A to another user so the history & contacts are preserved for them and so that I'm still able to use my own Developer profile.) |
| An existing user cannot become associated with an existing business | An existing user _can_ become associated with an existing business via a role defined by the business owner. (Use case: Company A hires Developer B who has their profile posted on RailsDevs. Developer B then becomes a standard member of Company A and uses RailsDev to recruit developers to a new team they are building at Company A.) |
| If a user deletes themselves, the associated developer and business profiles on which they are a superadmin are deleted as well | If a user deletes themselves, the associated developer and business profiles on which they are a superadmin are deleted as well. However, there is no change to the other business profiles they are associated with other than that the records granting them roles with those businesses are deleted. |

joemasilotti:
This sounds exactly like what I'm thinking! I don't think we need most of the second half in the first pass, but eventually and ideally all of this will become true.

To start, a user can only create their business and developer profile (that they are the owner of) and cannot transfer nor invite other users.

joemasilotti:
@jayeclark, I paired on this a bit with a Test Double employee at RailsConf this week. I'm not sure if it will help, but here's what we sketched out.

```ruby
class User < ApplicationRecord
  has_one :business, dependent: :destroy

  has_many :business_users
  has_many :businesses, through: :business_users
end

class BusinessUsers < ApplicationRecord
  belong_to :user
  belong_to :business

  enum :role, %i[owner member], default: :member
end

# Render a link to this in the user dropdown.
class ActiveBusinessessController < ApplicationController
  def create
    set_current_business(current_user.businesses.find(params[:id]))
  end
end

class ApplicationController
  include CurrentBusiness
end

module CurrentBusiness
  def current_business
    return @current_business if defined?(@current_business)
    @current_business = business || fallback_business
  end
  helper_method :current_business

  def set_current_business(business)
    session[:business_id] = business
  end

  private

  def business
    current_user&.businesses&.find_by(id: session[:business_id])
  end

  def fallback_business
    current_user&.businesses&.first
  end
end
```

jayeclark:
@joemasilotti Yeah, this is almost exactly the structure I had in mind! 

I was thinking more along the lines of BusinessUserRoles instead of BusinessUsers just for clarity on what the table is actually doing (matching existing users to existing businesses via the definition of a role.) I was also thinking of the enumerated roles as `%i[admin member]` instead of `%i[owner member]`  - there is already an ownership relationship through `has_one :business`, using admin would avoid having to write controls to make sure more than one owner role doesn’t get created, and it’s very likely a business owner would want to give admin privileges to one or more users. (I ran a small business for 5 years and used all kinds of recruiting apps & job boards to hire staff, and there was never a time when I didn’t have at least one other employee as an admin on those accounts.)

But otherwise this is almost exactly where I was headed, so, happy to continue on that path! Let me know your thoughts on those two points.

joemasilotti:
Love it, that's awesome to hear!

`BusinessUserRole` sounds great to me as does the names of the enum. We will most likely want multiple admins for each business, so good call there.

I think my only remaining question with this modeling is how we can tell if the user is signed in to their developer account (vs. a business account). I want to figure that out before we move ahead with this approach. Do you have any ideas on that?

joemasilotti:
Here's an idea, but it feels a little heavy handed considering there either is a developer account or there isn't. But maybe I'm overthinking it? Excited to hear what you think!

```ruby
module CurrentAccount
  def current_account
    if current_business.present?
      current_business
    elsif current_developer.present?
      current_developer
    end
  end
  helper_method :current_account

  # For views and display logic.
  def business_account?
    current_business.present?
  end
  helper_method :business_account?

  def developer_account?
    current_developer.present?
  end
  helper_method :developer_account?

  def current_business
    return @current_business if defined?(@current_business)
    @current_business = business || fallback_business
  end
  helper_method :current_business

  def set_current_business(business)
    session[:business_id] = business
    session[:developer_id] = nil
  end

  def current_developer
    return @current_developer if defined?(@current_developer)
    @current_developer = developer || fallback_developer
  end
  helper_method :current_developer

  def set_current_developer(developer)
    session[:developer_id] = developer
    session[:business_id] = nil
  end
end
```

In `#current_account` I'm checking for business first. This will follow our product goals of "if you have a business you default to your business" when we first deploy. Falling back to the developer, if present.

jayeclark:
I’ll take a look this weekend. I could have sworn that back in December users could switch freely among roles (ie you signed in to your account and then could choose to act in your dev role or your business role) but I see from recently posted issues that that seems to have changed or I’m mis-remembering. (So far I’ve just been looking at the code since I’ve been thinking mainly about the models & controllers - haven’t poked around too much in the UI yet.)

I’d think it would be best if it worked like Slack or Discord… when you log in, you’re ‘in’ whatever your last role/workspace/server was - and then you can switch freely among them as you need. There could easily be a situation where someone has a developer profile, a business they own, and a business they’re a member of. So there would be two businesses … maybe that gets covered in `current_business`? Even then, it feels like if I most recently was acting in my Developer role, when I log in again that’s where I should be, unless I’m logging in in response to a notification from a different workspace. 

I’d also expect that when I look at my notifications they’d be organized by entity (developer + businesses) and clicking on an entity to see the unread notifications also switches the role I’m acting in.

Another model that might apply is the GitHub model, where every time you have to specifically choose which repo (workspace) you’re operating in, but that model doesn’t seem as applicable.

I’ll check my old corporate Indeed account too - I recall you could have a personal account and also be associated with several different businesses as an owner, admin, or member. There may be some useful UI features that can be gleaned from that.

joemasilotti:
> I could have sworn that back in December users could switch freely among roles...

Maybe you're thinking of `/role/new` where it asks what kind of account you want to sign up for?

image

There currently isn't a way to switch accounts or context. The only thing built in is listing _both_ the developer and business profile if you have them in the user dropdown. But those only link to your profile – they don't switch how the app perceives you as a user.

image

> ...when you log in, you’re ‘in’ whatever your last role/workspace/server was...

I agree. I think if we swap out `session` for `cookies` in my example code that behavior should work. I don't want to persist it to the database. I want users to be able to open two browsers signed in as the same user but accessing different contexts/accounts.

> I’d also expect that when I look at my notifications they’d be organized by entity (developer + businesses) and clicking on an entity to see the unread notifications also switches the role I’m acting in.

Oh, yeah, forgot about notifications. That approach sounds great to me! So the bell would always show _all_ of your notifications, but grouped by context/account.

That works really well for when you open a notification from an email. As long as you are signed in to the _user_ account with access to the developer/business you are good to go.

joemasilotti:
@jayeclark, I'm going to close this in favor of your WIP PR #490.

jayeclark:
@joemasilotti Sounds good - apologies that I’ve been taking so long to get to it. I’ve been in a machine learning rabbit hole this month trying to finish up a professional cert in data science.   Once I can actually focus on it, I think it should go pretty quickly.

joemasilotti:
No worries at all. It's been fun to see your ML progress on Twitter!

Issue #419, opened 2022-05-09
Add admin search


joemasilotti:
I often find myself trying to find a developer or business in an admin context.

## A PR addressing this issue should: 

- [ ] Create a new page, `/admin/developers`
    - [ ] List all developers, paginated, in reverse `created_at` order
    - [ ] Display their name, only, in a table like `/admin/conversations`
    - [ ] Add a search field to filter developers by name and email address
- [ ] Create a new page, `/admin/businesses`
    - [ ] List all businesses, paginated, in reverse `created_at` order
    - [ ] Display their company and contact name, only, in a table like `/admin/conversations`
    - [ ] Add a search field to filter businesses by contact name, company, and email address

Search should be partial matching like we are already doing for `Developer` with `pg_search_scope`.

## Stretch goals (not need for this PR and could be in a follow up one):

- [ ] Add stats to `/admin/developers`
    - [ ] Number of developers
    - [ ] Number of actively looking + open developers
- [ ] Add stats to `/admin/businesses`
    - [ ] Number of businesses
    - [ ] Number of actively subscribed businesses

esmale:
Question regarding the data displayed vs the search criteria:
Developers
 * Search: name & email
 * Display: name

Businesses
 * Search: Contact Name, Company, Email address
 * Display: Company & Contact Name

Primary Question: It feels strange to have the search and display criteria not in sync with each other.  If you execute a search that matches a developer's email address, but that email address isn't displayed, you can't immediately see on the search results page what matched against your query.  So I want to verify that you don't want to see the developer email address in the table?

Secondary Question: I'm not seeing an email address field on the Business model itself.  Is the email address mentioned supposed to be the address of the User to which the Business record belongs?  (And if so, the same question as above also applies: if we're searching against it, should it also be displayed on the /admin/businesses page?)

joemasilotti:
> Primary Question: It feels strange to have the search and display criteria not in sync with each other. If you execute a search that matches a developer's email address, but that email address isn't displayed, you can't immediately see on the search results page what matched against your query. So I want to verify that you don't want to see the developer email address in the table?

Great point. Let's add email address to the table as well.

> Secondary Question: I'm not seeing an email address field on the Business model itself. Is the email address mentioned supposed to be the address of the User to which the Business record belongs? (And if so, the same question as above also applies: if we're searching against it, should it also be displayed on the /admin/businesses page?)

[...]

PR #430, opened 2022-05-19
Allow admin user to search through all developers and businesses in the app

esmale:
Implementing suggested enhancements from the [Add admin search](https://github.com/joemasilotti/railsdevs.com/issues/419) issue.

### Pull request checklist

- [ ✅ ] My code contains tests covering the code I modified
- [ ✅ ] I linted and tested the project with `bin/check`
- [ ✅ ] I added significant changes and product updates to the [changelog](CHANGELOG.md)


config/locales/en.yml
```diff
@@ -60,6 +60,15 @@ en:
...
   admin:
+    businesses:
+      businesses:
+        company: Company
+        contact_name: Contact Name
+        email: Email
+      index:
+        search_query_label: Search Businesses
+      stats:
+        number_of_businesses: Number of Businesses
```
joemasilotti, line 71:
```suggestion
        businesses: Businesses
```

config/locales/en.yml
```diff
@@ -84,6 +93,15 @@ en:
...
         reply_rate: Reply rate
         started: Started
+    developers:
+      developers:
+        email: Email
+        name: Name
+      index:
+        search_query_label: Search developer profiles
+      stats:
+        actively_looking_and_open: Actively Looking + Open
```
joemasilotti, line 103:
```suggestion
        actively_looking_or_open: Actively looking or open
```

app/views/admin/businesses/_businesses.html.erb
```diff
@@ -0,0 +1,32 @@
...
+            <th scope="col" class="py-3.5 pl-4 pr-3 text-left text-sm font-semibold text-gray-900 sm:pl-6"><%= t(".company") %></th>
+            <th scope="col" class="px-3 py-3.5 text-left text-sm font-semibold text-gray-900"><%= t(".contact_name") %></th>
+            <th scope="col" class="px-3 py-3.5 text-left text-sm font-semibold text-gray-900"><%= t(".email") %></th>
+          </tr>
+          </thead>
+          <tbody class="divide-y divide-gray-200 bg-white">
+          <% query.records.each do |business| %>
+            <tr>
+              <td class="whitespace-nowrap px-3 py-4 text-sm text-gray-500">
+                <%= link_to business.company, business_path(business), class: "hover:underline" %>
```
joemasilotti, line 17: Let's link to the business's admin conversation page instead.

app/views/admin/businesses/_businesses.html.erb
```diff
@@ -0,0 +1,32 @@
...
+        <table class="min-w-full divide-y divide-gray-300">
+          <thead class="bg-gray-50">
+          <tr>
+            <th scope="col" class="py-3.5 pl-4 pr-3 text-left text-sm font-semibold text-gray-900 sm:pl-6"><%= t(".company") %></th>
+            <th scope="col" class="px-3 py-3.5 text-left text-sm font-semibold text-gray-900"><%= t(".contact_name") %></th>
+            <th scope="col" class="px-3 py-3.5 text-left text-sm font-semibold text-gray-900"><%= t(".email") %></th>
+          </tr>
+          </thead>
+          <tbody class="divide-y divide-gray-200 bg-white">
+          <% query.records.each do |business| %>
```
joemasilotti, line 14: I think this needs to be indented once more.

app/models/admin/businesses_context.rb
```diff
@@ -0,0 +1,18 @@
...
+  class BusinessesContext
+    attr_reader :title, :search_query
+
+    def initialize(title:, search_query:)
+      @title = title
+      @search_query = search_query
+    end
+
+    def query
+      @query ||= BusinessQuery.new({search_query:})
```
joemasilotti, line 11:
```suggestion
      @query ||= BusinessQuery.new(search_query:)
```

app/models/admin/developers_context.rb
```diff
@@ -0,0 +1,18 @@
...
+  class DevelopersContext
+    attr_reader :title, :search_query
+
+    def initialize(title:, search_query:)
+      @title = title
+      @search_query = search_query
+    end
+
+    def query
+      @query ||= DeveloperQuery.new({search_query:, include_not_interested: true, search_type: :name_and_email})
```
joemasilotti, line 11:
```suggestion
      @query ||= DeveloperQuery.new(search_query:, include_not_interested: true, search_type: :name_and_email)
```

app/queries/developer_query.rb
```diff
@@ -15,6 +15,7 @@ def initialize(options = {})
     @role_levels = options.delete(:role_levels)
     @include_not_interested = options.delete(:include_not_interested)
     @search_query = options.delete(:search_query)
+    @search_type = options.delete(:search_type)
```
joemasilotti, line 18: Let's move all of these changes to a new model instead of trying to fit them inside of this object.

app/views/admin/businesses/_businesses.html.erb
```diff
@@ -0,0 +1,32 @@
...
+  <div class="-my-2 -mx-4 overflow-x-auto sm:-mx-6 lg:-mx-8">
+    <div class="inline-block min-w-full py-2 align-middle md:px-6 lg:px-8">
+      <div class="overflow-hidden shadow ring-1 ring-black ring-opacity-5 md:rounded-lg">
+        <table class="min-w-full divide-y divide-gray-300">
+          <thead class="bg-gray-50">
+          <tr>
+            <th scope="col" class="py-3.5 pl-4 pr-3 text-left text-sm font-semibold text-gray-900 sm:pl-6"><%= t(".company") %></th>
+            <th scope="col" class="px-3 py-3.5 text-left text-sm font-semibold text-gray-900"><%= t(".contact_name") %></th>
+            <th scope="col" class="px-3 py-3.5 text-left text-sm font-semibold text-gray-900"><%= t(".email") %></th>
+          </tr>
```
joemasilotti, line 11: This should be indented once more.

app/views/admin/developers/_developers.html.erb
```diff
@@ -0,0 +1,28 @@
+<div class="mt-8 flex flex-col">
+  <div class="-my-2 -mx-4 overflow-x-auto sm:-mx-6 lg:-mx-8">
+    <div class="inline-block min-w-full py-2 align-middle md:px-6 lg:px-8">
+      <div class="overflow-hidden shadow ring-1 ring-black ring-opacity-5 md:rounded-lg">
+        <table class="min-w-full divide-y divide-gray-300">
+          <thead class="bg-gray-50">
+          <tr>
+            <th scope="col" class="py-3.5 pl-4 pr-3 text-left text-sm font-semibold text-gray-900 sm:pl-6"><%= t(".name") %></th>
+            <th scope="col" class="px-3 py-3.5 text-left text-sm font-semibold text-gray-900"><%= t(".email") %></th>
+          </tr>
```
joemasilotti, line 10: Indentation.

app/views/admin/developers/_developers.html.erb
```diff
@@ -0,0 +1,28 @@
...
+          <% query.records.each do |developer| %>
+            <tr>
+              <td class="whitespace-nowrap px-3 py-4 text-sm text-gray-500">
+                <%= link_to developer.name, developer_path(developer), class: "hover:underline" %>
+              </td>
+              <td class="whitespace-nowrap px-3 py-4 text-sm text-gray-500">
+                <%= developer.user.email %>
+              </td>
+            </tr>
+          <% end %>
```
joemasilotti, line 22: Indentation.

app/views/admin/developers/_developers.html.erb
```diff
@@ -0,0 +1,28 @@
...
+          <tr>
+            <th scope="col" class="py-3.5 pl-4 pr-3 text-left text-sm font-semibold text-gray-900 sm:pl-6"><%= t(".name") %></th>
+            <th scope="col" class="px-3 py-3.5 text-left text-sm font-semibold text-gray-900"><%= t(".email") %></th>
+          </tr>
+          </thead>
+          <tbody class="divide-y divide-gray-200 bg-white">
+          <% query.records.each do |developer| %>
+            <tr>
+              <td class="whitespace-nowrap px-3 py-4 text-sm text-gray-500">
+                <%= link_to developer.name, developer_path(developer), class: "hover:underline" %>
```
joemasilotti, line 16: Let's link to their conversations here, as well.

app/models/business.rb
```diff
@@ -15,5 +16,9 @@ class Business < ApplicationRecord
   validates :bio, presence: true
   validates :developer_notifications, inclusion: {in: developer_notifications.keys}
 
+  scope :newest_first, -> { order(created_at: :desc) }
+
+  pg_search_scope :filter_by_search_query, against: [:company, :contact_name], associated_against: {user: :email}, using: {tsearch: {prefix: true}}
```
joemasilotti, line 21: What does `prefix: true` do?

Perhaps related, if someone's email is `foo@bar.com` I'd like to be able to search for `bar.com` and see them in the results. I don't think that's currently possible.

app/views/admin/businesses/_stats.html.erb
```diff
@@ -0,0 +1,13 @@
+<dl class="grid grid-cols-1 gap-5 sm:grid-cols-2">
```
joemasilotti, line 1: Until we add a second stat let's make this a single column.

joemasilotti (changes requested):
Looking good! Everything is more or less working so far outside of an inline question about `prefix`. I also made some requests in the code for minor code-style changes.

### Queries

My big piece of feedback is moving the querying logic _out_ of `DeveloperQuery` and into a new object. I don't think we will ever expose name/email searching outside of an admin context. So let's create a new object like `Admin::DeveloperQuery` or `Developers::AdminQuery` maybe.

Once we do that we might be able refactor `Admin::BusinessesContext` and `Admin::DevelopersContext` to a single object since they will pass along the same parameters to the query.

### Form

Let's add a submit button to the search forms. "Search" to the right of the input field should be good!

### Copy

Also, can you please update all translations to be sentence case instead of title case? For example, change "Search Businesses" to "Search businesses".

app/models/business.rb
esmale: Yeah, `prefix: true` will allow partial searches from the beginning of a database string (`foo` will give you `foo@bar.com`), but won't allow for search results starting anywhere else in the string (`@bar` won't give you `foo@bar.com`).  

So, in order to allow for searches the way you're describing, I'm assuming that I need to create another virtual column similar to the `textsearchable_index_col` column that already exists, but base the value on name & email instead of the hero and bio.  Is that the right direction to go in?

joemasilotti: That sounds right to me!

Unless removing `prefix: true` gets us what we want? But it doesn't sound like it will.

esmale:
Screen Shot 2022-05-21 at 11 20 17 PM

Is this adequate for the Search button?  From a design perspective I think I'd want the button rendered closer to the search field, instead of all the way on the other side of the page, but I'm not yet as proficient with Tailwind as I'd like to be.

esmale:
So I'm looking at setting up the ts_vector to allow for more versatile searching of Developers by name and email.  The problem I'm running into is that the name is on the Developer model, and the email is on the User model.  According to [this](https://www.postgresql.org/docs/current/ddl-generated-columns.html) Postgresql documentation (also shown in the screenshot), I don't think it's possible to generate a ts_vector column with these values in separate tables. 

Screen Shot 2022-05-21 at 11 44 10 PM

So...the simplest solution I can think of is caching the User email on the Developer email as well. Is that the direction you'd like things to go?  Any other suggestions here?

joemasilotti:
> Is this adequate for the Search button?

I think we can change `justify-between` to `justify-start` on the parent element to get what we want.

> The problem I'm running into is that the name is on the Developer model, and the email is on the User model. 

Darn, I forgot about that. Good catch.

Can we cache the value in the database entirely? Or do we need some sort of callback on `Developer` to update the record in Ruby?

If we can do it all via the database then let's go for it. If we can't then I'm inclined to drop the email search feature (for now) and add it in a new PR.

app/views/admin/developers/_developers.html.erb
```diff
@@ -0,0 +1,24 @@
...
+          </tr>
+          </thead>
+          <tbody class="divide-y divide-gray-200 bg-white">
+          <% query.records.each do |developer| %>
+            <tr>
+              <td class="whitespace-nowrap px-3 py-4 text-sm text-gray-500">
+                <%= link_to developer.name, admin_developer_conversations_path(developer), class: "hover:underline" %>
+              </td>
+            </tr>
+          <% end %>
```
joemasilotti, line 18:
```suggestion
            <% query.records.each do |developer| %>
              <tr>
                <td class="whitespace-nowrap px-3 py-4 text-sm text-gray-500">
                  <%= link_to developer.name, admin_developer_conversations_path(developer), class: "hover:underline" %>
                </td>
              </tr>
            <% end %>
```

joemasilotti:
Hey again @esmale, do you still have appetite to get this over the finish line?

If not, not worries. Let me know and I can take it from here.

joemasilotti:
I'm going to close this for now. It's gone a little stale and I'm still not sure the best way to address the cross-table query.

Instead, a more basic user-focused search will be implemented for now. See #601.

PR #434, opened 2022-05-21
Service layer

### LINKED ISSUE #433: Refactor business logic to a service layer
I attended a talk at RailsConf that really got me thinking. [Your Service Layer Needn't be Fancy, It Just Needs to Exist](https://railsconf.org/schedule#session-1281) from David Copeland, author of [Sustainable Web Development with Ruby on Rails](https://sustainable-rails.com).

The gist was that all of your business logic is scattered across your app. If you need to know _what_ your app does you have to open models, controllers, jobs, and maybe even some views. Instead, he proposed moving all business-specific code to a service _layer_.

Note "layer" and not object. He doesn't care _how_ you do it (POROs were recommended) just that it has to exist. And it doesn't need to be complicated.

---

Based on that I'm inspired to try and consolidate some business logic of railsdevs into `services/`. I kind of already started this with #387 but it could be extended further. Here are some high-level guidelines for what I'm planning.

* Move non-database-related logic out of Active Record models
  * Exception: keep validations in the model
  * "Creating a user via the website is not the same adding a row to the database"
* Move business-related logic into a PORO under `services/`
  * `CreateDeveloper#create_developer(developer_params)` is better than `DeveloperService.create(params)` - explicit about exactly what each class and method does
* Service layer _could_ be moved to a Ruby command line app
  * Return a state and the model, let the controller switch on that and generate the path
  * Return an error message, let the controller shove it in the flash
* Roll our own authorization
  * Consolidate _logic_ into specific services, but Pundit has proved too limiting

joemasilotti:
This PR closes #433.

## Pull request checklist

## To-do

- [x] Move analytics tracking business logic to `app/services/`
- [x] Move subscription businesses logic to `app/services/`
- [x] Move `BusinessesController` business logic to `app/services/`
- [ ] Move `DevelopersController` business logic to `app/services/`
- [ ] Move `MessagesController` business logic to `app/services/`
- [x] Move `ColdMessagesController` business logic to `app/services/`
- [ ] Move `Availability` to `app/services/`
- [ ] Move `BusinessSubscriptionCheckout` to `app/services/`
- [ ] Move `Conversation` business logic to `app/services/`
- [ ] Move `Message` business logic to `app/services/`
- [ ] Move `Developer` business logic to `app/services/`


joemasilotti:
Closing in favor of keeping logic in domain models.

With one caveat: if there's a _concept_ that can be named I think it should be its own "thing". `Businesses::Permission` and `Businesses::Subscription` are two great examples that I pulled over to `main` in #449.

Ideas for future refactors:
* Move the `before_action` logic of `ColdMessageController` into `ColdMessage`
* Move `app/models/businesses/permission.rb` to `app/permissions/business.rb`

PR #441, opened 2022-05-23
Prompts for developers

jacobdaddario:
This a pull-request that adds the first stage of allowing for the creation of developer prompts on the profile. This is referenced in #412. It enables an admin to add, delete, and enable or disable prompts for response on the site. At this time, the code still needs a set of end-to-end regression tests.

### Pull request checklist

This PR is at this time a draft. I will ensure to handle each of these todo items before marking it for review.


jacobdaddario:
@joemasilotti This is in what I consider a minimal state for functionality of the feature now. I need to add tests for everything, but I've marked this for review because I'd like you to take a look at everything and okay it before I write a bunch more tests. This is a huge feature, so let me know if you need help getting through everything and understanding how it works.

joemasilotti:
This is looking like a great start! I haven't looked at the code yet but I have a few comments on the UX.

1. The prompts feel kind of thrown in my face on the developer profile. Maybe all that's missing is a title/heading, though?
2. I think I'd rather see the _responses_ to a prompt be paywalled - not the entire section. Then developers without any responses won't have a huge paywalled component on their profile.
3. I'm not sure how I feel about having the CTA on the developer `show` page vs. on the `edit` screen. What are your thoughts for keeping it there?
4. Disabling a prompt should hide the prompts for folks who have already filled that out, too.
5. Design-wise, I think the prompts should flow vertically, not in a grid. This encourages more to be written, as well.
6. Along those lines, some guidelines or description of what these prompts are and why you might want to use them is needed. Feel free to copy/paste some Tailwind UI design into there and I can tweak the copy.

How do you want to proceed next? Want me to dive into the code or do you want to take a pass at these UX items first?

jacobdaddario:
Hey! I can take a look at this after work. I think a few questions that you mentioned have already been taken care of. I can handle the bulk of it so long as you aren't in a rush to get it merged. I'm moving sometime next week so I'm a bit more limited after work than usual.

Feel free if you want to just run with it too. Still need tests around the response code.

Regarding the length of the prompts, are you thinking going for something longer? I've capped them to be pretty short, reason being that I figured this feature could be used in lots of other places. For instance, this could lead to a `blind_digest` model that is used to serve up a handful of prompts and responses with no name or image attached to them. I know that diversity is something that's important to you, and this can help with implicit bias. 

Additionally, I've already sketched out a likes feature that could go with that perfectly. We could serve up a digest of 5 little response cards on the home page, the company rep can like the cards that interest them, and then they see a feed of liked devs on another view. Then once you start collecting digests with likes, you can start doing a recommendation engine.

I know that's a big info dump, but I definitely had this idea after watching a video about matrix factorization a few months. I googled "Ruby sparse matrix factorization" and Andrew Kane had already written library that I wanted to try and use somewhere. 😆

jacobdaddario:
Alright, finally getting around to this. Going to handle a couple of things that you mentioned, but I still think there's a point of discussion to be had around a few. I'll document this discussion points after I've corrected the obvious stuff.

-[ ] The prompts feel kind of thrown in my face on the developer profile. Maybe all that's missing is a title/heading, though?
-[ ] I think I'd rather see the responses to a prompt be paywalled - not the entire section. Then developers without any responses won't have a huge paywalled component on their profile.
-[ ] I'm not sure how I feel about having the CTA on the developer show page vs. on the edit screen. What are your thoughts for keeping it there?
-[ ] Disabling a prompt should hide the prompts for folks who have already filled that out, too.
-[ ] Design-wise, I think the prompts should flow vertically, not in a grid. This encourages more to be written, as well.
-[ ] Along those lines, some guidelines or description of what these prompts are and why you might want to use them is needed. Feel free to copy/paste some Tailwind UI design into there and I can tweak the copy.

joemasilotti:
## I'm not sure how I feel about having the CTA on the developer show page vs. on the edit screen. What are your thoughts for keeping it there?

I love the idea of breaking down the developer profile into multiple pieces.

Tailwind UI, for example, has Submit/Save buttons one very section of the form. This could be a good starting point as this form continues to grow. You can come back and edit a small piece without having to scroll to the bottom. And to start, we could disable the "inline" submit buttons when creating a new profile.

Thinking ahead, having some sort of wizard or multi-part form submission probably makes sense. I'd like to hand hold developers (and businesses) through onboarding a bit and this could be a good excuse to kick that off.

But back to the point you made - yes. It does feel a little against the grain right now. What about adding a sidebar/menu to the left of the developer form that has Profile and Prompts in it? Then prompts could have its own index page and could use Hotwire (or not) to get everything in there?

This Tailwind UI example is a pretty good approximation of where I'd like to see the Developer form on railsdevs in the future.
form-layouts 05-two-column-cards-with-sidebar-xl

## Design-wise, I think the prompts should flow vertically, not in a grid. This encourages more to be written, as well.

OK, you sold me. :)

I think the super rounded corners and default height of the cards made it look too a bit too... aggressive? in my mind. Maybe we can take some inspiration form Tailwind UI's [Card Headings](https://tailwindui.com/components/application-ui/headings/card-headings) section to lighten up the borders, bolden the title, and let the card grow based on the content (while still making two side-by-side equal).

How does that sound?

joemasilotti:
All good! This is definitely a long term thing, so no sense in rushing it.

I _think_ nested attributes should work? As in, these are records that a developer instance owns? Something like this, maybe.

```ruby
class Developer < ApplicationRecord
  has_many :prompt_responses

  accepts_nested_attributes_for :prompt_responses
```

jacobdaddario:
For sure, that'd absolutely work. Just wanted to ensure that was the approach that you were looking for. It'll take some retooling on my end to get that compatible, since the Hotwire approach for nested models like that is a bit different. We'd be injecting forms into the page instead of actually creating/updating. Those actions will be removed.

I need to brush up on how nested validations works again too. Last time I got into that, I got down into the weeds a fair bit. My first ever "big" project for my student club involved some serious injection of nested models for a form generator, and that was a huge pain to do by hand with JS. Hotwire should make this much easier.

Last thing, I remember from that project that by convention, Rails has a weird `_delete` convention that only truly removes records after submitting the form. That seems like an awful modern design pattern, so I imagine we'd immediately want to remove the record, right? The only reason I ask is because that will differ from our behavior for editing and creating which will only save after submitting the form.

joemasilotti:
Oh, good call. You know what? What if we keep all the Hotwire stuff you've built but only allow existing developers to respond to prompts? So when you create your profile initially you don't see anything to do with prompts. But we could redirect them somewhere to answer their first question or two?

That also means we can have a dedicated page to answer prompts that's all Hotwire-powered.

jacobdaddario:
That sounds good to me. I like breaking up the flow, into separated submissions. You good with separating the the behavior of the responses from the behavior of the form?

I don’t mind retooling the interface. I’m more concerned about developing the right interface for you. I think nested attrs is the right solution for a single form.

joemasilotti:
Yeah let's split it then!

joemasilotti:
Hey @jacobdaddario, I've been thinking about this a bit over the past month and I'm not sure I'm ready to merge something like this in just yet.

I love the idea of enabling developers to share more information. And this feels like a great start. But I'd like to do a little more research before we bring in the maintenance and complexity of this feature.

I'm planning to start talking to developers (and businesses) on the site more in the coming month. Do you mind if I close this PR for now? We can always revisit when I have some more information from folks using the site.

jacobdaddario:
Yeah of course! I'm so sorry for letting this go stale 🤦🏻 That's on me for not following through on this. I've always tried to stay on top of things opened up on Railsdevs and dropped the ball on this anyways. Let me know what you come up with, I'd be happy to implement. I think there's a lot of chances to improve discoverability right now for the site.

joemasilotti:
No worries and no need to apologize!

I agree that now that there are 500+ (!) developers on the site discoverability is becoming an actual issue. I'll keep you posted and thanks for understanding.

joemasilotti:
Jacob, I'm going to close this for now. Feel free to reopen if you want to pick it back up!

I'll also keep you in mind for when the next batch of work comes in related to developer discoverability.

PR #451, opened 2022-05-28
Fix flash messages for bussiness and developer

### LINKED ISSUE #442: Flash messages don't appear when creating a business or developer profile
**Describe the bug**

When creating a developer or business profile the flash message doesn't appear.

**To Reproduce**

1. Create an account
2. Visit `/developers/new` or `/businesses/new` and fill in the required fields
3. Submit the form
4. Note that no flash message appears

**Expected behavior**

A flash message appears without needing to refresh the page.

franklinjosmell:
I was digging and I found that the flash messages don't show because both Developers and Businesses controllers make two redirects. So in the first redirect it's set the notice message and then redirect again in EventsController with the new values for flash event. So because it's made a second request the notice message is lost.

For fix the bug I added flash.now in EventsController to preserve the previos flash messages for the next request. Also added more assert tests.

Fix #442

### DIFF
--- a/app/controllers/analytics/events_controller.rb
+++ b/app/controllers/analytics/events_controller.rb
@@ -8,6 +8,7 @@ module Analytics
         flash[:event] = {goal: event.goal, value: event.value}
       end
 
+      flash.keep
       redirect_to event.url
     end
   end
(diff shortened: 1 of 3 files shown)


joemasilotti:
Oh! Excellent catch. Did you verify that the analytics event was tracked online once?

joemasilotti:
I checked myself - everything is working as expected.

Thanks for the contribution!

PR #452, opened 2022-05-29
Add stale developer notifications

### LINKED ISSUE #420: Send an email to a developer who hasn't updated their profile in 30 days
Some profiles on railsdevs are growing stale and might not accurately reflect their current search status or bio.

- [ ] Run a background job every day that:
    - [ ] Finds developers who's `updated_at` is ≥ 30 days old
    - [ ] Sends them an email asking if their profile is still up to date
- [ ] Only send to a developer once every 30 days - see notes below
- [ ] Send the email from `railsdevs <reminders@railsdevs.com>` set from `Rails.application.reminders_email`
- [ ] Set up Postmark to use the Broadcast Stream - see notes below
- [ ] Include an unsubscribe link at the bottom
    - [ ] Which links to `/developers/:id/notifications` with a checkbox for this email type - see notes below
    - [ ] Is checked by default
    - [ ] Doesn't send if unchecked
- [ ] Include the email subject/body below

**Subject**: Is your railsdevs profile up to date?

> Hey `#{developer.first_name}`,
>
> Joe, solo-founder of [railsdevs](https://railsdevs.com) here. It looks like your developer profile hasn't been updated in a while. Is everything still accurate?
>
> If not, you can [update your profile](#{edit_developer_path}). Or, if you've found a gig and are no longer looking let me know by replying to this email!
> 
> `<%= render "shared/signature" %>`

---

### Only send to a developer once every 30 days

One idea is to create a new noticed notification to trigger this email. Then we can query to see if that notification was created within the last 30 days before sending another email. But I'm open to ideas!

### Set up Postmark to use the Broadcast Stream

Add the following option to the `mail` option in the mailer to use the broadcast stream:

```ruby
stream: "broadcast"
```

### Which links to `/developers/:id/notifications` with a checkbox for this email type

This requires a new screen. Feel free to use an unstyled form if you don't want to design it yourself and I can take over when the logic is implemented.

If you _do_ want to design the form then it can look like the bottom of the business profile form for notifications.

Image

mrtnin:
Closes #420 - Send an email to a developer who hasn't updated their profile in 30 days

This PR adds feature for notifying developers of their stale profiles. It adds a rake task to find and loop over developers who needs notifying.I think I have handled most of the requirements put fourth in #420 except scheduling the background job and anything related to postmark. I've added my suggestions to locale, but they obviously might need some extra work.

Finding the stale developers is done in `app/queries/stale_developer_query.rb` by looking up all `Developer` records that has `updated_at` older than 30 days, but where their `user_id`'s are not present in a list of recent `Notification` records.
This seems for me to be a relatively easy approach compared to tracking last sent notification as a separate attribute. As discussed on discord I've added @joemasilotti suggestions to use Arel to make the query more readable, and I hope I was able to do that.

## Pull request checklist


joemasilotti:
Thanks for the contribution @mrtnin! PR looks great – I made some minor copy/design tweaks after seeing it in the browser.

A bigger change: I moved `/developers/:id/notifications` to `/developers/:id/edit`. A new screen for a single notification preference felt like overkill. Moving it to the profile screen also matches the business notification settings.

In the future we will probably want to move this to its own screen – but until then I'm going to run with this and keep it simple. And yes, I 100% recognize that _I_ initially asked for the new screen. So thanks for understanding my flip-flopping. :D

Issue #422, opened 2022-05-09
Send an email when a developer gets their first message with tips on how to respond


joemasilotti:
It's important for developers to respond to every message they receive, even if they aren't interested in the role. It creates good rapport, validates the connection, and can even lead to interesting but unrelated work down the road.

## A PR addressing this issue should:

- [ ] Send an automated email when a developer receives their first message
- [ ] Include the subject/body below
- [ ] Send the message from `Rails.configuration.support_email`

**Subject**: Tips on responding to your first message on railsdevs

> Hey, `#{developer.first_name}`,
>
> I'm Joe, the solo-founder of [railsdevs](https://railsdevs.com). You received your first message from a business today. Congratulations!
>
> Here are some tips to make the most of the conversation:
>
> 1. Respond to every message. Let a business know if you aren't interested.
> 2. Don't be afraid to ask for salary or budget early in the process.
> 3. Feel free to move the conversation to video chat to get to know each other.
>
> Finally, when you land your first gig let me know by responding to this email!
>
> Don't hesitate to reach out if you run into issues or need help on the site.
>
> `<%= render "shared/signature" %>`

joemasilotti:
Closed via #454.

PR #454, opened 2022-05-30
Add send email to developer when they get first message

seagalputra:
Hi, i would like to propose my solution regarding the issue #422. When user get first message, they will be notified by email with the solution how to respond the message. I'm still on going to write the testing for this feature, but i already test manually by sending message using business account.

## Pull request checklist


app/mailers/message_mailer.rb
```diff
@@ -12,4 +12,9 @@ def new_message
 
     mail(to: recipient.email, subject: "#{@sender} sent you a message on railsdevs")
   end
+
+  def first_message
+    @developer = params[:developer]
+    mail(from: Rails.configuration.support_email, to: @developer.user.email, subject: "Tips on responding to your first message on railsdevs")
```
joemasilotti, line 18:
```suggestion
    mail(
      from: Rails.configuration.support_email,
      to: @developer.user.email,
      subject: "Tips on responding to your first message on railsdevs"
    )
```

app/models/conversation.rb
```diff
@@ -31,6 +31,10 @@ def hiring_fee_eligible?
     developer_replied? && created_at <= 2.weeks.ago
   end
 
+  def self.first_message?(developer)
+    where(developer:).count == 1 && where(developer:).first.messages.count == 1
+  end
```
joemasilotti, line 36: Let's move this to `Message` and use the following query since it will only make a single database call.

```ruby
class Message < ApplicationRecord
  def self.first_message?(developer)
    joins(:conversation).where(conversation: {developer:}).one?
  end
end
```

joemasilotti (changes requested):
This is looking good! I left an architecture suggestion and a code style request.

Once those are addressed and you've added some tests this should be good to merge!

Tests needed:

1. `Message.first_message?` query is accurate
2. `Message#save` sends or doesn't send the email
3. A mailer preview for `MessageMailer#first_message`

Let me know if you need help with any of these.

seagalputra:
Thank you for your feedback @joemasilotti,

I already added several improvement based on your suggestions above and also several tests.

Regarding the testing point that you mention at no. 2, can you explain more clearly? Should I check the email will also send when I call `Message#save` or something else?

app/models/conversation.rb
seagalputra: Done, I moved those method to `Message` and modify the query

joemasilotti:
This is great, thanks for the contribution @seagalputra!

I moved the mailer to `DeveloperMailer` and made a few code style changes.

PR #455, opened 2022-05-30
Conversations Revamp part 1

kaspth:
@joemasilotti hey, here's my stab at revamping the conversations a bit as a starting point.

I think the main step we'd want to progress to is the conversation not knowing about the business and developer but instead just hanging the associations off the participating users. That turned out to be a bigger change than I had time for, so I ended up focusing on extracting domain logic that was spread out in various controllers into Conversation, and then showcasing extracting a helping domain object in `Conversation::Block`.

I also explored marking the User as a `delegated_type :usable, types: %i[ Business Developer ]`, but that also had a lot of ramifications.

Let me know what you think!

### DIFF
--- a/app/models/conversation/block.rb
+++ b/app/models/conversation/block.rb
@@ -1,0 +1,5 @@
+class Conversation::Block < ApplicationRecord
+  belongs_to :conversation
+  belongs_to :blocker, class_name: "User"
+  belongs_to :blockee, class_name: "User"
+end
--- a/app/controllers/blocks_controller.rb
+++ b/app/controllers/blocks_controller.rb
@@ -8,7 +8,7 @@ class BlocksController < ApplicationController
 
   def create
     authorize conversation
-    conversation.blocked_by(current_user)
+    conversation.block_by(current_user)
     redirect_to root_path, notice: t(".notice", other_recipient:)
   end
 
--- a/app/controllers/blocks_controller.rb
+++ b/app/controllers/blocks_controller.rb
@@ -8,17 +8,14 @@ class BlocksController < ApplicationController
 
   def create
     authorize conversation
-    conversation.blocked_by(current_user)
-    redirect_to root_path, notice: t(".notice", other_recipient:)
+    block = conversation.blocked_by(current_user)
+
+    redirect_to root_path, notice: t(".notice", other_recipient: block.blockee.name)
   end
 
   private
 
   def conversation
     @conversation ||= Conversation.find(params[:conversation_id])
-  end
-
-  def other_recipient
-    conversation.other_recipient(current_user).name
   end
 end
--- a/app/controllers/blocks_controller.rb
+++ b/app/controllers/blocks_controller.rb
@@ -8,14 +8,17 @@ class BlocksController < ApplicationController
 
   def create
     authorize conversation
-    block = conversation.blocked_by(current_user)
-
-    redirect_to root_path, notice: t(".notice", other_recipient: block.blockee.name)
+    conversation.blocked_by(current_user)
+    redirect_to root_path, notice: t(".notice", other_recipient:)
   end
 
   private
 
   def conversation
     @conversation ||= Conversation.find(params[:conversation_id])
+  end
+
+  def other_recipient
+    conversation.other_recipient(current_user).name
   end
 end
--- a/app/controllers/blocks_controller.rb
+++ b/app/controllers/blocks_controller.rb
@@ -8,7 +8,7 @@ class BlocksController < ApplicationController
 
   def create
     authorize conversation
-    conversation.touch(blocked_by_column)
+    conversation.blocked_by(current_user)
     redirect_to root_path, notice: t(".notice", other_recipient:)
   end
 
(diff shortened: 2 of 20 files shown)


joemasilotti:
Ohh, the "blocked" extraction is nice! And I _really_ love the reduced controller/policy logic now that `Conversation` has a bit more domain knowledge.

I'm going to make a few tweaks to the naming of methods (personal preference) before merging this in.

> I also explored marking the User as a `delegated_type :usable, types: %i[ Business Developer ]`, but that also had a lot of ramifications.

Any thoughts on where to start with that? And how that will help clean up the code?

kaspth:
> Ohh, the "blocked" extraction is nice! And I _really_ love the reduced controller/policy logic now that `Conversation` has a bit more domain knowledge.

nice, glad to hear that! 🙌

> I'm going to make a few tweaks to the naming of methods (personal preference) before merging this in.

Totally, I just went with what made most sense to me — hopefully the naming is easier now that it's encapsulated in the model.

> Any thoughts on where to start with that? And how that will help clean up the code?

I think there's a missing abstraction to lay on top of a business and developer, a common polymorphic type that can then reference either. Because a lot of code seems to have to accommodate and know about both versus a model that can aggregate and delegate to either as needed. The closest tooling that Rails has to that is `delegated_type` and the closest to that model right now is the User, but it doesn't have to be either. E.g. if a business [is more of a team](https://blog.bullettrain.co/teams-should-be-an-mvp-feature/) and can have multiple users on it.

app/models/conversation/block.rb
```diff
@@ -0,0 +1,5 @@
+class Conversation::Block < ApplicationRecord
```
joemasilotti, line 1: @kaspth, what's your rationale behind this over:

```ruby
module Conversations
  class Block

  # ...
```

I'm currently doing the `module` with a plural name so it doesn't conflict with the model name (e.g. `Businesses::Permission`) but really like this approach of not having to pluralize.

Is there a downside to doing this in one line?

jacobdaddario: Whoh... This is really interesting. I didn't know that you could use the exact name of a class as a namespace if its a single line declaration. Is that a Zweitwerk related feature? What causes that difference in behavior?

I guess Ruby doesn't complain because we don't explicitly declare `Conversation` as a module. What assumption does it make about the kind of constant that `Conversation` is? If it assumes it's a class, why can't it cause a load sequence error like doing a multi-line declaration?

joemasilotti: > I hadn't thought about the fact that this wouldn't work if you're using the model name.

That's what I was afraid of and why I avoided the `Developer` _module` and went for `Developers` in the past. I really don't want to have a class be the namespacing because it requires an additional layer of complexity to understand what's going on.

But I do like that you can include the module without having to namespace it another layer!

Ideally the entire app follows the same convention (which is currently pluralized namespaces) but I do like how this feels a bit better. I'll have to think on it a bit more because the trade-offs might not be worth it.

Also super curious as to what @jacobdaddario is asking.

kaspth: > Is that a Zweitwerk related feature? What causes that difference in behavior?

It's a standard Ruby feature since is `Class` is a subclass of `Module`. The difference in behavior is due to loading order.

> I guess Ruby doesn't complain because we don't explicitly declare `Conversation` as a module. What assumption does it make about the kind of constant that `Conversation` is? If it assumes it's a class, why can't it cause a load sequence error like doing a multi-line declaration?

`Conversation` has already been loaded at this point, so Ruby doesn't assume anything. It's a class but classes can be namespaces too.

kaspth: > That's what I was afraid of and why I avoided the `Developer` module

Note: it wouldn't be a module though, it would still be the `Developer` class, but classes can be namespaces too.

> I really don't want to have a class be the namespacing because it requires an additional layer of complexity to understand what's going on.

What's the extra layer of complexity? To me it's more complex that I have to define a separate module — which exposes the namespace complexity anyhow, and could lead to loading order issues (though that doesn't happen with Zeitwerk).

> Ideally the entire app follows the same convention (which is currently pluralized namespaces) but I do like how this feels a bit better. I'll have to think on it a bit more because the trade-offs might not be worth it.

Looks like RuboCop has a rule that could be enforced for this (and the whole app could be converted in one-go if need be). https://www.rubydoc.info/gems/rubocop/RuboCop/Cop/Style/ClassAndModuleChildren

jacobdaddario: Gotcha, so that makes more sense. It's inferring the one line assignment essentially as:
```
class Conversation
  class Block
  end
end
```
And the reason it has to be one-line is because of the load ordering. Sorry for the confusion around that. For some reason in my head the double semicolon always meant namespace, but obviously that's not true. It just means it's a constant holding another constant.

joemasilotti:
I'm going to close this PR for now. I've let it go stale and there's a lot of code I don't want to bother merging.

But if we decide to revamp conversations again via a `delegated_type` then we can start with this branch as a base.

PR #470, opened 2022-06-02
Developer search query is based on active sub

### LINKED ISSUE #460: Feature gate developer search to paying customers
Currently, developer search is only enabled for admins (and in development). I'd like to enable this for folks with an active business subscription in production.

### A PR addressing this issue should:

- [ ] Remove `:search_feature` from `Feature.rb`
- [ ] Show the enabled search field if the current user has an active business subscription
- [ ] Show a disabled search field otherwise with a lock next to it - similar to viewing private developer information
- [ ] Filter parameters in `DevelopersController` to ensure someone without an active subscription can't use the search feature

I realize the disabled search field is a little ambiguous. Feel free to push something undesigned and I can take over or we can discuss via GitHub!

troyizzle:
Remove the feature `:search_feature` and change the logic into being based on the current user having an active subscription. 

I used the same component used in showing the developers private information (`PaywalledComponent`) which with the current text doesn't make the most sense.

Looking forward to any feedback!

Screen Shot 2022-06-02 at 5 34 40 PM
Screen Shot 2022-06-02 at 6 34 43 PM

Closes #460


joemasilotti (changes requested):
This is looking great! I made one change to work in the design/copy I want for the `PaywalledComponent`. I have two change requests before we merge this in.

First, let's move the business permission check from `DeveloperQuery` (and related tests) to a policy. We can use [Pundit's Strong Parameters](https://github.com/varvet/pundit#strong-parameters) to not permit the search query parameter if the user doesn't have an active business subscription. There's an example in `BusinessesController` and `BusinessPolicy#permitted_attributes`.

Once that's done, can you please resolve the merge conflicts with `main`?

joemasilotti:
Looks great - thanks for the contribution!

PR #484, opened 2022-06-10
Add empty state for `/developers`

### LINKED ISSUE #472: Empty state for `/developers`
With the new filters it is now easier to filter everyone out and not understand what's going on. The blank page doesn't help much.

- [ ] Add an empty state when `/developers` returns 0 results
- [ ] Add a custom empty state when `/developers` returns 0 results in dev (telling the user to seed their database via  `bin/rails db:seed`)

Design should be modeled after [Tailwind UI's Empty States](https://tailwindui.com/components/application-ui/feedback/empty-states) but without the button. If you need some HTML to get started let me know and we can figure something out.

bensheldon:
Connects to #472.

I used the existing `EmptyStateComponent` to structure the content. Is the language and icon appropriate?

Screen Shot 2022-06-10 at 3 23 04 PM

## Pull request checklist


app/views/developers/index.html.erb
```diff
@@ -41,15 +41,21 @@
...
     <!-- Paginated list of developers -->
     <div class="col-span-12 md:col-span-9">
-      <%= render Developers::CardComponent.with_collection(@query.featured_records, featured: true) %>
-
-      <%= render PagyPaginatorComponent.new(id: "developers", pagy: @query.pagy, url_array: [:developers], container_classes: "divide-y divide-gray-200", options: @query.filters) do |component| %>
-        <% component.loading_icon do %>
-          <div class="py-4">
-            <%= render LoadingComponent.new %>
-          </div>
+      <% if @query.featured_records.any? || @query.records.any? %>
```
joemasilotti, line 44:
```suggestion
      <% if @query.records.any? %>
```

I'd rather not _only_ show a featured developer!

app/views/developers/index.html.erb
```diff
@@ -41,15 +41,21 @@
...
+          <% component.loading_icon do %>
+            <div class="py-4">
+              <%= render LoadingComponent.new %>
+            </div>
+          <% end %>
+          <%= render Developers::CardComponent.with_collection(@query.records) %>
         <% end %>
-        <%= render Developers::CardComponent.with_collection(@query.records) %>
+      <% else %>
+        <div class="mt-8">
```
joemasilotti, line 56:
```suggestion
        <div class="my-16">
```

config/locales/en.yml
```diff
@@ -244,6 +244,10 @@ en:
         title: Work preferences
     index:
       description: "%{developers_count}+ Ruby on Rails developers looking for their next gig. Juniors to seniors and everyone in between, you'll find them all here."
+      empty_state:
+        body: No developers match these attributes.
+        dev_body_html: Run <code>bin/rails db:seed</code> to create development data.
+        title: No matching developers
```
joemasilotti, line 250:
```suggestion
        body: Try changing or clearing your filters.
        dev_body_html: Run <code>bin/rails db:seed</code> to create development data.
        title: No developers found
```

joemasilotti (changes requested):
Looking good! I requested some small copy and design tweaks.

Other than that, I don't love the inline logic for dynamically setting the copy. Let's create a new component that inherits from `EmptyStateComponent` and move the logic into the new component.

app/views/developers/index.html.erb
bensheldon: Your suggestion of `my-16` removed all the margin 😬 I changed it to `mt-16`.

bensheldon:
@joemasilotti I believe the best change is to add a new argument to the existing `EmptyStateComponent`, called `unseeded_body:` and move the logic into the existing component. Mainly because I liked keeping the translations relative to the view template (rather than the ViewComponent); and maybe some other EmptyStateComponents would be able to use that functionality too.

If you don't like that, I can make a EmptyDevelopersComponent and push the logic inside of there.

I was unable to test that functionality though. When I stub out the Rails.env, for some reason a Stripe API request get's dispatched 🤷  

```ruby
# test/components/empty_state_component_test.rb

class EmptyStateComponentTest < ViewComponent::TestCase
  # ...
  test "Renders title, unseeded_body and icon" do
    title = "No notifications"
    body = "Nothing to see"
    unseeded_body = "Seed the database"

    User.destroy_all
    Rails.stub(:env, ActiveSupport::StringInquirer.new('development')) do
      render_inline(EmptyStateComponent.new(title:, body:, unseeded_body:, icon: "icons/brands/github.svg"))

      assert_selector("h3", text: title)
      assert_selector("p", text: unseeded_body)
      assert_selector "svg[title='GitHub logo']"
    end
  end
end

  ```

**Comment by joemasilotti** [comment]
`EmptyStateComponent` is generic and shouldn't know anything about what "seeds" are. Subclasing allows us to add some more specific logic without inflating that class. 

Instead of stubbing Env we can inject the environment via an initializer.

```
def initialize(show_unseeded_message:) 

Developers::EmptyStateComponent.nee(show_unseeded_message: Rails.env.dev?)
```

**Comment by bensheldon** [comment]
@joemasilotti Thanks for the feedback! I added a new `Developers::EmptyStateComponent` that sub-renders the initial `EmptyStateComponent`. I did that because I didn't like changing the `initialize` method signature of a subclass (Liskov Substitution). 

Observation: I'm not sure if where I ended up is far from where you would have gone yourself, but it does seem like there's a lot more lines of code just to [rewrite the initial ternary in the template](https://github.com/joemasilotti/railsdevs.com/pull/484/commits/2822a840283b99e539f2db28277c9db778559cbd#diff-bd7dfd6db3f6a09b5ed98fe1578c67c0ac609d5525738ce977d572bd3e451c36R57) by introducing a new component. I don't have visibility into opportunities for reuse of the component though, so if there are future plans for lots of Developers views, I could see it being worth it.

**Comment by joemasilotti** [comment]
> Observation: I'm not sure if where I ended up is far from where you would have gone yourself, but it does seem like there's a lot more lines of code just to [rewrite the initial ternary in the template](https://github.com/joemasilotti/railsdevs.com/pull/484/commits/2822a840283b99e539f2db28277c9db778559cbd#diff-bd7dfd6db3f6a09b5ed98fe1578c67c0ac609d5525738ce977d572bd3e451c36R57) by introducing a new component. I don't have visibility into opportunities for reuse of the component though, so if there are future plans for lots of Developers views, I could see it being worth it.

Apologies, my quickly written response earlier wasn't very clear. I meant to only pass in the environment to the View Component and then do the querying and dev check inside the component.

It's not documented (yet), but all of the components inside a domain folder are not meant to be reused outside of their specific application. They should be treated like glorified partials (with testing). The components that live at the root level of `app/components/` are meant to be reused and shouldn't have knowledge of any (or very little) domain models.

(I'm hoping to document some of this in `app/components/README.md` in the near future.)

So for this component I would rather have it know how to query for zero developers if there are no results and if we are in the development environment. That way we can pass only the environment flag or nothing at all in.

Does that make sense?

**Comment by bensheldon** [comment]
@joemasilotti I think it makes sense. I made changes inline with my understanding. Hopefully it meets your needs 😄 

More broadly, I think this is just some unsurfaced complexity. Thinking through it:

- Seeded/unseeded state is global to the application.
- The EmptyStateComponent (or even the new Developer::EmptyStateComponent) doesn't have any insight into the conditions/state under which it is rendered or not (e.g. it doesn't have access to which developers are empty and if a filter has been applied). 
- Therefore, I think the component is going to be debatably messy in code because of the mismatch between having to both access global state, but also not consider local/specific state, but be a highly localized component.
- ...and thus if this set of changes doesn't meet your expectations, my recommendation would be to use a banner/statusbar that was more application layout-level because I think that's going to more closely align the visual-data-code hierarchy.

**Comment by bensheldon** [comment]
@joemasilotti sounds good 👍 I'll slightly rewrite the issue and let's split it into 2 PRs. That way we can land the relatively simple empty state and get that benefit to users, and then land the development message.

**Comment by joemasilotti** [comment]
Good idea. Let me know when this one is ready for another review.

**Comment by bensheldon** [comment]
@joemasilotti this is ready for another review 😄

**Comment by bensheldon** [comment]
I also just pushed up #494 for db:seed hinting. I can rebase it once this PR goes in.

File: app/views/developers/index.html.erb [hunk 4]
```diff
@@ -41,15 +41,21 @@
...
+          <% component.loading_icon do %>
+            <div class="py-4">
+              <%= render LoadingComponent.new %>
+            </div>
+          <% end %>
+          <%= render Developers::CardComponent.with_collection(@query.records) %>
         <% end %>
-        <%= render Developers::CardComponent.with_collection(@query.records) %>
+      <% else %>
+        <div class="mt-16">
```
joemasilotti, line 56:
```suggestion
        <div class="my-16">
```

For mobile.

test/integration/developers_test.rb
```diff
@@ -11,6 +11,12 @@ class DevelopersTest < ActionDispatch::IntegrationTest
     assert_select "h2", developers(:one).hero
   end
 
+  test "no developers produce empty state" do
+    Developer.destroy_all
+    get developers_path
+    assert_select "h3", "No developers found"
```
joemasilotti, line 17:
```suggestion
    assert_select "h3", I18n.t("developers.index.empty_state.title")
```

joemasilotti (changes requested):
Two tiny suggestions and we are good to merge!

joemasilotti (approved):

joemasilotti:
Awesome! Thanks for the contribution and feedback on this one. I like where it ended up much better than what I originally imagined. 💪

PR #485, opened 2022-06-11
Add Read At Indicator

troyizzle:
This PR should add a new checkmark read at indicator if the message has
been read and is the last message in the conversation. The popup does
not look the best but can adjust with any feedback. There is a case in
the fixtures where a message can have two notifications, is that a use
case to worry about?

* Add a new feature `message_read_indicator`
* Add a new component `Messages::ReadAt`

Screen Shot 2022-06-11 at 4 06 49 PM
Screen Shot 2022-06-11 at 4 07 20 PM


joemasilotti:
Thanks for the PR! After playing with the feature a bit I'd like to make a few changes to the original issue.

If we are only going to show a single checkmark then I don't think it should be inside the message `<div>`. Instead, let's move it in between the message list and the input field.

image

Having it inside the message `<div>` means we would have to also remove it when a new message is sent since they are done via Turbo Streams. Keeping it outside of the `<ul>` but inside the `#messages` Turbo Frame means it will get updated automatically when a new message is sent.

Here's some starting HTML for how it could look.

```diff
diff --git a/app/views/conversations/show.html.erb b/app/views/conversations/show.html.erb
index bb40fa7..340e513 100644
--- a/app/views/conversations/show.html.erb
+++ b/app/views/conversations/show.html.erb
@@ -22,6 +22,21 @@

   <ul id="messages" class="mt-8 space-y-2">
     <%= render @conversation.messages %>
+
+    <!-- When last message has a read notification from the recipient. -->
+    <div class="flex justify-end my-3 mr-4">
+      <div class="flex items-center justify-center space-x-2 text-sm">
+        <%= inline_svg_tag "icons/solid/check.svg", class: "h-5 w-5 text-green-600" %>
+        <p class="text-gray-500">Message read.</p>
+      </div>
+    </div>
+
+    <!-- Otherwise. -->
+    <div class="flex justify-end my-3 mr-4">
+      <div class="flex items-center justify-center space-x-2 text-sm">
+        <p class="text-gray-500">Message unread.</p>
+      </div>
+    </div>
   </ul>

   <%= render Businesses::HiringFeeComponent.new(current_user, @conversation) %>
```

---

I also don't think we need the tooltip. If we migrate these read indicators to _each_ message then I'd like to add them. But for now only having the checkmark feels good enough. Knowing _when_ someone read the message, IMO, doesn't feel like it adds much value.

---

Finally, each message might have multiple notifications associated with it. The query should make sure to only check that the "other" user in the conversation has read it. You can also use `Notification#read?` instead of checking if the column isn't null.

troyizzle:
@joemasilotti hopefully this is ready to go!

app/components/conversations/read_indicator_component.html.erb
```diff
@@ -0,0 +1,16 @@
+<% return unless render? %>
+
+<% if show_read? %>
+  <div class="flex justify-end my-3 mr-4">
+    <div class="flex items-center justify-center space-x-2 text-sm">
+      <%= inline_svg_tag "icons/solid/check.svg", class: "h-5 w-5 text-green-600" %>
+      <p class="text-gray-500">Message read.</p>
```
joemasilotti, line 7: Please localize this string.

app/components/conversations/read_indicator_component.html.erb
```diff
@@ -0,0 +1,16 @@
...
+    </div>
+  </div>
+<% else %>
+  <div class="flex justify-end my-3 mr-4">
+    <div class="flex items-center justify-center space-x-2 text-sm">
+      <p class="text-gray-500">Message unread.</p>
```
joemasilotti, line 13: Please localize this string.

app/models/feature.rb
```diff
@@ -3,6 +3,8 @@ def self.enabled?(feature_name, user:)
     case feature_name.to_sym
     when :new_developer_fields_banner
       true
+    when :message_read_indicator
+      true
```
joemasilotti, line 7:
```suggestion
      !Rails.env.production? || user&.admin?
```

app/views/messages/_message.html.erb
```diff
@@ -11,6 +11,6 @@
   </div>
 
   <div class="mt-4 text-sm text-gray-800 prose">
-    <%== message.body_html %>
-  </div>
+  <%== message.body_html %>
+</div>
```
joemasilotti, line 15: Please revert this change.

test/components/conversations/read_indicator_component_test.rb
```diff
@@ -0,0 +1,44 @@
+require "test_helper"
+
+module Messages
```
joemasilotti, line 3: Let's replace the stubbing of the methods with actual `Message` and `Conversation` data. It feels like we are testing a private method (even though that's not entirely true).

test/models/conversation_test.rb
```diff
@@ -65,6 +65,47 @@ class ConversationTest < ActiveSupport::TestCase
...
+
+    assert_equal messages(:from_business), conversation.latest_message
+  end
+
+  test "latest_message_read_by_other_user? returns false if no messages" do
+    conversation = conversations(:one)
+    conversation.messages.destroy_all
+    user = users(:developer)
+
+    assert_not conversation.latest_message_read_by_other_user?(user)
```
joemasilotti, line 79:
```suggestion
    refute conversation.latest_message_read_by_other_user?(user)
```

joemasilotti: Same goes for the rest of the file.

app/models/conversation.rb
```diff
@@ -33,6 +33,17 @@ def hiring_fee_eligible?
     developer_replied? && created_at <= 2.weeks.ago
   end
 
+  def latest_message_read_by_other_user?(other_user)
```
joemasilotti, line 36: I think this can be refactored to `#latest_message_read_by_other_recipient?` where we pass in the current user and use `#other_recipient) to get the other recipient.

joemasilotti (changes requested):
Looking good! I made a few code suggestions and one or two style change requests. But the architecture and core of the code is solid.

The biggest change is that I would also like to make sure the "Message unread" text appears when sending the message. It feels weird for it to show up only after you refresh the page. This will require wrapping the indicator in a new Turbo Frame tag for `messages/create.turbo_stream.erb` to update.

test/components/conversations/read_indicator_component_test.rb
troyizzle: I went with stubbing because it felt like the actually code had coverage through model tests but I can change that!

joemasilotti:
Looks great! Thanks for the contribution. Merging this in now.

PR #494, opened 2022-06-14
Warn to seed the database when no Developer records in development

bensheldon:
Connects to #472; broken out of #484 .

Adds a warning when in the Rails Development environment and the database has no Developer records to seed the database.

Screen Shot 2022-06-15 at 5 52 52 PM

- [x] I added significant changes and product updates to the [changelog](CHANGELOG.md) -- I think #484 will cover it.

### DIFF
--- a/app/views/developers/index.html.erb
+++ b/app/views/developers/index.html.erb
@@ -41,6 +41,7 @@
 
     <!-- Paginated list of developers -->
     <div class="col-span-12 md:col-span-9">
+      <%= render Developers::UnseededWarningComponent.new(seedable: Rails.env.development?) %>
       <%= render Developers::CardComponent.with_collection(@query.featured_records, featured: true) %>
 
       <%= render PagyPaginatorComponent.new(id: "developers", pagy: @query.pagy, url_array: [:developers], container_classes: "divide-y divide-gray-200", options: @query.filters) do |component| %>
--- a/app/views/developers/index.html.erb
+++ b/app/views/developers/index.html.erb
@@ -41,7 +41,6 @@
 
     <!-- Paginated list of developers -->
     <div class="col-span-12 md:col-span-9">
-      <%= render Developers::UnseededWarningComponent.new(seedable: Rails.env.development?) %>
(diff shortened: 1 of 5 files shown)


app/views/developers/index.html.erb
```diff
@@ -41,6 +41,7 @@
 
     <!-- Paginated list of developers -->
     <div class="col-span-12 md:col-span-9">
+      <%= render Developers::UnseededWarningComponent.new(seedable: Rails.env.development?) %>
```
joemasilotti, line 44: Let's move this to `app/views/shared/_main.html.erb:2` so it shows up on every page.

test/components/developers/unseeded_warning_component_test.rb
```diff
@@ -0,0 +1,21 @@
...
+module Developers
+  class UnseededWarningComponentTest < ViewComponent::TestCase
+    test "hidden when unseedable" do
+      render_inline UnseededWarningComponent.new(seedable: false)
+      assert_no_text "Run bin/rails db:seed"
+    end
+
+    test "hidden when seedable but developers exist" do
+      render_inline UnseededWarningComponent.new(seedable: true)
+      assert_no_text("Run bin/rails db:seed")
```
joemasilotti, line 12: Let's consolidate using parenthesis or not for these assertions.

joemasilotti (changes requested):
Two minor requests and we are good to merge!

test/components/developers/unseeded_warning_component_test.rb
bensheldon: Nice catch. I removed them to be consistent 👍

app/views/developers/index.html.erb
bensheldon: 👍

bensheldon:
@joemasilotti I've updated this PR based on your feedback. Looking sharp ✨ 

Screen Shot 2022-06-15 at 5 52 52 PM

joemasilotti:
Excellent - thanks for the contribution!

PR #497, opened 2022-06-15
Move APNs p8 key to credentials

joemasilotti:
This PR moves the p8 file used to send APNs push notifications to encrypted credentials.

Thanks to @marcoroth for calling out that it shouldn't be public and @NateMatykiewicz for notes on how to configure it via Noticed.


natematykiewicz:
Looks great!

PR #507, opened 2022-06-18
Long-tail SEO pages for /developers

### LINKED ISSUE #506: Generate long-tail SEO pages for filter combinations
The goal of this issue is to generate some SEO juice for specific combinations of developer filters. For example, "freelance Ruby on Rails developers", "senior Ruby on Rails developers", or "Ruby on Rails developers in Canada".

To start, we can generate unique pages based on the existing filters and tweak the `<h1>` to help differentiate the pages to search engines. Then, ask for them to be indexed in `config/sitemap.rb`. Here's what a first pass could look like.

## New pages

Iterate over all the possible combinations of:

1. If a single `RoleType` is selected then add this to the heading. e.g. `Hire junior Ruby on Rails developers`
2. If both `:part_time_contract` and `:full_time_contract` are checked then add "freelance" to the heading. e.g. `Hire freelance Ruby on Rails developers`
3. If a single _top_ country is selected then add "in #{country}" to the heading. e.g. `Hire Ruby on Rails developers in Canada`

Iterating over all of these combinations creates 50 unique pages (5 role types x (freelance or not) x top 5 countries) to be indexed. 

## `<h1>` logic

Pseudo-code logic for how the header is rendered.

```ruby
role_type = query.role_type.one? ? query.role_type.first.humanize : ""
freelance = query.full_time_contract? && query.part_time_contract? && !query.full_time_epmployment? ? "freelance" : ""
country = query.countries.one? ? "in #{query.countries.first}" : ""

<h1>"Hire #{role_type} #{freelance} Ruby on Rails developers #{country}"</h1>
```

This leads to pages like:

* Hire senior Ruby on Rails developers in United Kingdom
* Hire freelance Ruby on Rails developers
* Hire staff freelance Ruby on Rails developers in Brazil

## Implementation details

`config/sitemap.rb` needs paths to generate the sitemap. But we also need display logic to render the header. I'm thinking of a new model that a new View Model and sitemap will use.

The model will be responsible for doing the logic of "is there one `role_type` selected? and generating the unique URLs. The View Component will be responsible for rendering the header based on the model.

An integration test can also make sure that with all the generated URLs (for the sitemap) that no two `<h1>` rags are unique.

*

joemasilotti:
Closes #506. Description incoming (but see the issue for more details in the meantime).

## Pull request checklist

### DIFF
--- a/app/models/developers/title.rb
+++ b/app/models/developers/title.rb
@@ -1,0 +1,45 @@
+module Developers
+  class Title
+    private attr_reader :query
+
+    def initialize(query)
+      @query = query
+    end
+
+    def title
+      components = []
+      components << "Hire"
+      components << role_level if role_level
+      components << "freelance" if freelance?
+      components << "Ruby on Rails developers"
+      components << "in #{country}" if country
+      components.join(" ")
+    end
+
+    private
+
+    def role_level
+      if query.role_levels.one?
+        if query.role_levels.first == :mid
+          "mid-level"
+        elsif query.role_levels.first == :c_level
+          "C-level"
+        else
+          query.role_levels.first
+        end
+      end
+    end
+
+    def freelance?
+      query.role_types.include?(:part_time_contract) &&
+        query.role_types.include?(:full_time_contract) &&
+        !query.role_types.include?(:full_time_employment)
+    end
+
+    def country
+      if query.countries.one?
+        query.countries.first
+      end
+    end
+  end
+end
--- a/app/models/developers/title.rb
+++ b/app/models/developers/title.rb
@@ -8,18 +8,28 @@ module Developers
 
     def title
       components = []
-      components << I18n.t("hire", scope:)
+
+      # i18n-tasks-use t("developers.title.hire")
+      components << i18n("hire")
+
       components << role_level if role_level
-      components << I18n.t("freelance", scope:) if freelance?
-      components << I18n.t("ruby_on_rails_developers", scope:)
-      components << I18n.t("in_country", country:, scope:) if country
+
+      # i18n-tasks-use t("developers.title.freelance")
+      components << i18n("freelance") if freelance?
+
+      # i18n-tasks-use t("developers.title.ruby_on_rails_developers")
+      components << i18n("ruby_on_rails_developers")
+
+      # i18n-tasks-use t("developers.title.in_country")
+      components << i18n("in_country", country:) if country
+
       components.join(" ")
     end
 
     private
 
-    def scope
-      "developers.title"
+    def i18n(key, options = {})
+      I18n.t(key, **options.merge(scope: "developers.title"))
     end
 
     def role_level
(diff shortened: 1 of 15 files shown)


app/models/developers/title.rb
```diff
@@ -0,0 +1,45 @@
...
+      if query.role_levels.one?
+        if query.role_levels.first == :mid
+          "mid-level"
+        elsif query.role_levels.first == :c_level
+          "C-level"
+        else
+          query.role_levels.first
+        end
+      end
+    end
```
joemasilotti, line 31: I feel like I should be able to get these from the translations file but can't figure out how.

app/models/developers/title.rb
```diff
@@ -0,0 +1,45 @@
...
+
+    def title
+      components = []
+      components << "Hire"
+      components << role_level if role_level
+      components << "freelance" if freelance?
+      components << "Ruby on Rails developers"
+      components << "in #{country}" if country
+      components.join(" ")
+    end
```
joemasilotti, line 17: Translating this in the future will be very difficult. How can I make this easier to manage?

app/models/developers/title.rb
benedikt: You can use a fallback for this:

joemasilotti: Hmm, is there any where I can use the existing `activerecord.attributes.role_level` keys?

app/models/developers/title.rb
```diff
@@ -0,0 +1,49 @@
...
+      components << I18n.t("ruby_on_rails_developers", scope:)
+      components << I18n.t("in_country", country:, scope:) if country
+      components.join(" ")
+    end
+
+    private
+
+    def scope
+      "developers.title"
+    end
```
joemasilotti, line 23: I feel like there's a better approach to this. But trying to make anything more dynamic makes `bundle exec i18n-tasks missing en` fail because it can't find the translations.

app/models/developers/title.rb
benedikt: Yeah, just adjust the scope: `[:activerecord, :attributes, :role_level]`

app/models/developers/title.rb
joemasilotti: I've gone with manual commenting of the used keys for now. I don't love it, but it works.

app/models/developers/title.rb
```diff
@@ -0,0 +1,55 @@
...
+
+      # i18n-tasks-use t("developers.title.in_country")
+      components << i18n("in_country", country:) if country
+
+      components.join(" ")
+    end
+
+    private
+
+    def i18n(key, options = {})
```
dorianmariefr, line 31: I usually name this method `t` similar to how it’s done in the views

dorianmariefr: Also I would do **options so that

dorianmariefr (approved):
Well done, maybe you can also optimize the description?

app/models/developers/title.rb
joemasilotti: Sadly `i18n-tasks` thinks that using `t` means we are using the view helper and gets all confused.

Issue #237, opened 2022-01-13
Account vs. User


joemasilotti:
Well, it was bound to come eventually! A potential customer expressed interest in needing multiple users per paid account. I'm still working through how this will all work, so this comment will serve as a living doc of how things will be handled.

### Proposal

1. Create an `Account` model with an `owner_id` column, `User`
1. Add `account_id` to `User` model, `belongs_to: :account`
1. Migrate `user_id` on `Business` and `Developer` to `account_id`
    * Right now each account will only have the one user, so create a single account per user
    * Set the `owner` attribute on the account to the user
1. Migrate all related queries from `current_user` to `current_user.account`
    * Create abstraction, `#current_account`, in `ApplicationHelper` or `ApplicationController`
    * Pundit policies, `PaywalledComponent`, `DeveloperPrimaryActionComponent`
1. Build out invitation flow for the account owner to invite a non-owner user associated with the account
    * To start, they have full permissions with the exception of inviting someone else

### Open questions

1. How does the noticed integration change?
    * The account should now be notified, so maybe all users associated with the account get the notification.
    * But if person A reads one, does person B's notification (from the same account) also get read?
1. How does the Pay integration change?
    * `pay_customer` expects an email and name, but this will live on `Account` now. How can the owner's details be passed to/from Pay?
1. Having an additional query (`user.account.object`) every time we need data feels like it could be bad, but I might be pre-optimizing.

---

Perhaps this [~30 line multi-tenant strategy](https://dev.to/kolide/a-rails-multi-tenant-strategy-thats-30-lines-and-just-works-58cd) can help?

tommasongr:
I'm not an expert but for an app I'm working on I used this configuration.

Basically Accounts and Users are associated through a Membership model. This way a User can have multiple accounts and an account can have multiple owners or other roles.

You still have to ensure that the last owner on the account is not deleted before the account itself but apart from that this setup should cover all the good points from @subdigital (I guess).

### Account
```ruby
class Account < ApplicationRecord
  has_many :memberships, dependent: :destroy
  has_many :users, through: :memberships

  accepts_nested_attributes_for :memberships
end
```

### User
```ruby
class User < ApplicationRecord
  has_many :memberships, dependent: :destroy
  has_many :accounts, through: :memberships
end
```

### Membership
```ruby
class Membership < ApplicationRecord
  belongs_to :account
  belongs_to :user

  accepts_nested_attributes_for :account

[...]

PR #508, opened 2022-06-20
Naming Things - Data migration and scaffolding for Account

joemasilotti:
This is the code from the the railsdevs episode of [Naming Things](https://www.namingthings.org/) with @joeldrapper.

The goal was to rework how `Business` vs. `Developer` is managed in the app. There is currently a lot of code required to know what "context" a user is in. We discovered that explicitly setting a context could be a great start.

The changes kick off the first steps towards having accounts and membership. To start, your current account will be your business account. If you don't have one then it will fall back to your developer account.

For most people they will never even see the idea of an account. But for the few who have both types of profiles a context switching UI will need to be added.

I like this approach because it means we can incrementally add on features of the accounts without having to worry about doing it all at once! Note that there are a few TODOs noted in

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -7,6 +7,7 @@ class ApplicationController < ActionController::Base
 
   rescue_from Pundit::NotAuthorizedError, with: :user_not_authorized
 
+  before_action -> { Current.user = current_user }
   around_action :set_locale
   helper_method :resolve_locale
   helper_method :turbo_native_app?
--- a/app/models/account.rb
+++ b/app/models/account.rb
@@ -1,0 +1,8 @@
+class Account < ApplicationRecord
+  has_many :account_memberships
+  has_many :users, through: :account_memberships
+
+  delegate :name, to: :profile
+
+  delegated_type :profile, types: %w[Business Developer]
+end
--- a/app/models/account_membership.rb
+++ b/app/models/account_membership.rb
@@ -1,0 +1,6 @@
+class AccountMembership < ApplicationRecord
+  belongs_to :user
+  belongs_to :account
+
+  enum role: %w[admin]
+end
--- a/app/models/current.rb
+++ b/app/models/current.rb
@@ -1,0 +1,18 @@
+class Current < ActiveSupport::CurrentAttributes
+  attribute :user
+
+  def account
+    return nil unless account_id.present?
+    @account ||= user&.accounts&.find(account_id)
+  end
+
+  private
+
+  def account_id
+    session[:account_id] ||= default_account&.id
+  end
+
+  def default_account
+    user.business || user.developer
+  end
+end
--- a/app/models/profile.rb
+++ b/app/models/profile.rb
@@ -1,0 +1,6 @@
+module Profile
+  extend ActiveSupport::Concern
+
+  included do
+  end
+end
--- a/db/migrate/20220620170036_create_accounts.rb
+++ b/db/migrate/20220620170036_create_accounts.rb
@@ -1,0 +1,43 @@
+class CreateAccounts < ActiveRecord::Migration[7.0]
+  def change
+    create_table :accounts do |t|
+      t.belongs_to :profile, polymorphic: true, null: false
+
+      t.timestamps
+    end
+
+    create_table :account_memberships do |t|
+      t.belongs_to :user, null: false, foreign_key: true
+      t.belongs_to :account, null: false, foreign_key: true
+
+      t.string :role, default: "admin", index: true
+
+      t.timestamps
+    end
+
+    # TODO: Write a test to ensure the correct account is migrated.
+    [Business, Developer].each do |model|
+      model.in_batches do |batch|
+        account_attrs = batch.map do |business|
+          {
+            profile_id: business.id,
+            profile_type: model
+          }
+        end
+
+        accounts = Account.insert_all(
+          account_attrs, returning: %w[profile_id id]
+        ).rows.to_h
+
+        membership_attrs = batch.map do |business|
+          {
+            user_id: business.user_id,
+            account_id: accounts[business.id]
+          }
+        end
+
+        AccountMembership.insert_all(membership_attrs)
+      end
+    end
+  end
+end
--- a/app/models/business.rb
+++ b/app/models/business.rb
@@ -2,6 +2,7 @@ class Business < ApplicationRecord
   include Avatarable
   include Businesses::Notifications
   include PersonName
+  include Profile
 
   enum :developer_notifications, %i[no daily weekly], default: :no, suffix: true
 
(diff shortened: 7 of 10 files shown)


app/controllers/application_controller.rb
```diff
@@ -7,6 +7,7 @@ class ApplicationController < ActionController::Base
 
   rescue_from Pundit::NotAuthorizedError, with: :user_not_authorized
 
+  before_action -> { Current.user = current_user }
```
joeldrapper, line 10: We might want to trigger this form a devise hook rather than assuming it will happen after devise authentication here.

app/models/account.rb
```diff
@@ -0,0 +1,8 @@
+class Account < ApplicationRecord
+  has_many :account_memberships
+  has_many :users, through: :account_memberships
+
+  delegate :name, to: :profile
```
joeldrapper, line 5: Alternatively, we could delegate all missing methods to the concrete account types.

```suggestion
  delegate_missing_to :profile
```

app/models/account_membership.rb
```diff
@@ -0,0 +1,6 @@
+class AccountMembership < ApplicationRecord
```
joeldrapper, line 1: I love having this as a record. It's so much better than an `admin` boolean.

app/models/current.rb
```diff
@@ -0,0 +1,18 @@
+class Current < ActiveSupport::CurrentAttributes
+  attribute :user
+
+  def account
+    return nil unless account_id.present?
```
joeldrapper, line 5: It might be better to raise from here if the `account_id` isn't present. It means you're looking for an account when not signed in. Alternatively, return a null account object with the same interface as real accounts so there's no error. We should lean hard one way or the other.

db/migrate/20220620170036_create_accounts.rb
```diff
@@ -0,0 +1,43 @@
...
+        membership_attrs = batch.map do |business|
+          {
+            user_id: business.user_id,
+            account_id: accounts[business.id]
+          }
+        end
+
+        AccountMembership.insert_all(membership_attrs)
+      end
+    end
```
joeldrapper, line 41: Let's extract an operation class here so we can test this logic carefully.

joeldrapper:
Besides running this data migration, we need to ensure all new users, businesses and developer profiles are connected through this new architecture too.

joemasilotti:
Thanks again for the opportunity to pair together, @joeldrapper! I had a blast and learned a ton.

One thing I'm a little confused about is how to actually use `Current.account`. I remember you mentioning we don't have to pass it around all over the place.

Do you have an example or code snippet on how I could use it? Say, for example, when creating a new developer in `DevelopersController`?

app/models/current.rb
```diff
@@ -0,0 +1,18 @@
...
+    @account ||= user&.accounts&.find(account_id)
+  end
+
+  private
+
+  def account_id
+    session[:account_id] ||= default_account&.id
```
wasabigeek, line 12: This felt like a bit of an unexpected side effect (assigning something to the `session`).

Also, out of curiosity, are there other areas where the `account_id` might get set or overridden?

app/models/account.rb
```diff
@@ -0,0 +1,8 @@
...
+
+  delegated_type :profile, types: %w[Business Developer]
```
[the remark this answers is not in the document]
joeldrapper: We're saying here that Developer accounts and Business accounts are both _accounts_ so that they can share things in common and can often be used polymorphically through `Current.account`.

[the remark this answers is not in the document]
wasabigeek: Thanks! My understanding of the [original issue](https://github.com/joemasilotti/railsdevs.com/issues/237) is that Accounts represent a "paying entity" e.g. a business that is paying railsdevs. Is my interpretation too narrow? I don't understand why we'd need a "Developer account" I guess 🙏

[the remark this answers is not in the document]
wasabigeek: Ah, my bad, after reading the thread more I see that it's also something to hang subscriptions and permissions onto, so they can apply to a group of users, or a single user.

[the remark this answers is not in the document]
joeldrapper: My understanding from discussing this with @joemasilotti was that users can have a business profile or a developer profile or both, and in the future we may want users to be able to have more than one business profile.

Additionally, people may want to add more than one user to a single business account. The new architecture covers all of these cases. Users can have many accounts, and each account is either a business or developer profile. They have accounts through account membership, an object that lets us model their role within that account (e.g. owner, manager, employee, etc.)

app/models/profile.rb
```diff
@@ -0,0 +1,6 @@
+module Profile
+  extend ActiveSupport::Concern
+
+  included do
+  end
```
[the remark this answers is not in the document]
jayeclark: Hey @joemasilotti I've got reliable internet again so am finally starting to sort through & digest this PR. If I'm reading this correctly, a user would be associated with various accounts (each of which has their own profile) but there would not actually be a profile (avatar/name/bio) for that user specifically? 

Doesn't that mean that the only way a user gets a contact name/avatar is through a developer or business account? (I realize that's how it's currently set up, I guess I'm just asking - is that the plan for the future as well?) Wouldn't that mean that all users associated with XYZ Business would appear to be that particular business as they acted in the app? (And that I'd have a different avatar and potentially a different name based on what context I was operating in?)

This might be the intended way for it to be set up. But it feels off to me that if I'm using RailsDevs I don't really 'exist' as a person other than through the accounts I'm associated with (and for some of those accounts I might appear to be a completely _different_ person than who I am - ie the main contact person for the business.)

db/migrate/20220620170036_create_accounts.rb
```diff
@@ -0,0 +1,43 @@
...
+      t.belongs_to :user, null: false, foreign_key: true
+      t.belongs_to :account, null: false, foreign_key: true
+
+      t.string :role, default: "admin", index: true
+
+      t.timestamps
+    end
+
+    # TODO: Write a test to ensure the correct account is migrated.
+    [Business, Developer].each do |model|
```
jayeclark, line 19: I'm confused by this migration. It looks like this code is iterating through every business & developer record (in batches) and creating an account for each (along with an associated profile details). But then on lines 21 & 32 only 'business' is used as the variable name. The record that actually gets yielded to the block could be either a business or a developer, right? I realize the variable `|business|` could be called anything under the sun & it wouldn't make a functional difference, but from a readability standpoint it feels confusing to have Business and Developer called out on line 19 but then only reference businesses later in the block.

joeldrapper: Yes, you're right. We extracted the `[Business, Developer]` loop after writing this initially for businesses, and then ran out of time and stopped pairing before we could rename the variable.

joemasilotti:
> @joemasilotti sorry we never really got into this. Want to pair again sometime and see what we can do?

Yes, please. I'd love to!

joemasilotti:
Closing this as I've let it gone stale and am trying to clean up some old work on the repo.

Issue #504, opened 2022-06-18
Site-wide banner indicating your profile is invisible


joemasilotti:
Developers have been marking their profile as invisible to not show up on the site. Some of these are because they found a job and are longer looking! Odds are they won't come back to the site for a while.

But for the others, I'd like to let them know a bit more aggressively that their profile isn't appearing for anyone else.

## A PR addressing this issue should:

- [ ] Add a banner to every page indicating such linking to edit your profile (if your developer profile is invisible).

Design can be copied form `NewFieldsComponent` and rendered from `app/views/shared/_main.html.erb`.

joshio1:
Hi there! 👋
I'd be happy to work on this one!

joemasilotti:
Sure thing – all yours!

joshio1:
Thanks, I have a few questions:

1. When the user clicks on the CTA and is on the `Edit Developers` page, do we still show the banner? Or is there any change we need to do on that screen?
2. Do you have anything specific in mind as far as the verbiage for the banner is concerned? (for both the title and CTA)
3. I saw that the NewFieldsComponent does not have an equivalent version in other languages. Do we require localization for this component?

joemasilotti:
Closed via #509.

PR #509, opened 2022-06-21
Add invisible banner if profile is invisible

joshio1:
Site-wide banner indicating your profile is invisible https://github.com/joemasilotti/railsdevs.com/issues/504

- [x] Add a banner to every page indicating such linking to edit your profile (if your developer profile is invisible).

## Pull request checklist

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### June
 
+* June 26 - Site-wide banner indicating your profile is invisible - #509 @joshio1
 * June 24 - Added configuration to lazily load images - #520 @joshio1
 * June 23 - Added "Invisiblize" button on business profiles (for admins) - #510 @DRBragg
 * June 19 - Admins can view invisible developer profiles - #505 @kaushikhande
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,7 +4,7 @@
 
 ### June
 
-* June 26 - Site-wide banner indicating your profile is invisible - #509 @joshio1
+* June 28 - Site-wide banner indicating your profile is invisible - #509 @joshio1
 * June 24 - Added configuration to lazily load images - #520 @joshio1
 * June 23 - Added "Invisiblize" button on business profiles (for admins) - #510 @DRBragg
 * June 19 - Admins can view invisible developer profiles - #505 @kaushikhande
--- a/app/components/developers/banner_component.html.erb
+++ b/app/components/developers/banner_component.html.erb
@@ -1,0 +1,1 @@
+<%= render selected_banner if selected_banner %>
--- a/app/components/developers/banner_component.html.erb
+++ b/app/components/developers/banner_component.html.erb
@@ -1,1 +1,1 @@
-<%= render selected_banner if selected_banner %>
+<%= render selected_banner %>
--- a/app/components/developers/banner_component.rb
+++ b/app/components/developers/banner_component.rb
@@ -1,0 +1,20 @@
+module Developers
+  class BannerComponent < ApplicationComponent
+    private attr_reader :banners, :selected_banner
+
+    def initialize(banners)
+      @banners = banners
+    end
+
+    def render?
+      banners.each do |banner|
+        if !!banner.render?
+          @selected_banner = banner
+          return true
+        end
+      end
+    end
+
+    private
+  end
+end
--- a/app/components/developers/banner_component.rb
+++ b/app/components/developers/banner_component.rb
@@ -1,6 +1,7 @@
 module Developers
   class BannerComponent < ApplicationComponent
-    private attr_reader :banners, :selected_banner
+    attr_reader :selected_banner
+    private attr_reader :banners
 
     def initialize(banners)
       @banners = banners
@@ -13,8 +14,7 @@ module Developers
           return true
         end
       end
+      false
     end
-
-    private
   end
 end
--- a/app/views/shared/_main.html.erb
+++ b/app/views/shared/_main.html.erb
@@ -1,7 +1,8 @@
-<%= unseeded_warning_component = Developers::UnseededWarningComponent.new(seedable: Rails.env.development?)
-    new_fields_component = Developers::NewFieldsComponent.new(current_user, enabled: Feature.enabled?(:new_developer_fields_banner, user: current_user))
-    invisible_banner_component = Developers::InvisibleBannerComponent.new(current_user)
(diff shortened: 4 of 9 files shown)


joemasilotti (changes requested):
This looks great! Two requests and we are good to merge:

1. We now need to ensure only a single banner gets displayed. See the screenshot below. You can trigger this by seeding the database and signing in with `invisible@example.com`.
2. When you're done, can you update the CHANGELOG again?

image

joemasilotti:
Something like a `Developers::BannerComponent` could help, where it knows to render one of the three banners based on if each banner renders.

joshio1:
> Something like a `Developers::BannerComponent` could help, where it knows to render one of the three banners based on if each banner renders.

Thanks for the feedback. What are the rules as to which banner should be shown when more than one can be rendered?
For eg. if `NewFieldsComponent` and `InvisibleBannerComponent` are both enabled, then should we just show the `NewFieldsComponent`?

joemasilotti:
> What are the rules as to which banner should be shown when more than one can be rendered?

Great question! Here's the priority. If one is visible then the others should not be shown.

1. `Developers::UnseededWarningComponent`
2. `Developers::InvisibleBannerComponent`
3. `Developers::NewFieldsComponent`

joshio1:
> > What are the rules as to which banner should be shown when more than one can be rendered?
> 
> Great question! Here's the priority. If one is visible then the others should not be shown.
> 
> 1. `Developers::UnseededWarningComponent`
> 2. `Developers::InvisibleBannerComponent`
> 3. `Developers::NewFieldsComponent`

Okay, if this is the case, then do we really need the `Developers::BannerComponent` to decide which component to render?
What I'm saying is, correct me if I'm wrong here, the `Developers::UnseededWarningComponent` would never be shown together with either of those two because it needs unseeded data. Only possibility is for `Developers::InvisibleBannerComponent` to be shown together with `Developers::NewFieldsComponent`. We can avoid this by having a check in the `render?` method of the `Developers::NewFieldsComponent` to only show when the profile is visible. Am I missing anything here?
Will this suffice or do you still think it's better design-wise to have a `Developers::BannerComponent` which decides which component to show?

joemasilotti:
Thinking about this some more, I think we want a single banner component that takes an array of child components. 

Then, it can render the first one that responds to `render?`, in the order they are passed in. 

Does that make sense?

joshio1:
> Thinking about this some more, I think we want a single banner component that takes an array of child components.
> 
> Then, it can render the first one that responds to `render?`, in the order they are passed in.
> 
> Does that make sense?

Yes, I think I mostly get it. I'll take a stab at it and get back to you.

joshio1:
> Thinking about this some more, I think we want a single banner component that takes an array of child components.
> 
> Then, it can render the first one that responds to `render?`, in the order they are passed in.
> 
> Does that make sense?

Hi @joemasilotti, So I've added a `BannerComponent` that takes an array of child components and used the `render?` method to decide which component to render in an FCFS manner. This is the first time I've used this type of a pattern so please let me know if anything can be done in a different / better way.

joemasilotti:
Ohh what's FCFS? I've never heard of that before.

joshio1:
> Ohh what's FCFS? I've never heard of that before.

Sorry for using jargons. FCFS(First-come first-served) just means that the child components will be processed according to their order in the array. It's just a fancy word for queue or FIFO (First-In First-Out).

joemasilotti:
Got it - very cool!

app/components/developers/banner_component.html.erb
```diff
@@ -0,0 +1 @@
+<%= render selected_banner if selected_banner %>
```
joemasilotti, line 1:
```suggestion
<%= render selected_banner %>
```

I _think_ that should work. Because `#render?` will return `false` if we don't have a banner to render so this will never get called.

app/components/developers/banner_component.rb
```diff
@@ -0,0 +1,20 @@
...
+    def render?
+      banners.each do |banner|
+        if !!banner.render?
+          @selected_banner = banner
+          return true
+        end
+      end
+    end
+
+    private
```
joemasilotti, line 18: We can get rid of this if there are no private methods.

app/components/developers/banner_component.rb
```diff
@@ -0,0 +1,20 @@
+module Developers
+  class BannerComponent < ApplicationComponent
+    private attr_reader :banners, :selected_banner
```
joemasilotti, line 3:
```suggestion
    attr_reader :selected_banner
    private attr_reader :banners
```

It doesn't make a difference for View Components but it shows better intent.

app/views/shared/_main.html.erb
```diff
@@ -1,5 +1,7 @@
-<%= render Developers::UnseededWarningComponent.new(seedable: Rails.env.development?) %>
-<%= render Developers::NewFieldsComponent.new(current_user, enabled: Feature.enabled?(:new_developer_fields_banner, user: current_user)) %>
+<%= unseeded_warning_component = Developers::UnseededWarningComponent.new(seedable: Rails.env.development?)
+    new_fields_component = Developers::NewFieldsComponent.new(current_user, enabled: Feature.enabled?(:new_developer_fields_banner, user: current_user))
+    invisible_banner_component = Developers::InvisibleBannerComponent.new(current_user)
+    render Developers::BannerComponent.new([unseeded_warning_component, invisible_banner_component, new_fields_component]) %>
```
joemasilotti, line 4:
```suggestion
<%= render Developers::BannerComponent.new([
      Developers::UnseededWarningComponent.new(seedable: Rails.env.development?),
      Developers::NewFieldsComponent.new(current_user, enabled: Feature.enabled?(:new_developer_fields_banner, user: current_user)),
      Developers::InvisibleBannerComponent.new(current_user)
    ]) %>
```

Personal preference, but keeping it on "one" line cleans it up a bit.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### June
 
+* June 26 - Site-wide banner indicating your profile is invisible - #509 @joshio1
```
```suggestion
* June 27 - Site-wide banner indicating your profile is invisible - #509 @joshio1
```

joemasilotti (changes requested):
Looks great! A few code style requests and this is good to merge.

joshio1:
Thanks for reviewing the code. I've addressed the suggested changes.

joemasilotti:
Awesome! Looks great – thanks for the contribution.

PR #510, opened 2022-06-21
Invisibilize for businesses

### LINKED ISSUE #500: Invisibilize for businesses
There has been a tiny uptick in spam-ish requests in the past few days. Similar to how we can "invisibilize" developers, I'd like to do the same for businesses.

- [ ] Add a column to `Business` to mark if it is visible
- [ ] 404 if a business's show page is visited but that business is set to invisible
- [ ] Show a native confirm dialog before proceeding
- [ ] Add a button for admins to "Invisibilize" a business and send an email

A lot of the last to-do can use #367 as a reference.

DRBragg:
Closes #500 

This PR gives admins the ability to flag a business profile as invisible if it looks like spam. Per the request the following was done:

 - [x] Add a column to Business to mark if it is visible
 - [x] 404 if a business's show page is visited but that business is set to invisible (per a twitter conversation with Joe this wont actually show a 404 but redirect to root with a flash message, mirroring the behavior of marking a developer profile as invisible)
 - [x] Show a native confirm dialog before proceeding
 - [x] Add a button for admins to "Invisibilize" a business and send an email

Additionally a refactoring of the `DeveloperMailer` was done.  The `invisiblize` method was extracted to it's own mailer so businesses could use it also.

## Pull request checklist


app/controllers/admin/businesses/invisiblizes_controller.rb
```diff
@@ -0,0 +1,10 @@
+module Admin
+  module Businesses
+    class InvisiblizesController < Admin::ApplicationController
```
joemasilotti, line 3: I don't think the `Admin::` namespace is needed here.

app/components/businesses/admin_component.rb
```diff
@@ -0,0 +1,14 @@
+module Businesses
+  class AdminComponent < ApplicationComponent
+    private attr_reader :business, :user
```
joemasilotti, line 3: `business` is used in the view, so let's make that public and keep `user `private.

app/mailers/invisiblize_mailer.rb
```diff
@@ -0,0 +1,25 @@
...
+
+    @developer = @notification.developer
+
+    mail(
+      to: recipient.email,
+      subject: @notification.title
+    )
+  end
+
+  def to_business
```
joemasilotti, line 14:
```suggestion
  def business
```

config/routes.rb
```diff
@@ -50,6 +50,7 @@
 
     resources :businesses, only: [] do
       resources :conversations, only: :index, controller: :business_conversations
+      resources :invisiblizes, only: :create, module: :businesses
```
joemasilotti, line 53: Why is `module` needed? I don't see it for developers.

joemasilotti (commented):
Looking good! Outside of tests and some inline change requests this is close to done.

After that, a few more things:

1. I think we dropped the `text` extension for the mailers. Let's make sure to add those back in.
2. Let's make sure we have a invisible business seeded.
3. Don't forget to add the mailer previews, and check existing ones.

config/routes.rb
DRBragg: > Looking good! Outside of tests and some inline change requests this is close to done.
> 
> After that, a few more things:
> 
> 1. I think we dropped the `text` extension for the mailers. Let's make sure to add those back in.
> 2. Let's make sure we have a invisible business seeded.
> 3. Don't forget to add the mailer previews, and check existing ones.

1. Looks like the `text` extension wasn't there before I moved the file so I totally missed that! Good catch, I'll add it.
2. Copy!
3. You got it

DRBragg:
> Looking good! Outside of tests and some inline change requests this is close to done.
> 
> After that, a few more things:
> 
> 1. I think we dropped the `text` extension for the mailers. Let's make sure to add those back in.
> 2. Let's make sure we have a invisible business seeded.
> 3. Don't forget to add the mailer previews, and check existing ones.

1. Looks like the `text` extension wasn't there before I moved the file so I totally missed that! Good catch, I'll add it.
2. Copy!
3. You got it

app/controllers/admin/businesses/invisiblizes_controller.rb
DRBragg: Hm, why's that? do you mean you want me to move this out of the admin namespace?

app/mailers/invisiblize_mailer.rb
```diff
@@ -0,0 +1,25 @@
+class InvisiblizeMailer < ApplicationMailer
+  def to_developer
```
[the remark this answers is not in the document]
DRBragg: This will break the notification unfortunately. I originally hade `developer` but the noticed notification kept trying to call it's own `developer` method

app/mailers/invisiblize_mailer.rb
DRBragg: This will break the notification unfortunately, see above

config/routes.rb
DRBragg: Because we need to tell the router to look in the business dir not the admin dir.  Developers doesn't have it because that controller is only in the admin namespace

joemasilotti: > Developers doesn't have it because that controller is only in the admin namespace

I'm confused – shouldn't they both be exactly the same?

app/controllers/admin/businesses/invisiblizes_controller.rb
DRBragg: Oh! duh, by bad totally wasn't tracking. I am now 👍

app/mailers/invisiblize_mailer.rb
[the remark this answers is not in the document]
DRBragg: That's what it looks like 🤷‍♂️ 
This:
```ruby
deliver_by :email, mailer: "InvisiblizeMailer", method: :developer
```
causes a `NoMethodError` saying that `:to_sym` isn't a method on the `Developer` model instance, but using `:to_developer` works fine

config/routes.rb
DRBragg: The developer version is at `app/controllers/admin/invisiblizes_controller.rb` but the business one is at `app/controllers/admin/businesses/invisiblizes_controller.rb` so we need to tell the route to look in the `businesses` directory for the `InvisiblizesController`

joemasilotti: Got it! Let's move the developer one to match the business one then, too.

app/mailers/invisiblize_mailer.rb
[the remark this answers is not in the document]
joemasilotti: That's frustrating. Thanks for documenting – we might need to open an issue in Noticed.

Issue #511, opened 2022-06-21
Conversation page doesn't _always_ scroll to bottom on load


joemasilotti:
## Describe the bug

When viewing a conversation the page is supposed to scroll to the bottom. Sometimes, it does not.

## To reproduce

1. Seed or reset the database `$ bin/rails db:reset`
1. Sign in as `business@example.com`/`password`
2. Add 20+ messages so the bottom one is rendered off screen
    * `$ bin/rails console`
    * `20.times { Conversation.first.messages.create!(sender: Developer.first, body: "Hi!") }`
    * `exit`
4. View the conversation `http://localhost:3000/conversations/1`

## Expected behavior

The page should scroll to the bottom.

## Actual behavior

Sometimes, the page does _not_ scroll to the bottom. The following error can be observed in the web console.

```
[Error] TypeError: undefined is not an object (evaluating 'this.element.scrollIntoView')
	scrollIntoView (application-e39b1a2774a52ebb17d63cebf7aaea7d637996fa48c2e68c676bfa67e0a3cb12.js:2504)
	dispatchEvent
	dispatch (application-e39b1a2774a52ebb17d63cebf7aaea7d637996fa48c2e68c676bfa67e0a3cb12.js:2836)
	pageBecameInteractive (application-e39b1a2774a52ebb17d63cebf7aaea7d637996fa48c2e68c676bfa67e0a3cb12.js:5100)
	pageIsInteractive (application-e39b1a2774a52ebb17d63cebf7aaea7d637996fa48c2e68c676bfa67e0a3cb12.js:4646)
	pageIsComplete (application-e39b1a2774a52ebb17d63cebf7aaea7d637996fa48c2e68c676bfa67e0a3cb12.js:4650)
	(anonymous function) (application-e39b1a2774a52ebb17d63cebf7aaea7d637996fa48c2e68c676bfa67e0a3cb12.js:4618)
```

## Additional context

I can't nail down what scenario makes this occur every time. Some possibilities:

1. Navigating directly to the conversation
2. Navigating from the conversation list `/conversations`
3. Navigating from a redirect, like when viewing a notification

pkayokay:
I'd love to tackle this one!

pkayokay:
After investigating, I don't think this was actually working? It looks like it's "scrolling" due to the `autofocus: true` on that form. The bug happens when navigating back or forwards in the browser, on a fresh visit it does work (found a related issue https://github.com/hotwired/turbo/issues/354). 

[...]

PR #521, opened 2022-06-24
Fix scroll to bottom on conversation page

pkayokay:
This PR fixes scrolling to the bottom on the individual conversation page (the TypeError)
https://github.com/joemasilotti/railsdevs.com/issues/511

It was not actually "scrolling" but autofocusing on the `text_area` field. This essentially scrolls most of the time but it doesn't ensure that the text area is fully visible all the time.

Note: It is expected that scrolling to the bottom does not happen when going back or forward to the conversation page.

## Pull request checklist

### DIFF
--- a/app/javascript/controllers/page_scroll_controller.js
+++ b/app/javascript/controllers/page_scroll_controller.js
@@ -4,23 +4,29 @@ export default class extends Controller {
   static values = {
     onConnect: { type: Boolean, default: true },
     focus: { type: Boolean, default: true },
+    shouldScroll: { type: Boolean, default: true },
   }
 
   connect() {
     requestAnimationFrame(() => {
-      if (this.onConnectValue) this.scrollIntoView()
-      if (this.focusValue) this.focus()
+      if (this.shouldScrollValue) {
+        if (this.onConnectValue) this.scrollIntoView()
+        if (this.focusValue) this.focus()
+      }
+
+      // Persit scroll position when navigating backwards or forwards
+      window.onpopstate = (e) => {
+        this.shouldScrollValue = false
+      }
     })
   }
-
-  // actions
 
   scrollIntoView() {
     this.element.scrollIntoView(false)
   }
 
   focus() {
-    // focus on the first input , select or textarea in the element
+    // Focus on the first input , select or textarea in the element
     this.element.querySelector("input, select, textarea").focus()
   }
 }
--- a/app/javascript/controllers/page_scroll_controller.js
+++ b/app/javascript/controllers/page_scroll_controller.js
@@ -2,8 +2,8 @@ import { Controller } from "@hotwired/stimulus"
 
 export default class extends Controller {
   connect() {
-    setTimeout(() => {
+    requestAnimationFrame(() => {
       this.element.scrollIntoView(false)
-    }, 0);
+    })
   }
 }
--- a/app/javascript/controllers/page_scroll_controller.js
+++ b/app/javascript/controllers/page_scroll_controller.js
@@ -1,9 +1,26 @@
 import { Controller } from "@hotwired/stimulus"
 
 export default class extends Controller {
+  static values = {
+    onConnect: { type: Boolean, default: true },
+    focus: { type: Boolean, default: true },
+  }
+
   connect() {
     requestAnimationFrame(() => {
-      this.element.scrollIntoView(false)
+      if (this.onConnectValue) this.scrollIntoView()
+      if (this.focusValue) this.focus()
     })
+  }
+
+  // actions
+
+  scrollIntoView() {
+    this.element.scrollIntoView(false)
+  }
+
+  focus() {
+    // focus on the first input , select or textarea in the element
+    this.element.querySelector("input, select, textarea").focus()
   }
 }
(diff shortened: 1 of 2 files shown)


joemasilotti:
Hmm, this doesn't seem to be working for me. When I navigate to a conversation from the conversation index the page doesn't scroll. However, if I refresh the page then it does scroll. I tested in Safari and Chrome on macOS.

pkayokay:
Wanted to give an update

It seems like we were previously adding an event listener for the turbo frame to load `turbo:load`, on the `connect()` lifecycle of the same div's controller. But in this timing, it can never listen for itself? Should it not happen sometime _before_ the actual frame is loaded (before connect)? 

What I have currently works fine on a fresh visit since it's merely running the actualy scroll on `connect()`, but navigating from another page (like the conversation index page you tested) does not possibly due to the [async timing](https://stimulus.hotwired.dev/reference/lifecycle-callbacks#order-and-timing) of Stimulus lifecycle callbacks?

Possible solutions to fix scrolling to bottom from the conversations index page
1. On the conversations index page, wrap the conversations in the page scroll controller, then add an action when a user clicks/pressed key to trigger a scroll. (we'd have to add this in other places where we link conversations within the app)
2. Add a set time out of `500` (maybe less), which should then fix any weird async loading of the frame



Still investigating and tackling this, but if you have any pointers let me know.

joemasilotti:
Interesting! Could we add the controller to `<body>` to hope it gets loaded earlier? Then we could add a `data-target` to the actual element we want to scroll to.

pkayokay:
This is definitely unnecessarily complicated 🤔. 
- [x] ` <body data-controller="page-scroll"...`
- [x] `<div data-page-scroll-target="messagesForm"...`
- [x] Added scroll on _target_ connect lifecycle (`messagesFormTargetConnected`)

https://user-images.githubusercontent.com/13375749/176576250-02ad3236-2896-4d77-b374-d6469e358d13.mov



Does not work. I think what matters is when the frame we're wanting to scroll to mounts

pkayokay:
Added a set timeout with 0 to ensure this actually fires after the element renders.

https://user-images.githubusercontent.com/13375749/176662405-dc2c3f86-8d57-49a7-afce-00f49b8e9023.mov


TIL

"Using setTimeout you create a new task in the bucket after the delay and let the thread deal with it as soon as it´s available for more work."

https://stackoverflow.com/questions/9083594/call-settimeout-without-delay

"...function waits for the call stack to be empty in order to execute the callback"

https://medium.com/@jschapir/settimeout-with-no-delay-the-javascript-event-loop-2e30ed94736a

joemasilotti:
Looking good! What happens when you scroll up then navigate backwards then forwards?

pkayokay:
Oh looks like it's scrolling to the bottom ALL the time now. Are we okay with that or do we want to have it not scroll when going back and forth?

joemasilotti:
Ideally it would only scroll when clicking to the conversation. But if it's really a pain we could ship what you have now.

pkayokay:
Yea..might be more of an additional effort to actually implement that.

With the autofocus on the text area, if autofocus isn't blocked by the browser it will still "scroll" by focusing on the input. We'd need to also not focus when navigating backwards and forwards.

pkayokay:
Nvm, I guess autofocus doesn't really force a scroll..

app/javascript/controllers/page_scroll_controller.js
```diff
@@ -2,14 +2,8 @@ import { Controller } from "@hotwired/stimulus"
...
-
-  disconnect() {
-    document.removeEventListener("turbo:load", this.scrollIntoView)
-  }
-
-  scrollIntoView() {
-    this.element.scrollIntoView(false)
+    setTimeout(() => {
+      this.element.scrollIntoView(false)
+    }, 0);
```
adrienpoly, line 7: Rather than using a setTimeout I would suggest using the next available animation frame

```suggestion
    requestAnimationFrame(() => {
      this.element.scrollIntoView(false)
    })
```

From my testing it does now scroll into view but we are loosing focus on the textarea after submiting a message. For chat message I would expect focus to be retained.

here is what I ended with to maitain focus

```js
import { Controller } from "@hotwired/stimulus"

export default class extends Controller {
  static values = { 
    onConnect: {type: Boolean, default: true},
    focus: {type: Boolean, default: true},
  }                                

  connect() {
    requestAnimationFrame(() => {
      if (this.onConnectValue) this.scrollIntoView()
      if (this.focusValue) this.focus()
    })
  }

  // actions

  scrollIntoView() {
    this.element.scrollIntoView(false)
  }

  focus() {
    // focus on the first input , select or textarea in the element
    this.element.querySelector("input, select, textarea").focus()
  }
}
```

pkayokay: This works well! Great suggestion, didn't think through to test focus after a submission. I will add the rest of the code on the bottom

pkayokay:
https://user-images.githubusercontent.com/13375749/178367603-d834cc5d-a05f-4b0c-9c7c-95169f365691.mov

joemasilotti:
Ah, amazing! Thanks for the continued work on this. I made some code style changes and this is ready to merge.

PR #529, opened 2022-06-29
Self host fonts

### LINKED ISSUE #525: Self-host Inter font
The biggest opportunity according to PageSpeed Insights is to eliminate render-blocking resources.

- [ ] Self-host the [Inter font](https://rsms.me/inter/)

image

adrienpoly:
close #525 

This converts the https://rsms.me link to download the fonts to a self hosted method.

Font files added to the project are downloaded from the [lastest version of Inter ](https://github.com/rsms/inter/releases/tag/v3.19).

I have added all font variations but it looks like the project is using "Inter var" which is a variable font variation (it includes the range of 100 to 900). [Variable fonts is widely supported](https://caniuse.com/variable-fonts).

From my tests adding the "Inter var" leads to a 200kb download whereas downloading individual font weight results in a larger overall weight.

If we are ok to keep only the "Inter var" I can remove all the other files to simplify even further this PR.

## Pull request checklist


joemasilotti:
Lovely! Is there a breakdown of which browsers support the variable weight font somewhere?

adrienpoly:
Here it is 

https://caniuse.com/variable-fonts

adrienpoly:
> Excellent. That's good enough for me to go with the variable font! What changes need to be made to get that working?

Nothing 🎉 
It is what is currently being use on www.railsdevs.com and I picked the same strategy for the self hosting.

I will remove the extra file now that we confirm this route.

will be curious for the Page speed results. I am not yet 100% sure we will remove the warning as it still remains a blocking ressource.

joemasilotti:
> I will remove the extra file now that we confirm this route.

Excellent, thank you!

> will be curious for the Page speed results. I am not yet 100% sure we will remove the warning as it still remains a blocking ressource.

Same. Worse case we can revert the PR.

app/assets/stylesheets/fonts.css
```diff
@@ -0,0 +1,17 @@
+@font-face {
+  font-family: 'Inter var';
```
joemasilotti, line 2: Do these _need_ to be single quoted? If not, let's update this file to only use double quotes.

joemasilotti:
I posted one inline question/comment. Other than that this looks good to merge!

app/assets/stylesheets/fonts.css
adrienpoly: done

joemasilotti:
Awesome – thank you for this!

joemasilotti:
> will be curious for the Page speed results. I am not yet 100% sure we will remove the warning as it still remains a blocking ressource.

95 on mobile and 99 on desktop! Render blocking resources is now passing, too.

adrienpoly:
🎉 very cool 

next up is converting images to webp 😄

joemasilotti:
Yep lol. Any chance you want to take that one? :D

Issue #519, opened 2022-06-23
Serve avatars at 1x and 2x


bradpurchase:
As of #518, we generate variants for avatars @2x the displayed size instead of scaling down the full image. For instance, avatar thumbnails are displayed at 32x32, while the thumbnail itself is 64x64.

As per [this comment](https://github.com/joemasilotti/railsdevs.com/pull/518#issuecomment-1164461961), we want to push this further this by also serving avatars at 1x and 2x via HTML.

## A PR addressing this issue should:

- [ ] Create HTML markup that ensures 1x and 2x images are rendered for the correct screen resolution

I _think_ this might involve using the `srcset` attribute on `<img>` but can't say for sure.

prikeshsavla:
Something like this probably 
https://www.dofactory.com/html/img/srcset

```html
<img srcset="/img/html/vangogh-sm.jpg 120w,
             /img/html/vangogh.jpg 193w,
             /img/html/vangogh-lg.jpg 278w"
     sizes="(max-width: 710px) 120px,
            (max-width: 991px) 193px,
            278px">
```

joemasilotti:
Ooh nice! That will be a fun helper to build. 

From my understanding a good portion of desktop monitors are still 1x. And desktop makes up almost 50% of traffic.

Regardless, PageSpeed checks this. :)

PR #530, opened 2022-06-29
Replace ui-avatars image by real unsplash images

adrienpoly:
As a first step to address #519, I though it would be best to have real avatar image as part of the seed.

This PR replaces the `ui-avatars` images with a random avatar image from **Unsplash**

To avoid adding the unsplash Gem with credientials etc, I fetched 10 random avatar images that are stored within a constant. During the seed we can randomly pick from that constant.

## Preview

Capture d’écran 2022-06-29 à 11 28 52 PM

## Pull request checklist


joemasilotti:
Oooh, I like this idea! Two questions:

1. How much longer does it take to download these images vs. from `ui-avatars`?
2. Is it safe to include "my" Unsplash API key in the public code? I ask because their docs make it sound like it is free forever with no limits.

adrienpoly:
> 1. How much longer does it take to download these images vs. from `ui-avatars`?

I pick the `regular` version of the image it is pretty light weight images. As an example the first one is 100kb.
I didn't noticed any particular slowdown in the `seed` task but I didn't measured it neither.

> 2. Is it safe to include "my" Unsplash API key in the public code? I ask because their docs make it sound like it is free forever with no limits.

When you create a new app on Unsplash it has a demo mode. API credentials for this demo mode are limited to 50 call per hour (largely enough in our case as a seed would require a single call max 2 if we need more than 30 images). 
But it still remains API credentials which are linked to your account and I think should be kept secret. 

> If 2. doesn't make sense then I'd rather add 30+ images and make sure that every developer profile uses a unique one.

Yes that can be easy to do. I can fetch 30 and rather than a random select I can  circle on this list.
Also links can be added to a yaml file rather than a constant if we have 30+ that would look better.


## Credits

officially Unsplash require credits for images to be displayed (link to the author and to unsplash). Given that this is for localhost development, I think we are fine for not doing it.

joemasilotti:
Got it, thanks!

Currently there are 28 developer profiles that get seeded. And 4 businesses. Let's go with a YAML file that has 30 portraits and 10 business-y things. Then, in the seeds, pull from this file in order. If possible, an error stating we ran out of Unsplash images if we go past 30 would be great, too.

And I agree, I don't think we need to worry about attribution for now since these are only for local development.

Long term, I'd like to make seeds less random. And this is a great step in that direction!

adrienpoly:
I have update the PR with now 60 developer urls in a yml file and 30 businesses in another yml file

source for the images are 
- https://unsplash.com/s/photos/avatar
- https://unsplash.com/s/photos/logo-brand

In case they are more than 60 developers then it goes in circle and developer 61 will have again image 1. Do you want a warning for that ?

For the record, here is the script I used (after installing locally the unsplash gem)

```ruby
images = Unsplash::Photo.search("avatar", 1, 30)
urls = images.map{ |i| i.urls.regular }
images = Unsplash::Photo.search("avatar", 2, 30)
urls += images.map{ |i| i.urls.regular }
File.write(File.join(Rails.root, "db", "seeds", "avatars.yml"), urls.to_yaml)

images = Unsplash::Photo.search("logo-brand", 1, 30)
urls = images.map{ |i| i.urls.regular }
File.write(File.join(Rails.root, "db", "seeds", "business_avatars.yml"), urls.to_yaml)
```

db/seeds/avatars.yml
```diff
@@ -0,0 +1,61 @@
+---
+- https://images.unsplash.com/photo-1494790108377-be9c29b29330?crop=entropy&cs=tinysrgb&fit=max&fm=jpg&ixid=MnwzNDIyNzJ8MHwxfHNlYXJjaHwxfHxhdmF0YXJ8ZW58MHx8fHwxNjU3MTQyMTI3&ixlib=rb-1.2.1&q=80&w=1080&utm_source=railsdevs&utm_medium=referral&utm_campaign=api-credit
```
joemasilotti, line 2: Let's only keep the unique photo ID/key in this file and build the full URL in the seeds.

joemasilotti: I'm also not sure if we need all of the params, like `ixid`?

lib/seeds_helper.rb
```diff
@@ -60,15 +60,28 @@ def create_user!(name)
...
       file = uri.open
       record.avatar.attach(io: file, filename: "avatar.png")
     end
 
     def location_seeds
       @location_seeds ||= YAML.load_file(File.join(Rails.root, "db", "seeds", "locations.yml"))
     end
+
+    def developer_avatar_urls
+      @developer_avatar_urls ||= YAML.load_file(File.join(Rails.root, "db", "seeds", "avatars.yml"))
```
joemasilotti, line 80: We could build the full URLs here, or move it to a new file.

joemasilotti (changes requested):
This is awesome – I love how it looks!

One more request - can we shuffle the first 3 into the list? They look very similar and I'd like to spread them out a bit.

db/seeds/avatars.yml
adrienpoly: I will remove those params and only keep the required ones. 

I would suggest keeping the full url rather than just the ID as it makes it much easier for you to check one particular images from the seed. But can do the change if you prefer

joemasilotti: That's a good point. But I don't it's worth the trade-off of having so much additional code in the YAML file.

As a compromise, we could expose the helper to a public class method on the helper so it can be used without much work in the console.

adrienpoly:
here is what the seed now looks like. 
Capture d’écran 2022-07-10 à 9 43 54 PM

Url is now build by a method from an Unsplash ID and I removed some of the strange looking images

joemasilotti:
Ah, perfect! Let's drop that Chucky image (he scares the poop out of me) and update the CHANGELOG and we can merge!

joemasilotti:
Thanks again for this - merging in now! I'm loving how it looks.

PR #531, opened 2022-06-30
run standard rb -fix

adrienpoly:
2 files had StandardRb offenses on main. This PR is just the results of running `standard -fix`

## Pull request checklist


joemasilotti:
Nice catch! I think you might be using a newer version of standard because CI runs the check on every push.

Can you update the gem to the latest so this change makes a bit more sense? Because right now running `bundle exec standardrb --fix` doesn't change anything.

adrienpoly:
indeed I was using a newer version. I added a commit with the latest `standard` version

PR #532, opened 2022-06-30
crowdin config

michelson:
This PR contains the crowding config necessary for the current layout of translations for RailsDevs. it will consider the translations for two letter locales, like `es` `en` `pt` `br`. 

a couple of things will happen when crowdin start to work:

1) It will send Pull requests taking the `en` translation as the cannonical lang.
2) It will rewrite all the other files so you might loose the other translations and it will be needed do the translations again. Luckily you have a small number of translations so I guess this transition will be manageable.


michelson:
Hey Joe, any thoughts on this one?

joemasilotti:
Finally getting back to this one! I set up a new branch (and PR) so crowdin can find the code and push. But it doesn't seem like anything is actually being translated?

Here's my configuration:
image

And the PR: https://github.com/joemasilotti/railsdevs.com/pull/582

michelson:
you will need to do the translations again on the crowding panel, so when the translations are being created the crowding app will start to send PR to the branch. The english "en.yml" version will be set as the base version and from there the new translations will be generated.
¿where is the rails-devs crowdin subdomain?

joemasilotti:
I'm going to close this PR for now. I don't think I have the appetite to set this up right now. Thanks for kicking off the work though!

PR #545, opened 2022-07-08
add a webp image test

adrienpoly:
This is a test to display a webp image and another small step towards #519 where the ultimate goal would be a picture element with retina variations and webp for browsers supporting it.

Since I am not aware of the full Heroku configuration you have, I assume that image processing is performed with libvips under the hood and based upon the build packs it might require an additional APT package to work.

This PR adds a new route /webp that display an avatar image converted to webp. If the image displays correctly in production or staging then all of it can be removed.

There is also a system test to ensure the CI configuration handles webp too.

If you already know that your configuration support webp just ignore this PR


joemasilotti:
Why do we need webp to fix #519?

adrienpoly:
> Why do we need webp to fix #519?

You are right it is not really needed but it is usually an improvement to provide a picture element with both jpg @2x variants as well as webp @2x. 

It is also a follow up of the talks regarding page speed performances https://github.com/joemasilotti/railsdevs.com/pull/529#issuecomment-1171779893

joemasilotti:
Got it. I think for now I'd rather focus on getting 1x/2x images working. Maybe we don't even need to use webp!

adrienpoly:
closing it for now. we can easily reopen it if needed

PR #555, opened 2022-07-10
Add 1x and 2x avatar image variant

### LINKED ISSUE #519: Serve avatars at 1x and 2x

adrienpoly:
close #519

This PR enhance the AvatarComponent to provide both 1x and 2x images. This is a performance improvement, standard monitor will receive the 1x image (32x32) for thumbnail. Whereas Retina monitors will receive the 64x64. Everything is handled by the standard `srcset` html attribute.

I added 2 new variants in the `Avatarable` concern `thumb_2x` and `medium_2x`. From my test locale with a db having only the previous variant, Rails generate automatically the new ones on the first request.

## Pull request checklist

### DIFF
--- a/app/models/concerns/avatarable.rb
+++ b/app/models/concerns/avatarable.rb
@@ -3,8 +3,10 @@ module Avatarable
 
   included do
     has_one_attached :avatar do |attachable|
-      attachable.variant :thumb, resize_to_limit: [64, 64]
-      attachable.variant :medium, resize_to_limit: [256, 256]
+      attachable.variant :thumb, resize_to_limit: [32, 32]
+      attachable.variant :thumb_2x, resize_to_limit: [64, 64]
+      attachable.variant :medium, resize_to_limit: [128, 128]
+      attachable.variant :medium_2x, resize_to_limit: [256, 256]
     end
 
     validates :avatar, content_type: ["image/png", "image/jpeg", "image/jpg"],
--- a/test/components/avatar_component_test.rb
+++ b/test/components/avatar_component_test.rb
@@ -2,6 +2,8 @@ require "test_helper"
 
 class AvatarComponentTest < ViewComponent::TestCase
   include ActionView::Helpers::AssetUrlHelper
+  include ActionView::Helpers::UrlHelper
+  include Rails.application.routes.url_helpers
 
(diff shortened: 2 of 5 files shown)


app/models/concerns/avatarable.rb
```diff
@@ -3,8 +3,10 @@ module Avatarable
 
   included do
     has_one_attached :avatar do |attachable|
-      attachable.variant :thumb, resize_to_limit: [64, 64]
-      attachable.variant :medium, resize_to_limit: [256, 256]
+      attachable.variant :thumb, resize_to_limit: [32, 32]
```
joemasilotti, line 6: Is there any risk in the variant being named the same thing but the size changed... so we would never see the correct variant size?

adrienpoly: from my test using a seeded data from #530 and then adding those variant it is working as expected but a double check could be useful indeed

test/components/avatar_component_test.rb
```diff
@@ -2,6 +2,8 @@
 
 class AvatarComponentTest < ViewComponent::TestCase
   include ActionView::Helpers::AssetUrlHelper
+  include ActionView::Helpers::UrlHelper
+  include Rails.application.routes.url_helpers
```
adrienpoly, line 6: needed those lines ignorer to use `url_for` not sure if someone has a better approach for that

app/models/concerns/avatarable.rb
joemasilotti: Nice. I'm going to run the backfill again on deploy, just in case.

joemasilotti:
Awesome – thanks for this! Going to merge in now.

PR #556, opened 2022-07-10
Fix JS warning : add missing bind(this)

adrienpoly:
In the message view there is a console js warning poping up

Capture d’écran 2022-07-10 à 11 21 22 PM

This PR adds the mising bind(this) to the `eventListener` declaration so that this is defined insisde the `scrollIntoView` method.

> **Note**
This being said I am not sure this Stimulus controller is realy doing what it is meant to do. On page load it will `scrollIntoView` this is I think the expected behavior. But when typing new message the form will disapear from the Viewport. Isn't the expected behavior to keep the form visible after a new message has been posted ?

Capture d’écran 2022-07-10 à 11 33 10 PM

## Pull request checklist

### DIFF
--- a/app/javascript/controllers/page_scroll_controller.js
+++ b/app/javascript/controllers/page_scroll_controller.js
@@ -2,7 +2,7 @@ import { Controller } from "@hotwired/stimulus"
 
 export default class extends Controller {
   connect() {
-    document.addEventListener("turbo:load", this.scrollIntoView)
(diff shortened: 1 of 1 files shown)


joemasilotti:
This is being revamped in #521. If you have any insight on that it would be greatly appreciated!

adrienpoly:
ok yeah I didn't saw that one.

Issue #554, opened 2022-07-10
The `speakerphone` icon is announced to screen reader


metamoni:
## Describe the bug
The `speakerphone` icon in `app/components/developers/new_fields_component.html.erb` and `app/components/developers/invisible_banner_component.html.erb` is not hidden from screen readers. As a result, screen readers announce it as "image" without a description or any additional context. Because this image is purely decorative, it should be hidden from screen readers.

## To reproduce

1. Make sure your developer profile is incomplete
2. Visit the [homepage](https://railsdevs.com/)
3. Turn VoiceOver on: `command + F5`
4. Read the page contents with `control + option + Right Arrow`
5. Notice the speakerphone icon is being announced as "image"
6. Turn VoiceOver off: `command + F5`

## Expected behavior
VoiceOver should skip the speakerphone icon

## Actual behavior
The speakerphone icon is being announced as "image", without a description or any additional context

## Screenshots
https://user-images.githubusercontent.com/22390758/178144191-6a566a36-3330-429f-ba64-821f8f19d0b7.mov

## Additional context
- [Accessibility in inline_svg](https://github.com/jamesmartin/inline_svg#accessibility)
- [More on accessible SVGs](http://www.a11yonrails.com/images#accessible-svgs)

joemasilotti:
Good idea – all yours!

joemasilotti:
Closed by #562.

Issue #551, opened 2022-07-10
Icons in `new` roles view not hidden from screen readers


metamoni:
## Describe the bug
The `briefcase` and `user_group` SVGs are being picked up by VoiceOver but have no accessible titles, so screen readers announce "image" with no other context.

Seeing as these images and decorative and not functional, they should be hidden from screen readers. 

## To reproduce

1. Go to [https://railsdevs.com/role/new](https://railsdevs.com/role/new)
2. Turn VoiceOver on: `command + F5`
3. Press `control + option + Right Arrow` to read the page content
4. Notice that when going over the `briefcase` and `user_group` icons VoiceOver announces "image" with no additional context
5. Turn VoiceOver off: `command + F5`

## Expected behavior
VoiceOver should skip icons entirely

## Actual behavior
VoiceOver announces "image"

## Screenshots
https://user-images.githubusercontent.com/22390758/178141929-707a7ee0-c04b-43e7-a86b-8d19788ca461.mov

## Additional context
Helpful resources:
- [Accessibility in `inline_svg`](https://github.com/jamesmartin/inline_svg#accessibility)
- [More on accessible SVGs](http://www.a11yonrails.com/images#accessible-svgs)

crabbits:
@joemasilotti I'm happy to work on this issue if you wish

joemasilotti:
Sure thing – all yours!

crabbits:
@joemasilotti @metamoni opened a PR here: https://github.com/joemasilotti/railsdevs.com/pull/562, looking for to feedback :)

joemasilotti:
Closed by #562.

PR #562, opened 2022-07-12
Hide icons in `new` roles from screen readers

crabbits:
This PR addresses the issue #551 

I've passed the `aria_hidden: true` option to all `inline_svg_tag` helpers in the `roles/new.html.erb` view. Also added a new integration test to check both options on the `role/new` page and check all svgs in the main DOM element have `aria-hidden=true` set. 

@joemasilotti @metamoni it might be worth me writing a helper to wrap the `inline_svg_tag` helper so `aria_hidden: true` is always passed in by default as I'm sure this is happening on other pages (such as #554)

Also @joemasilotti let me know if you're happy with the approach of passing the option into the `inline_svg_tag` helper vs adding `aria_hidden=true` to the parent `<span>`'s, not sure which one you prefer / will be easier to maintain going forward.


crabbits:
Woops not sure what happened to the CI build, I'll fix it sorry :(

crabbits:
@joemasilotti I also got a `[__NSCFConstantString initialize] may have been in progress in another thread when fork() was called` error while running bin/check, seems to be an [issue](https://github.com/rails/rails/issues/38560) with M1 Macs, would you like me to update the README in the test section to make a note of it?

joemasilotti:
> I've passed the `aria_hidden: true` option to all `inline_svg_tag` helpers in the `roles/new.html.erb` view. Also added a new integration test to check both options on the `role/new` page and check all svgs in the main DOM element have `aria-hidden=true` set. 

No need for the integration test! Feel free to remove that.

> @joemasilotti @metamoni it might be worth me writing a helper to wrap the `inline_svg_tag` helper so `aria_hidden: true` is always passed in by default as I'm sure this is happening on other pages (such as #554)

Great idea. Let's add a helper that does exactly that. `aria_hidden_inline_svg_tag` or `non_interactive_inline_svg_tag`, maybe?

Even better... `icon_tag` and `decorative_icon_tag` could wrap the exiting helper and clean up the repeated filename string a bit!

```ruby
<%= decorative_icon_tag "briefcase", :outline, class: "..." %>
<%= icon_tag "mail", :solid, class: "..." %>
```

> Also @joemasilotti let me know if you're happy with the approach of passing the option into the `inline_svg_tag` helper vs adding `aria_hidden=true` to the parent `<span>`'s, not sure which one you prefer / will be easier to maintain going forward.

If we go with the helper above that will always set it on the SVG directly. I'm inclined to go with that since we won't ever have to set it manually by using the two different helpers.

crabbits:
> No need for the integration test! Feel free to remove that.

Easy will do

> Great idea. Let's add a helper that does exactly that. aria_hidden_inline_svg_tag or non_interactive_inline_svg_tag, maybe?
>
> Even better... icon_tag and decorative_icon_tag could wrap the exiting helper and clean up the repeated filename string a bit!

I like this approach, will play around with it once I'm off work.

> What version of Ruby are you running for that to appear? The thread hints at this being fixed in a recent release.

I'm running `3.1.2`, but it may be something weird with my setup, will leave it for now.

joemasilotti:
@metamoni, would love to hear your expertise here. Is there a difference (or recommended approach) for these two options?

```html
<span class="text-blue-500" aria-hidden="true">
  <svg src="/icons/mail.svg"></svg>
</span>
```

```html
<span class="text-blue-500">
  <svg src="/icons/mail.svg" aria-hidden="true"></svg>
</span>
```

joemasilotti:
Hey @crabbits, apologies for overwriting some of your changes with my recent push. But I realized that we don't need anything fancy like what I was prescribing earlier. I opted to add a single option to each `inline_svg_tag` to make them hidden to screen readers.

Thanks for kicking off the work on this!

benmercerdev:
> @joemasilotti I also got a `[__NSCFConstantString initialize] may have been in progress in another thread when fork() was called` error while running bin/check, seems to be an [issue](https://github.com/rails/rails/issues/38560) with M1 Macs, would you like me to update the README in the test section to make a note of it?

Were you able to resolve this? Also running into the same issue. M1 Mac here as well

PR #575, opened 2022-07-14
remove handcoded json

### LINKED ISSUE #567: Remove hardcoded JSON string in fixture
During the Rails 7.0.3.1 upgrade I had to hardcode a JSON string into a test fixture. Ideally, this can be changed to a better representation (like a hash or more YAML) but I couldn't figure it out.

- [] Replace the hardcoded JSON string in the `OpenStartup::Metric` fixture

`test/fixtures/open_startup/metrics.yml:3`

gathuku:
## Pull request checklist

closes #567

When looking at the [`store` documentation](https://api.rubyonrails.org/classes/ActiveRecord/Store.html) i came across this section.

>> NOTE: If you are using structured database data types (e.g. PostgreSQL hstore/json, or MySQL 5.7+ json) there is no need for the serialization provided by [.store](https://api.rubyonrails.org/classes/ActiveRecord/Store/ClassMethods.html#method-i-store). Simply use [.store_accessor](https://api.rubyonrails.org/classes/ActiveRecord/Store/ClassMethods.html#method-i-store_accessor) instead to generate the accessor methods. Be aware that these columns use a string keyed hash and do not allow access using a symbol.

Not sure if i got this right? so i think we can avoid custom coder and use `store_accessor` 🤔

### DIFF
--- a/app/models/open_startup/metric.rb
+++ b/app/models/open_startup/metric.rb
@@ -2,7 +2,7 @@ module OpenStartup
   class Metric < ApplicationRecord
     self.table_name = "open_startup_metrics"
 
-    store :data, accessors: %i[mrr visitors], coder: JSON
+    store_accessor :data, :mrr, :visitors
 
     validates :occurred_on, presence: true
 
--- a/test/fixtures/open_startup/metrics.yml
+++ b/test/fixtures/open_startup/metrics.yml
@@ -1,3 +1,5 @@
 one:
   occurred_on: 2022-01-01
-  data: "{\"mrr\": 1000, \"visitors\": 5000}"
+  data:
+     mrr: 1000
+     visitors: 5000


joemasilotti:
Excellent and TIL! Thanks for your contribution.

Were you able to verify that existing data is still accessible after changing the code?

gathuku:
Am not sure if there's some seeds to metrics table. 
I was using test to load the data. 

```sh
➜  railsdevs.com git:(main) ✗ RAILS_ENV=test rails db:fixtures:load
➜  railsdevs.com git:(main) ✗ RAILS_ENV=test rails c               
Loading test environment (Rails 7.0.3.1)
[1] pry(main)> OpenStartup::Metric.first
  OpenStartup::Metric Load (0.5ms)  SELECT "open_startup_metrics".* FROM "open_startup_metrics" ORDER BY "open_startup_metrics"."id" ASC LIMIT $1  [["LIMIT", 1]]
=> #<OpenStartup::Metric:0x000000010bcfda30
 id: 980190962,
 occurred_on: Sat, 01 Jan 2022,
 legacy_data: {},
 created_at: Thu, 14 Jul 2022 17:29:01.782149000 UTC +00:00,
 updated_at: Thu, 14 Jul 2022 17:29:01.782149000 UTC +00:00,
 data: {"mrr"=>1000, "visitors"=>5000}>
[2] pry(main)> OpenStartup::Metric.first.mrr
  OpenStartup::Metric Load (1.2ms)  SELECT "open_startup_metrics".* FROM "open_startup_metrics" ORDER BY "open_startup_metrics"."id" ASC LIMIT $1  [["LIMIT", 1]]
=> 1000
[3] pry(main)> OpenStartup::Metric.first.visitors
  OpenStartup::Metric Load (2.2ms)  SELECT "open_startup_metrics".* FROM "open_startup_metrics" ORDER BY "open_startup_metrics"."id" ASC LIMIT $1  [["LIMIT", 1]]
=> 5000
```

joemasilotti:
Awesome, thanks! Merging now.

gathuku:
Thanks, that was smooth! 😄

joemasilotti:
Oh boy, looks like I jumped the gun. [I don't think that converted correctly!](https://railsdevs.com/open) I'm going to revert for now until we can figure out what is going on.

image

gathuku:
That's sad. can you able to see how the current data looks like in staging/production, maybe i can try reproducing locally.

joemasilotti:
I think checkout `main`, run the following code in the console, then make your changes.

Rerunning the second line should reproduce your issue.

```ruby
OpenStartup::Metric.create!(mrr: 200, occurred_on: Date.today)
OpenStartup::Metric.last.mrr # => 200
```

gathuku:
Just seen it, it's no compatible with current data

gathuku:
@joemasilotti i think the only option we have to use this code is to covert the data back to hash instead of JSON string.

joemasilotti:
The test or the code in the model?

gathuku:
The actual data in the production/staging database .

Something like
```
OpenStartup::Metric.find_each { |metric| metric.update(data: JSON.parse(metric.data)) }
```

Currently we have a json string stored
```
  data: "{\"mrr\":200}"

 ```
The goal is to endup with a hash like 
```
  data: {"mrr"=>400}
```

**Comment by joemasilotti** [comment]
That feels like it goes against the purpose of the store/accessor, no?

**Comment by gathuku** [comment]
I think it goes against the purpose of the `coder`, the coder is the one that converts it's to a string. i think the default json is stored as  a hash.

**Comment by joemasilotti** [comment]
I'm going to close this for now as I don't think it's worth the trouble to remove from tests. If we add more JSON then we can revisit. Thanks for diving in, @gathuku!

**Comment by gathuku** [comment]
@joemasilotti  if you would like to bring this back, i think this will be an elegant way to represent the json string data


```haml
one:
  occurred_on: 2022-01-01
  data: |
    {
     "mrr": 1000,
     "visitors": 5000
    }
```

joemasilotti:
Oooh, that does look much cleaner! Can you submit a PR to make sure the tests and data migration still work?

PR #579, opened 2022-07-17
Indicate conversations that have unread messages

benngarcia:
## Pull request checklist

Aims to close #570. (Goofed the PR). 

- [✅  ] My code contains tests covering the code I modified


benngarcia:
@joemasilotti Since we need to evaluate each conversation in the context of whether or not the `current_user` is the business or developer for that conversation, this is a bit beefier than I though originally. Will take a stab, but not expecting to get it right the first time :)

app/controllers/conversations_controller.rb
```diff
@@ -2,7 +2,7 @@ class ConversationsController < ApplicationController
   before_action :authenticate_user!
 
   def index
-    @conversations = current_user.conversations.order(updated_at: :desc)
+    @conversations = current_user.conversations.includes(:business).order(updated_at: :desc)
```
joemasilotti, line 5: Do we still need this if we are using the IDs?

app/models/conversation.rb
```diff
@@ -48,11 +49,11 @@ def latest_message
 
   def mark_notifications_as_read(user)
     notifications_as_conversation.where(recipient: user).unread.mark_as_read!
+    all_messages_read!
```
joemasilotti, line 52: Do we need to only mark the user's messages as read? If a a developer sends a message then opens the conversation again this code will mark it read even for the business.

benngarcia: Ah, great catch. I'll add some logic for that as well.

app/controllers/conversations_controller.rb
benngarcia: Unfortunately we do. Otherwise the `business?` method in the  `Conversation` model would shoot off n+1 queries. This should effectively make it so that computing every conversations `read_status` will take 2 queries. Unless I'm off with that 🤔

benngarcia:
Sorry for the delay. Got a littttttle busy :)

joemasilotti:
All good – thanks for following up! I'm excited to see more progress on this.

app/controllers/conversations_controller.rb
joemasilotti: Hmm. Since the notifications are scoped to a user (not a business or developer) what if we track the read status based on the user?

benngarcia: That was my first thought as well. Honestly have no idea why I didn't go with that approach first time around, lol. I think it was 4am when I took a first stab? lol.

Something akin to
```
current_user.conversations.order(updated_at: :desc).select(:'conversations.*', 
'EXISTS(SELECT 1 FROM notifications WHERE params @> conversations AND conversations.recipient_type = 'User' AND conversations.recipient_id = #{current_user.id} AND read_at IS NULL) AS unread_messages`
```

Syntax is wrong but that's the general idea. I think 4am me forgot about extending select? Or is this an antipattern, lol.

Either way, I would translate this to an active query type command. Just wanted to get your thoughts before re-digging in.

joemasilotti: I'm finding that SQL statement pretty hard to read, but if we can convert it to Active Record or Arel then that sounds like a better approach to me!

benngarcia: Ah, a bit SOL. I remember the problem now. We needs rails to serialize the object to query by it, which isn't supported OOTB by noticed, because it really isn't supported OOTB by rails as far as I can tell (happy to be proved wrong!). That's bc even though it's scoped to the user, the notifications params is where we need to find the relevant conversation to know whether or not to check the `read_at` column. 

Trying something, fingers crossed! If I can make it work I'll be busy submitting a PR for noticed gem :)

benngarcia: The reason for this is because the params are serialized with the `ActiveJob` serializer. Great for the DB, was never really built to be query-able.

benngarcia: Have a solution up, take a look and let me know what you have qualms with. Sort of had to force it to make it work with two queries, but it's now scoped to the user.

joemasilotti: Sorry, but I feel like I'm missing something...

If we are caching the unread status via `conversations.read_status` then why do we need a complex query? You've already added the callback when sending a message to update the conversation.

Why can't we iterate over conversations and then:

```erb
<% conversations.each do |conversation| %>
  <h2>Conversation with <%= conversation.other_recipient(current_user).name %></h2>
  <% if conversation.business?(current_user) && conversation. messages_for_business_unread? || conversation.developer?(current_user) && conversation.messages_for_developer_unread? %>
    <p>Unread messages</p>
  <% end %
<% end %>
```

benngarcia: What a d'oh moment! Way more straightforward.

benngarcia:
@joemasilotti Should be good to go now 😅, hopefully we can just sweep this whole pr under the rug and pretend it didn't happen... But on the other hand I really just never thought about using an association as a cache. Not sure where the blocker was in my head but it felt illegal. 

Learned something new!

P.s. I added a little unique index to match the uniqueness constraint. Should speed things up a millisecond ;)

joemasilotti:
This is looking great! Can you please address the CI failure and add a backfill for existing conversations? When those are done I'll do a more in-depth code review.

benngarcia:
Can do! Do you have any preferences for backfilling?

joemasilotti:
There's a rake file called `backfills.rake` or something. Please put your backfill there and make sure it is idempotent.

benngarcia:
Is that good enough? Wasn't sure how else to make it idempotent :p

db/schema.rb
```diff
@@ -10,8 +10,9 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.0].define(version: 2022_07_12_221754) do
+ActiveRecord::Schema[7.0].define(version: 2022_08_06_091758) do
   # These are extensions that must be enabled in order to support this database
+  enable_extension "pg_stat_statements"
```
joemasilotti, line 15: Can you remove this line?

db/schema.rb
```diff
@@ -249,7 +252,7 @@
     t.string "processor_id"
     t.boolean "default"
     t.jsonb "data"
-    t.datetime "deleted_at"
+    t.datetime "deleted_at", precision: nil
```
joemasilotti, line 255: Please remove this change.

db/schema.rb
```diff
@@ -286,8 +289,8 @@
     t.string "processor_plan", null: false
     t.integer "quantity", default: 1, null: false
     t.string "status", null: false
-    t.datetime "trial_ends_at"
-    t.datetime "ends_at"
+    t.datetime "trial_ends_at", precision: nil
+    t.datetime "ends_at", precision: nil
```
joemasilotti, line 293: Please remove this change as well.

lib/tasks/backfills.rake
```diff
@@ -1,3 +1,13 @@
 desc "These tasks are meant to be run once then removed"
 namespace :backfills do
+  task backfill_conversation_unread_status: :environment do
+    # As long as there is one unread message, don't run. Hopefully idempotent _enough_.
+    return if Conversation.where.not(user_with_unread_messages: nil).exists?
```
joemasilotti, line 5: Line 6 makes this idempotent enough for me. Please remove line 4 and 5.

test/models/conversation_test.rb
```diff
@@ -108,6 +108,22 @@ class ConversationTest < ActiveSupport::TestCase
     assert conversation.latest_message_read_by_other_recipient?(sender)
   end
 
+  test "unread_messages_for? returns false if there are no unread notifications for current_user" do
+    conversation = conversations(:one)
+    user = users(:business)
+    conversation.latest_message.notifications_as_message.destroy_all
```
joemasilotti, line 114:
```suggestion
    conversation.latest_message.notifications_as_message.mark_as_read!
```

lib/tasks/backfills.rake
```diff
@@ -1,3 +1,13 @@
...
+    Conversation.where(user_with_unread_messages: nil).includes(:messages).each do |conversation|
+      last_notification = conversation.messages.last.notifications_as_message.last
+      if last_notification.unread?
```
joemasilotti, line 8:
```suggestion
      if last_notification&.unread?
```

joemasilotti (changes requested):
Looking good! I left a few minor comments/suggestions. After that we should be good to merge.

benngarcia:
Should be all good to go 👍

joemasilotti:
LGTM! Thanks for your contribution. I appreciate the work as we went back a forth a few times.

Issue #539, opened 2022-07-04
Incomplete aria-expanded in query_sort_button_component.html.erb


metamoni:
## Describe the bug

The "Sort" button on the Developers list page uses `aria-expanded="false"` to tell screen readers when the sort menu is collapsed, but is not set up to change to `true` when the menu is expanded.

`query_sort_button_component.html.erb`

## To reproduce (on a Mac)

1. Sign in as a developer
2. Go to http://localhost:3000/developers
3. Turn on VoiceOver: `command + F5`
4. Tab to "Sort" and notice the screen reader announces the menu is "collapsed"
5. Hit `Enter` to open the menu and notice the screen reader announces "menu (2 items)"
6. Turn VoiceOver off: `command + F5`

## Expected behavior
When opening the "Sort", the screen reader should announce that it's "expanded"
When closing the "Sort" menu, the screen reader should announce that it's "collapsed"

See example of correct screen reader behaviour below:

https://user-images.githubusercontent.com/22390758/177088249-951fce1a-7786-41af-abc6-b67fc7462feb.mov

## Actual behavior
Expanded/collapsed state isn't announced when the "Sort" menu is open

Current screen reader behavior

https://user-images.githubusercontent.com/22390758/177088192-59640620-7c5d-4acb-8c1e-0f66c0cde8d4.mov

## Additional context
Example of using `aria-expanded` properly with an accordion component: [w3c.github.io](https://w3c.github.io/aria-practices/examples/accordion/accordion.html)

axhgl:
Could I give this one a try, @joemasilotti? Thank you @metamoni for a well described issue along with the additional context.

joemasilotti:
Sure thing - all yours!

joemasilotti:
Closed via #585.

PR #580, opened 2022-07-17
Toggle 'Sort' button's aria-expanded attribute value.

axhgl:
This should address issue #539.

Small background regarding the thinking for this implementation:

We add a Stimulus.js `AccessibilityController` where we define actions that handle the updating of values of accessibility related attributes, e.g. in this case it is `#toggleAriaExpanded` to handle "aria-expanded", while in the future it might be a case where we add `#toggleAriaChecked` for handling the "aria-checked" attribute.

I briefly wondered: should there be an integration test covering this?

## Pull request checklist

### DIFF
--- a/app/javascript/controllers/accessibility_controller.js
+++ b/app/javascript/controllers/accessibility_controller.js
@@ -1,0 +1,13 @@
+import { Controller } from "@hotwired/stimulus"
+
+export default class extends Controller {
+  static targets = ["button"]
+
+  toggleAriaExpanded() {
+    const element = this.buttonTarget;
+    const initialVal = element.getAttribute("aria-expanded")
+    const toggledVal = initialVal === "true" ? "false" : "true"
+
+    element.setAttribute("aria-expanded", toggledVal)
+  }
+}
--- a/app/javascript/controllers/index.js
+++ b/app/javascript/controllers/index.js
@@ -3,6 +3,9 @@
 
 import { application } from "./application"
 
+import AccessibilityController from "./accessibility_controller.js"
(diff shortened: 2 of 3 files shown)


app/javascript/controllers/index.js
```diff
@@ -29,3 +29,6 @@ application.register("turbo-native--in-app-purchases", TurboNative__InAppPurchas
 
 import TurboNative__SignOutController from "./turbo_native/sign_out_controller.js"
 application.register("turbo-native--sign-out", TurboNative__SignOutController)
+
+import AccessibilityController from "./accessibility_controller.js"
+application.register("accessibility", AccessibilityController)
```
joemasilotti, line 34: Can you run `bin/rails stimulus:manifest:update` so this gets sorted correctly?

app/javascript/controllers/accessibility_controller.js
```diff
@@ -0,0 +1,13 @@
+import { Controller } from "@hotwired/stimulus"
+
+export default class extends Controller {
+  static targets = ["button"]
+
+  toggleAriaExpanded() {
+    const el = this.buttonTarget;
```
joemasilotti, line 7:
```suggestion
    const element = this.buttonTarget;
```

joemasilotti (changes requested):
Looks good to me. I made two minor suggestions – once those are taken care of I can merge this in. Thanks for the contribution!

axhgl:
It should be alright now. By the way, I squashed the couple small fixing commits.

joemasilotti:
Looks great! Thanks for the update.

Issue #587, opened 2022-07-28
Refactor to use `class_names` helper


joemasilotti:
I recently learned about the [`class_names`](https://twitter.com/_swanson/status/1341080006169137152) helper from @swanson's [blog post on Hotwire-related CSS tricks](https://boringrails.com/articles/css-tips-and-tricks-hotwire/).

I'd love to leverage this to remove all string interpolation of class names across the codebase.

## A PR addressing this issue should:

- [ ] Remove all stripe interpolated class name logic
- [ ] Replace said logic with the `class_names` helper
- [ ] See if there are any additional abstractions we can pull out.

joshio1:
Hey @joemasilotti, can I take a look at this one?

joemasilotti:
Sure thing - all yours!

joshio1:
Hi @joemasilotti, while I' looking into this one, I'm not 100% sure which one you meant when you said "stripe interpolated class name logic".
I tried searching for classes in the `app/views` but couldn't find any interpolation for class names. Inside the `app/components` folder, there were a few places which had class name interpolation but I'm not sure if those were the ones you meant. These are the files I found which had class name interpolation:
1. toast_message_component.html.erb#13
2. avatar_component.html.erb#5
3. cover_image_component.html.erb#2
4. sort_button_component.html.erb#1
5. submit_button_component.html.erb#1
6. search_status_component.html.erb#4

joemasilotti:
Yep! Those are exactly what I'm talking about. I think a perfect example is `SortButtonComponent#dynamic_classes` - ideally those classes live in the HTML, not the Ruby file. So `app/components/sort_button_component.html.erb:1` could look like:

```erb
<%= button_tag title, form: form_id, name: name, value: value, type: :submit, class: class_names: "block w-full text-left px-4 py-2 text-sm cursor-pointer hover:bg-gray-100", "font-medium text-gray-900": active?, "font-normal text-gray-700": !active? %>
```

This wasn't checked, but I think we can pass multiple class names in like that. Let me know if that doesn't make sense (or doesn't work!).

joshio1:
Hey @joemasilotti, so I've raised this PR: https://github.com/joemasilotti/railsdevs.com/pull/593

[...]

PR #593, opened 2022-08-02
Use class names helper instead of string interpolation

joshio1:
Addresses https://github.com/joemasilotti/railsdevs.com/issues/587

## Pull request checklist


joemasilotti:
This looks great! Replying to your question in the issue: no, I don't think we need to add tests. Anything that already exist should be good enough.

Here are some more files that have CSS classes in the component's Ruby file:

* `app/components/collapse_control_component.rb:11`
* `app/components/avatar_component.rb:24`

These two files deal with CSS classes around iconography:

* `app/components/developers/availability_component.rb:16`
* `app/components/developers/search_status_component.rb:36`

I'm also now inclined to remove `yass` now that I know about `class_names`.

* `app/components/toast_message_component.rb:16`
* `app/components/toast_message_component/toast_message_component.html.erb:1`
* `app/helpers/form_helper.rb:4`

---

Can you take a look at the first two? If you have any ideas for the icon-related ones I'd love to hear it, too. But no worries if not. I'll take care of `yass` in a new issue/PR.

joshio1:
So, I was able to use the `class_names` helper for `collapse_control_component.rb:11`

For `avatar_component.rb:24`, the classes used are either passed as an argument to the component or we use a static set of classes. I am not sure how we can avoid interpolation in this case. 

For the two css files that deal with iconography (`availability_component.rb:16` and `search_status_component.rb:36`), I saw that we are using `inline_svg_tag` which comes from the `inline_svg` gem. I tried using the new style of passing class names for this tag but it didn't work as expected and I also couldn't find a way to do it in the gem's documentation. So, I'm not sure how we'd do it. I'd love to hear your thoughts.

joemasilotti:
Ah, good catch!

I moved the CSS classes to the HTML files at the expense of a few more if/else statements. This feels like a solid tradeoff for now as I prefer the styling elements in HTML, not Ruby files.

Thanks for your contribution! I'm going to merge this in once CI passes.

PR #594, opened 2022-08-03
WIP: Reasons for invisibilizing developers

joemasilotti:
This PR presents two options when an admin marks a developer profile as invisible. Their profile can be too short/spammy or they aren't responding to messages from businesses.

This came directly from a business complaining that two folks never responded to them. Turns out, neither of them have responded to _anyone_!

## Pull request checklist


joemasilotti:
I don't think this is worth the added complexity. Worse case I can manually mark a developer as invisible and send them an email directly.

PR #599, opened 2022-08-12
Move content pages into Sitepress

bradgessler:
This PR replaces controllers per content page with [Sitepress](https://sitepress.cc/), a lightweight content management gem that can be embedded in Rails. Opening a draft PR to see if you want to keep going down this path. If you don't, no worries! I'll close this PR.

Here's the benefits of Sitepress:

1.   To add a new content page, just throw it to `./app/content/foo/bar/blah.html.md` and it will be available at `/foo/bar/blah`. You don't have to create a controller like `AboutController`, hook up the routes, etc.

2.
    In your nav, footers, etc. you can loop through the pages in content and link to them. That might look like this: 
    
    ```ruby
    <ul>
    <% site.glob("*.html.*").each do |page| %>
      <li><%= link_to_page page %></li>
    <% end %>
    </ul>
    ```

    Then, when you add a content page in the future, it will automatically populate the nav bar.

3.   It's lightweight. No DB, etc. that you typically associate with a CMS. It's all just backed by boring old files, and tracked with git, which bills itself as "the stupid content tracker"

## Pull request checklist

<!-- Before you submit a pull request for review, please make sure... →

<!-- If this PR is not ready for review, please make sure to submit it as a draft. →


joemasilotti:
Hey Brad, thanks for this PR and letting me know about Sitepress. Just so I understand 100%, adding this gem will get rid of the empty controller and routes, right? Nothing else?

bradgessler:
Correct! You'd get to delete a bunch of controllers and routes ... so your LOC goes down quite a bit.

The upside is that its MUCH easier to add content. For example, if you wanted to put content at `/case-studies/acme` you'd create a file at `./app/content/case-studies/acme.html.erb` and it all just works. No need to create controllers, connect a route, etc.

joemasilotti:
Got it – cool stuff! Is the gem production ready? I don't see any open source projects relying on it and the build looks to be currently broken.

bradgessler:
Eugh yeah, my test runner on GitHub fails intermittently on the matrix of tests that runs diff versions of Ruby against diff versions of Rails. Something isn't tearing down the Rails env properly in the runner. Check it now and you'll see it's green.

Is it production ready? Yes! I've been running it for years at https://www.polleverywhere.com/, which is the largest deployment I know of. It's also running at https://www.thingybase.com/ and https://legiblenews.com/ dynamically inside of Rails apps. All of the above have survived HN traffic surges without breaking a sweat... and without any caching.

joemasilotti:
Got it, thanks Brad!

I think for now I'm not interested in bringing in this gem. Even though it's not a huge dependency, there are only a handful of static pages right now. And it doesn't feel worth it, in my opinion, to add another dependency to remove a few routes and empty controllers.

I'm going to close this PR. However, I'd love to explore this again if/when I add more static content to the site. Thanks for taking the time to integrate it into RailsDevs.

PR #611, opened 2022-08-24
Use correct spelling for stationery as in office supplies

brettskiii:
Stationary = not moving
Stationery = office supplies


joemasilotti:
Great catch – thanks for this!

PR #638, opened 2022-09-15
Allow rich text for Developer#bio

### LINKED ISSUE #633: Rich text for developer bio
The designer I'm working with noticed that multiple profiles are writing quasi-formatted lists and sections in their bio. I'd love to offer this as a real feature so folks can clean up their profiles a bit.

- [ ] Enable markdown in the developer bio field

When rendering, _only_ the following tags/elements should be displayed. Everything else should be stripped, removed, or sanitized.

* italics
* bold
* strikethrough
* ordered lists
* unordered lists
* headings
* code blocks
* inline code blocks

RailsDevs already has [Redcarpet](https://github.com/vmg/redcarpet), so ideally we can use that gem. Digging into the README it looks like there are options for disabling everything we don't want (images, styles, links, etc.).

rayhanw:
Allows markdown for `developer#bio`. Closes #633 

## Pull request checklist

### Notes:
- Redcarpet markup can be found [here](https://github.com/rayhanw/railsdevs.com/blob/207791d92e1c57c18a704c029e1bc3120fcccc24/app/models/concerns/developers/rich_text.rb#L9-L23).
- ~~I created a [custom renderer](https://github.com/rayhanw/railsdevs.com/blob/207791d92e1c57c18a704c029e1bc3120fcccc24/app/models/developer_bio_render.rb) object to handle links and images.
  If the user writes a markdown link/image, it will be saved into the DB but rendered as an empty string (check the custom renderer object above). I'm not sure if this is the best way, or whether actually removing it before the `bio` is saved would be better. Let me know!~~
- ~~I didn't write tests for the custom renderer object because I'm honestly not sure what to test of it and it should also be covered by the `developers/rich_text` concern tests.~~

### Examples:
Text:
```markdown
**Lorem ipsum dolor sit amet.**
```
Result:
```html
<strong>Lorem ipsum dolor sit amet.</strong>
```
---
Text:
```markdown
[link](https://example.com)
![image](https://example.com)
```
Result:
```html
[link](https://example.com)
![image](https://example.com)
```


joemasilotti:
Nice work so far!

I'd prefer to keep all of the user's inputted text for the bio instead of deleting anything. Can we instead render the markup directly, as in not an HTML `<a>` tag?

I think this might be possible with `filter_html` + `no_images` + `no_links` options all set to `true` via `Redcarpet::Render::HTML.new(render_options = {})`.

test/models/concerns/developers/rich_text_test.rb
```diff
@@ -0,0 +1,48 @@
...
+  end
+
+  test "ignores links" do
+    developer = developers(:one)
+    bio = <<~MD
+      [Text](https://example.com)
+    MD
+    developer.update!(bio:)
+
+    assert_match developer.rich_text_bio, ""
```
joemasilotti, line 36: As per my comment, I think I want this test to read:

```suggestion
    assert_match developer.rich_text_bio, "[Text][https://example.com)"
```

And then add another to ensure `<a href="https://example.com">Text</a>` gets rendered as text, not an `<a>` tag. `filter_html: true` should handle that, IIRC.

rayhanw:
With the chosen options for the renderer and markdown, this is what it looks like now:
image

Might I also suggest adding in a text in the bio section in the form that says something along the lines of "Markdown is supported".

Should now be ready for review 👍

joemasilotti:
Excellent! The preview you rendered looks perfect.

Let's add the "Markdown supported" text – great idea. Please use the following copy (it can be added below the EXAMPLE TOPICS list):

> Markdown is supported but links and images are removed.

rayhanw:
It is now just missing non-English translations 🙏

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ## September
 
+* September 15 - Allow rich text for Develoepr#bio #638 @rayhanw
```
```suggestion
* September 15 - Allow rich text for develoepr bio #638 @rayhanw
```

app/models/concerns/developers/rich_text.rb
```diff
@@ -0,0 +1,33 @@
...
+      superscript: true, # Disables convertion of text with four spaces at the front of each line to code blocks.
+      underline: true # Allows underline
+    }.freeze
+    RENDER_OPTIONS = {
+      filter_html: true, # Disallows manually written HTML tags
+      hard_wrap: true, # Adds newline everytime "Return" is hit
+      no_images: true, # Disables images
+      no_links: true, # Disables links
+      no_styles: true # Disables custom styling
+    }.freeze
```
joemasilotti, line 23: Instead of the comments lets link to the Redcarpet documentation for each constant as one comment, each. For example,

```ruby
# https://github.com/vmg/redcarpet#darling-i-packed-you-a-couple-renderers-for-lunch
RENDER_OPTIONS = {
```

joemasilotti (changes requested):
Looks and works great! I only have two tiny requests.

joemasilotti:
Oh, I just noticed one more thing. Developer cards are rendering the plain markdown. Let's update that to remove the markup tags and render plain text. For example:

`**hi** _bye_` should render as:

* **hi** _bye_ – on the developer profile page
* hi bye – in a developer card

rayhanw:
Looks like this now on the developer cards (with Redcarpet's stripdown renderer)
Screen Shot 2022-09-16 at 07 10 29

joemasilotti:
This looks great! I made a few tiny tweaks and will merge after CI passes.

Thanks for your contribution, @rayhanw.

Issue #640, opened 2022-09-15
Recently active badge


joemasilotti:
This kicks off the first batch of work for [developer discoverability](https://github.com/joemasilotti/railsdevs.com/discussions/639)!

The goal of this feature is to highlight developers who have been recently active on the platform. "Active" is not limited to only updating a profile, but also responding to messages.

A PR addressing this issue should:

- [ ] Add a "Recently active" badge to developer cards if they sent a message or updated their profile in the last 7 days

The badge can mirror the design of the existing "Featured developer" badge with the following colors: `bg-green-100 text-green-800`. It might make sense to extract a new View Component for badges (if it doesn't already exist).

Implementation-wise, we don't want to query for a developer's last sent message as this could lead to N+1 situation. Instead, let's make sure that sending a message touches `developers.updated_at`. This might be as straightforward as adding `touch: true` to `app/models/message.rb:11` but I need to experiment.

Alternatively, we could create a new column and manage that directly. But it will require more work to maintain and I'm not inclined to add additional complexity if it isn't needed.

sarvaiyanidhi:
I would like to work on this feature.. Can you please assign it to me?

joemasilotti:
Sure thing @sarvaiyanidhi! All yours. Let me know if you have any questions when diving in.

sarvaiyanidhi:
Sure. Will reach out in case of any query. Thanks

joemasilotti:
Closed via #649.

PR #649, opened 2022-09-16
Feature to show recently active badge for developers

sarvaiyanidhi:
This PR handles code to display recently active badge for developers who has updated profile or sent message in last 7 days. More details in issue #640 

## Pull request checklist

### DIFF
--- a/app/components/developers/badge_component.rb
+++ b/app/components/developers/badge_component.rb
@@ -1,0 +1,14 @@
+module Developers
+  class BadgeComponent < ApplicationComponent
+    private attr_reader :title, :classes
+
+    def initialize(title, classes)
+      @title = title
+      @classes = classes
+    end
+
+    def call
+      tag.span title, class: classes.to_s
+    end
+  end
+end
--- a/app/components/developers/badge_component.rb
+++ b/app/components/developers/badge_component.rb
@@ -1,14 +1,14 @@
 module Developers
   class BadgeComponent < ApplicationComponent
-    private attr_reader :title, :classes
+    private attr_reader :title, :color
 
-    def initialize(title, classes)
+    def initialize(title, color:)
       @title = title
-      @classes = classes
+      @color = color
     end
 
     def call
-      tag.span title, class: classes.to_s
+      tag.span title, class: "inline-flex items-center rounded-md px-2.5 py-0.5 text-sm font-medium bg-#{@color}-100 text-#{@color}-800"
     end
   end
 end
--- a/app/components/developers/badge_component.rb
+++ b/app/components/developers/badge_component.rb
@@ -8,7 +8,7 @@ module Developers
     end
 
     def call
-      tag.span title, class: "inline-flex items-center rounded-md px-2.5 py-0.5 text-sm font-medium bg-#{@color}-100 text-#{@color}-800"
+      tag.span title, class: "inline-flex items-center rounded-md px-2.5 py-0.5 text-sm font-medium bg-#{color}-100 text-#{color}-800"
     end
   end
 end
--- a/app/components/developers/badge_component.rb
+++ b/app/components/developers/badge_component.rb
@@ -8,7 +8,18 @@ module Developers
     end
 
     def call
-      tag.span title, class: "inline-flex items-center rounded-md px-2.5 py-0.5 text-sm font-medium bg-#{color}-100 text-#{color}-800"
+      tag.span title, class: class_names("inline-flex items-center rounded-md px-2.5 py-0.5 text-sm font-medium", color_classes)
+    end
+
+    def color_classes
+      case color.to_sym
+      when :blue
+        "bg-blue-100 text-blue-800"
+      when :green
+        "bg-green-100 text-green-800"
+      else
+        raise "Unknown color: #{color}"
+      end
     end
   end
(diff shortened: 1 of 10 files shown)


app/components/developers/badge_component.rb
```diff
@@ -0,0 +1,14 @@
...
+  class BadgeComponent < ApplicationComponent
+    private attr_reader :title, :classes
+
+    def initialize(title, classes)
+      @title = title
+      @classes = classes
+    end
+
+    def call
+      tag.span title, class: classes.to_s
```
joemasilotti, line 11: Jumping in before you marked this as ready for review, so apologies if you planned on making this change already.

I think this component could be more generic. `BadgeComponent`, for example, and you only pass in `title` and and an identifier to set the color classes. That would move the boilerplate CSS classes to this file and not require us to duplicate them.

We could also reconcile it with `Admin::BadgeComponent` since they would be the same thing!

app/components/developers/card_component.html.erb
```diff
@@ -15,11 +15,14 @@
...
       </div>
 
-      <% if featured? %>
-        <div class="mt-1 pb-1">
-          <span class="inline-flex items-center px-2.5 py-0.5 rounded-md text-sm font-medium bg-blue-100 text-blue-800"> <%= t(".featured") %> </span>
-        </div>
-      <% end %>
+     <div class="mt-1 pb-1">
+        <% if featured? %>
+          <%= render Developers::BadgeComponent.new(t(".featured"), classes: "inline-flex items-center rounded-md px-2.5 py-0.5 text-sm font-medium bg-blue-100 text-blue-800") %>
```
joemasilotti, line 20: Then, down here we could do:

```ruby
<%= render BadgeComponent.new(t(".featured"), color: :blue) %>
```

And on line 23:

```ruby
<%= render BadgeComponent.new(t(".featured"), color: :green) %>
```

app/components/developers/badge_component.rb
sarvaiyanidhi: Sure. That makes more sense. I will do the necessary changes. Thanks for the feedback

app/components/developers/card_component.html.erb
sarvaiyanidhi: Thanks. Will make required changes in the code.

app/components/developers/badge_component.rb
sarvaiyanidhi: Just one point here I had initially coded to pass only bg-color and text-color as parameters and had added other classes added here directly but on second thought I felt that may be sometime you want to even customize other styles like font so I ended up passing all styles. Will change now to pass only color and we can again change in future if we need to customize any other styles.

sarvaiyanidhi:
Here how it looks right now with recently-active badge

recently_active_badge

I need to add required test cases and PR should be ready after that for review.

app/components/developers/badge_component.rb
joemasilotti: Good thinking! I like to optimize for what we need now. We can always refactor to override fonts in the future.

joemasilotti:
Excellent! That is looking really good.

app/components/developers/badge_component.rb
```diff
@@ -0,0 +1,14 @@
...
+  class BadgeComponent < ApplicationComponent
+    private attr_reader :title, :color
+
+    def initialize(title, color:)
+      @title = title
+      @color = color
+    end
+
+    def call
+      tag.span title, class: "inline-flex items-center rounded-md px-2.5 py-0.5 text-sm font-medium bg-#{color}-100 text-#{color}-800"
```
joemasilotti, line 11: Sadly, Tailwind CSS won't be able to pick these up when "parsing" the code [if they are dynamic](https://tailwindcss.com/docs/content-configuration#dynamic-class-names). We need to write out the full class name for each color.

For now, we can do something like this:

```ruby
def call
  tag.span title, class: class_names("inline-flex items-center rounded-md px-2.5 py-0.5 text-sm font-medium", color_classes)
end

def color_classes
  case color.to_sym
  when :blue
    "bg-blue-100 text-blue-100"
  when :green
    "bg-green-100 text-green-100"
  else
    raise "Unknown color: #{color}"
  end
end
```

app/components/developers/card_component.rb
```diff
@@ -22,5 +22,9 @@ def bio
     def highlight?
       !!highlight_featured && featured?
     end
+
+    def recently_active?
+      developer.recently_active?
+    end
```
joemasilotti, line 28:
```suggestion
delegate :recently_active?, to: :developer
```

joemasilotti (changes requested):
Looks great outside of one suggestion and a minor code-style request!

app/components/developers/badge_component.rb
sarvaiyanidhi: You are right. As I had earlier used direct classes and passed as parameter so required css was already added in my file and did not realize it is not working dynamically.

Will make the necessary changes.

joemasilotti:
I pushed a tiny design tweak for the badges on mobile. Once CI passes I'll merge this PR in. Thanks for your contribution, @sarvaiyanidhi!

Issue #541, opened 2022-07-04
Focus in mobile filters dialog


metamoni:
## Describe the bug
In mobile view, the "Filters" button has `aria-expanded="false"`, but this doesn't update to `true` when the menu is expanded. As this button opens a dialog rather than a simple menu, we don't need to use `aria-expanded` here. However, we need to make sure to keep the focus inside the modal while this is open.

## To reproduce

1. Log in as a business
2. Go to [http://localhost:3000/developers](http://localhost:3000/developers)
3. Inspect the page and switch to mobile view
4. Turn on VoiceOver: `command + F5`
5. Tab to the "Filters" button and notice the screen reader announces "collapsed"
6. Hit "Enter" and tab through the modal
7. Keep tabbing through all the elements and notice that the focus moves outside the modal, back to the main page, while the modal is still open
8. Turn VoiceOver off: `command + F5`

## Expected behavior
- Focus should be trapped inside modal while modal is open (should not move back to main page)
- When modal is closed, focus should go back to "Filters" button
- When "Filters" button focused, screen readers should not announce "collapsed"

## Actual behavior
Once you tab through the elements in the modal, the focus goes back to the main page, outside modal
When the modal is closed, the focus is lost and does not go back to "Filters"

## Screenshots

https://user-images.githubusercontent.com/22390758/177101137-61fc8084-3730-4e6c-b443-ed9f49e005b3.mov

## Additional context
- [How to trap focus inside modal](https://uxdesign.cc/how-to-trap-focus-inside-modal-to-make-it-ada-compliant-6a50f9a70700)
- [Dialog Focus in Screen Readers](https://adrianroselli.com/2020/10/dialog-focus-in-screen-readers.html)

atiqueakhtar:
Hi @joemasilotti Can I take up this issue and raise a PR to resolve it?

joemasilotti:
Yeah, sure thing. Thanks @atiqueakhtar!

joemasilotti:
Please see #658 for some ideas and notes. Most importantly, I think this can be accomplished with HTML markup only, no JavaScript.

DamonClark:
Hi @joemasilotti can I pick this one up?

joemasilotti:
Yes, please. Thanks @DamonClark!

DamonClark:
[...]

PR #658, opened 2022-09-18
Trap focus in mobile filter modal while it is open

atiqueakhtar:
This PR resolves the issue described in #541 and satisfies all the given expected behaviour:

- Focus should be trapped inside the modal while the modal is open (should not move back to the main page)
- When the modal is closed, the focus should go back to the "Filters" button
- When the "Filters" button is focused, screen readers should not announce "collapsed"

## Pull request checklist

- [ ] My code contains tests covering the code I modified (Not Required)
- [ ] I added significant changes and product updates to the [changelog](CHANGELOG.md) (Not Required)


atiqueakhtar:
Hi @joemasilotti , Can you please review this PR which resolves #541 . Thanks.

joemasilotti:
Thanks for taking a pass, @atiqueakhtar! I'm not very familiar with a11y in this context, but this feels like a lot of custom JavaScript to achieve what we want. Is there any possibility that we could do something in HTML? I was wondering if manually setting `tabindex` could help.

@metamoni, do you have any ideas or feedback on this?

atiqueakhtar:
> Thanks for taking a pass, @atiqueakhtar! I'm not very familiar with a11y in this context, but this feels like a lot of custom JavaScript to achieve what we want. Is there any possibility that we could do something in HTML? I was wondering if manually setting `tabindex` could help.

Sure @joemasilotti , I'll see if I can accomplish this directly through HTML rather than using custom javascript.

joemasilotti:
Awesome, thanks for taking another pass!

metamoni:
Happy to have a look, yeah!

joemasilotti:
Hey @atiqueakhtar, do you still have appetite to take another pass with HTML? If not, no worries! Let me know and I can take over.

joemasilotti:
@atiqueakhtar, I'm going to close this for now as it has grown quite stale. If you'd like to pick up work again let me know!

Issue #657, opened 2022-09-18
Clean up enabled feature flags


joemasilotti:

ferryphang:
#660 please check if this correct

citizen428:
Should this issue be closed or did the merged PR not address everything?

PR #660, opened 2022-09-19
Cleanup feature model and all related to feature enabled implementation

ferryphang:
This PR resolves the issue described in #657 

- I delete the method and logic inside the Feature.rb
- Remove all if else statement that use the Feature.enabled?

## Pull request checklist


app/models/feature.rb
```diff
@@ -1,18 +1,2 @@
...
-    when :message_read_indicator
-      true
-    when :pricing_v3
-      true
-    when :pricing_v2
-      true
-    when :reply_via_email
-      true
-    when :obfuscate_developer_urls
-      !Rails.env.production?
```
joemasilotti, line 15: Sorry, I wrote this issue before this feature was added. Can you please keep this case and the code around it?

app/views/shared/_main.html.erb
```diff
@@ -1,7 +1,7 @@
 <%= render Developers::BannerComponent.new([
       Developers::UnseededWarningComponent.new(seedable: Rails.env.development?),
       Developers::InvisibleBannerComponent.new(current_user),
-      Developers::NewFieldsComponent.new(current_user, enabled: Feature.enabled?(:new_developer_fields_banner, user: current_user))
+      Developers::NewFieldsComponent.new(current_user)
```
joemasilotti, line 4: IIRC, we can now remove the parameter and code in `Developers::NewFieldsComponent` to check for "enabled" now, too.

app/controllers/pricing_controller.rb
```diff
@@ -8,12 +8,6 @@ def show
...
 
   def template
-    if Feature.enabled?(:pricing_v3, user: current_user)
-      "pricing/v3/show"
-    elsif Feature.enabled?(:pricing_v2, user: current_user)
-      "pricing/v2/show"
-    else
-      "pricing/v1/show"
-    end
+    "pricing/v3/show"
```
joemasilotti, line 11: We can now remove this entire method and let the `#show` method render `app/views/pricing/show.html.erb`. This means removing all of the v1 and v2 templates/partials and moving the v3 stuff to the "root" level of that directory.

joemasilotti (changes requested):
Thanks for taking a pass! I left a few comments and requests – let me know if you have any questions.

app/controllers/pricing_controller.rb
ferryphang: I have move the v3 view to root and improve the controller

app/views/shared/_main.html.erb
ferryphang: Okay i have remove the params for "enabled" and the test for that

app/models/feature.rb
ferryphang: Yup please check if its alright

joemasilotti (changes requested):
Looks great! Thanks for taking the time to go through all the pricing files.

I made one request that can be applied to a few other partials, too. Once that's done this should be good to merge.

app/views/pricing/show.html.erb
```diff
@@ -0,0 +1,13 @@
...
+<div class="bg-gray-50">
+  <%= render "pricing/hero" %>
+  <%= render "pricing/plan" %>
+  <%= render "pricing/stats", stats: @stats %>
+</div>
+
+<div class="bg-white pb-20 lg:pb-30">
+  <%= render "pricing/testimonials", testimonials: @testimonials %>
+  <%= render "pricing/faq" %>
+  <%= render "pricing/cta" %>
```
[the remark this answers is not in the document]
ferryphang: Opps forgot about that

joemasilotti:
Looks great! I pushed a merge commit and will merge when CI passes.

PR #661, opened 2022-09-19
Hide page 2+ of search results for non-subscribers

### LINKED ISSUE #651: Hide page 2+ of search results for non-subscribers
The goal of this issue is to only show the first page of search results unless you have an active hiring subscription.

- [ ] When page 2+ of `/developers` is hit render a CTA (not the developer profiles)
- [ ] CTA design can be copied from the `PaywalledComponent` for now
- [ ] Copy should include that a subscription is required to see more profiles and the total number of developers on the site
- [ ] Ensure that continuing to scroll doesn’t load page 3 of search results
- [ ] Ensure that accessing page 3+ directly via a URL shows the CTA, only

Feel free to use placeholder copy and I can fill it in once I see how it looks.

mansakondo:
Fixes #651.

## Pull request checklist

- [x] When page 2+ of /developers is hit render a CTA (not the developer profiles)
- [x] CTA design can be copied from the PaywalledComponent for now
- [x] Copy should include that a subscription is required to see more profiles and the total number of developers on the site
- [x] Ensure that continuing to scroll doesn’t load page 3 of search results
- [x] Ensure that accessing page 3+ directly via a URL shows the CTA, only


joemasilotti:
It looks like the pagination is still firing even after seeing a paywalled page. This only shows up when there are 30+ developers so I updated the seeds to reflect that. Run `bin/rails db:seed` to add the additional developers to your local database.

image

Once a paywalled page is reached no additional pages should be fetched, no matter how far down the page you scroll.

test/system/scroll_test.rb
```diff
@@ -27,7 +27,7 @@ class ScrollTest < ApplicationSystemTestCase
     refute_text "Developer number one"
 
     scroll_to_bottom_of_page
-    assert_text "Developer number one"
+    assert_text "Private information"
```
joemasilotti, line 30: Let's split this out into 2 tests: one for a subscriber (the existing test) and one for someone who sees the paywall (your updated test).

test/components/developers/results_visibility_component_test.rb
```diff
@@ -0,0 +1,75 @@
+# frozen_string_literal: true
+
```
joemasilotti, line 2: Please remove these lines.

test/components/developers/results_visibility_component_test.rb
```diff
@@ -0,0 +1,75 @@
...
+require "test_helper"
+
+module Developers
+  class ResultsVisibilityComponentTest < ViewComponent::TestCase
+    include DevelopersHelper
+
+    setup do
+      20.times do
+        create_developer(hero: "Available", available_on: Date.today)
+      end
```
joemasilotti, line 12: Instead of creating 20 additional developers (for every test) we can pass in `items_per_page` to `DeveloperQuery`. Then we might only have to create an additional developer or two, which should speed these tests up quite a bit.

app/components/developers/results_visibility_component.rb
```diff
@@ -0,0 +1,56 @@
+# frozen_string_literal: true
+
```
joemasilotti, line 2: Please remove these lines.

mansakondo:
Yeah, that's what I was afraid of. Thanks for updating the seeds, I'll try to fix that.

app/views/developers/index.html.erb
```diff
@@ -1,25 +1,25 @@
 <%= open_graph_tags title: @meta.title, turbo_native_title: t(".title"), description: @meta.description %>
```
joemasilotti, line 1: It looks like a lot of the classes were reordered. This is making it a bit hard to tell what was actually changed. Can you please remove the ordering changes so all that's left are additions/subtractions?

Same goes for the other HTML files.

app/components/developers/results_visibility_component.rb
```diff
@@ -0,0 +1,56 @@
...
+module Developers
+  class ResultsVisibilityComponent < ApplicationComponent
+    renders_one :results_set, ->(records:) do
+      CardComponent.with_collection(records)
+    end
+
+    def initialize(query:, user:, paywalled: nil, size: :large, title: nil, description: nil)
```
joemasilotti, line 9: I think this initializer only needs two parameters: what page are we on (`page`) and who is the current user (`user`).

This file has a lot of logic that looks like it was copied for showing your own profile information. That isn't relevant for search results since we don't care about the individual profile but the entire list at once.

I think everything else can be removed or moved to a private method.

app/components/developers/results_visibility_component.rb
```diff
@@ -0,0 +1,56 @@
...
+      @query = query
+      @user = user
+      @paywalled = paywalled
+      @size = size
+      @title = title
+      @description = description
+    end
+
+    def render_content?
+      pagy.page > 1 ? customer? || owner? : true
```
joemasilotti, line 19: This line is a bit hard to understand – can you please split it out to multiple statements?

joemasilotti (changes requested):
Thanks for taking a pass at this! I made some requests along with the comment posted above. Let me know if you have any questions on my feedback.

test/components/developers/results_visibility_component_test.rb
mansakondo: Indeed, that would be much better.

app/views/developers/index.html.erb
mansakondo: Sure no problem.

app/components/developers/results_visibility_component.rb
mansakondo: Yeah, I kinda copied the logic from the `PaywalledComponent` without cleaning it up.

app/components/developers/results_visibility_component.rb
mansakondo: Yeah, I'm too lazy sometimes. My bad 😅

joemasilotti: Ha, all good!

app/components/developers/results_visibility_component.rb
mansakondo: I used `query` instead of `page` because I also want to retrieve the records.

Also, it seems like we need `paywalled` to check if a user owns a profile unless you don't care about that.

app/components/developers/results_visibility_component.rb
```diff
@@ -0,0 +1,56 @@
...
+
+    def records
+      @query.records
+    end
+
+    private
+
+    def customer?
+      Businesses::Permission.new(@user&.subscriptions).active_subscription?
+    end
```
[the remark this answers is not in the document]
joemasilotti: Hmm. I think I'd rather have a method that exposes the permissions so we can ask all of the possible questions from one spot.

```ruby
def permissions
  Businesses::Permission.new(subscriptions)
end

user.permissions.active_subscription?
```

Also, `#customer` already exists on `User`, so having `#customer?` that does something different feels a bit weird.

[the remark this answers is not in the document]
mansakondo: Makes sense.

mansakondo:
I've improved the code let me know what you think.

app/models/user.rb
```diff
@@ -45,4 +45,8 @@ def remember_me
   def signed_hiring_agreement?
     HiringAgreements::Term.signed_by?(self)
   end
+
+  def permissions
+    Businesses::Permission.new(subscriptions)
```
joemasilotti, line 50: We use this in a bunch of other places in the app. If we are going to make the refactor we should also update the rest of the codebase.

Let me know if you'd like to do that in this PR or a future one. If we do it in a future one then let's remove this abstraction from this PR and add it all in at once.

app/queries/developer_query.rb
```diff
@@ -79,6 +78,17 @@ def query_and_paginate
     @pagy, @records = build_pagy(@_records, items: items_per_page)
   end
 
+  def build_pagy(collection, options = {})
```
joemasilotti, line 81: This needs a test in `test/queries/developer_query_test.rb` for this new behavior.

joemasilotti (changes requested):
Getting close! I made two requests – let me know if you have any questions.

app/queries/developer_query.rb
mansakondo: Sure, I'm on it.

app/models/user.rb
mansakondo: I think that it's better to introduce this abstraction in a different PR.

mansakondo:
Getting closer! Let me know what you think about the last commits.

mansakondo:
The CTA is displayed along with the `EmptyStateComponent` when accessing page2+ via URL:
Capture d’écran de 2022-09-20 20-25-27

Is it ok with you?

joemasilotti:
Code looks great! Thanks for taking another pass.

> The CTA is displayed along with the `EmptyStateComponent` when accessing page 2+ via URL

Ideally I would like to hide the empty state component in this scenario.

Also, and this should be the last change, can you wrap the logic in `DeveloperQuery` in a feature flag? `Feature.rb` can have a new case like `:paywalled_search_results` which is disabled in production.

mansakondo:
I updated the PR with the last changes.

joemasilotti:
We're very close! The last thing we need to take care of is when the feature flag is disabled and you don't have a subscription. It looks like the `SubscriptionCTAComponent` is still rendering.

image

mansakondo:
> We're very close! The last thing we need to take care of is when the feature flag is disabled and you don't have a subscription. It looks like the `SubscriptionCTAComponent` is still rendering.
> 
> image

Hmm, that's weird. I'll try to debug and fix this.

mansakondo:
It should work as expected now.

joemasilotti:
Works great! Thanks for your contribution @mansakondo and for all the back and forth. Merging in now!

mansakondo:
> Works great! Thanks for your contribution @mansakondo and for all the back and forth. Merging in now!

My pleasure @joemasilotti . It was nice working with you on this!

mansakondo:
I guess the only thing left to do is the refactoring with `User#permissions`.

joemasilotti:
You as well! And yeah, that would be great if you're up for it.

mansakondo (commented):
What do you think about those suggestions?

app/queries/developer_query.rb
```diff
@@ -79,6 +78,23 @@ def query_and_paginate
...
+    results = [pagy, collection.offset(pagy.offset).limit(pagy.items)]
+
+    if Feature.enabled?(:paywalled_search_results)
+      unless subscribed_business?
+        results = [pagy, []] if pagy.page > 1
+      end
+    end
+
+    results
+  end
```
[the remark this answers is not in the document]
joemasilotti: Mind opening a new PR for this change? It's a little hard to track inside of an already merged one. Thanks!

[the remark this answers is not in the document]
mansakondo: Sure. I put that in there to know if I should open a PR or not.

joemasilotti:
Hey again @mansakondo. I'm hoping to flip the feature flag on this soon but am running into some issues. I'm not sure if I overlooked them initially or some recent changes broke things. Do you have time to take a look and see what's going on?

Visiting `/developers` as a non-subscriber and scrolling down seems to redirect the entire page to `?page=2`. I'm not entirely sure why or how this is happening but I'd appreciate any insights you might hav!

joemasilotti:
If it helps, checking out this branch's code directly still shows the issue.

mansakondo:
Hey @joemasilotti. Sorry for the late reply.

I'll try to investigate this week to see what's going on.

mansakondo:
At first glance, I like this implementation! It's way cleaner with the `PaywalledSearchResults` abstraction 👍🏾

The tests are also cleaner with the stub feature flag.

mansakondo:
> I think this is a bit cleaner and gets us 90% of what we want.
What are the missing 10%?

joemasilotti:
> I didn't test the implementation yet, so what are the missing 10%?

Your implementation enabled us to change one line of code and paywall a different page. This implementation locks us in to paywalling page 2.

Not a big deal as we might never change it. But I think its worth calling out.

mansakondo:
Ah, I see. It doesn't feel like a big deal to me either (unless I'm missing something).

However, I just noticed that you also removed the paywall from the location and time zone filters. I'm curious to know why.

joemasilotti:
> However, I just noticed that you also removed the paywall from the location and time zone filters. I'm curious to know why.

I realized that (for now) I'd rather get Google indexing a few more pages instead of locking down results to potential customers.

mansakondo:
I see. Make sense

mansakondo:
> Visiting `/developers` as a non-subscriber and scrolling down seems to redirect the entire page to `?page=2`. I'm not entirely sure why or how this is happening but I'd appreciate any insights you might have!

I tried to reproduce this issue but I don't get redirected to `?page=2` when I'm scrolling down 🤔

PR #665, opened 2022-09-21
Publicly shareable developer profile URLs

sarvaiyanidhi:
Implemented feature to publicly share developer URL using public profile key.

- [x] Add a new column to the database, developers.public_profile_key
- [x] Unhide hidden information on a developers profile page if this key matches from the key param (e.g. /developers/abc123?key=xyz987)
- [x] Add a new button next to “Edit” when viewing your own developer profile titled “Share”. When clicked:
  - [x] Generate the public profile key via Secure Random (if not already generated)
   - [x]  Copy the full URL (with key param) to the clipboard

## Pull request checklist

### DIFF
--- a/app/components/developers/primary_action_component.rb
+++ b/app/components/developers/primary_action_component.rb
@@ -15,5 +15,9 @@ module Developers
     def conversation
       Conversation.find_by(developer:, business:)
     end
+
+    def share_url
+      developer.share_url(developer_url(@developer))
+    end
   end
 end
--- a/app/components/users/paywalled_component.rb
+++ b/app/components/users/paywalled_component.rb
@@ -40,7 +40,7 @@ module Users
     end
 
     def valid_public_profile_access?
-      @paywalled&.public_profile_key == @public_key
+      @paywalled&.public_profile_key == @public_key && @paywalled&.public_profile_key.present?
     end
   end
 end
--- a/app/components/users/paywalled_component.rb
+++ b/app/components/users/paywalled_component.rb
@@ -1,15 +1,16 @@
 module Users
   class PaywalledComponent < ApplicationComponent
-    def initialize(user:, paywalled:, size: nil, title: nil, description: nil)
+    def initialize(user:, paywalled:, size: nil, title: nil, description: nil, public_key: nil)
       @user = user
       @paywalled = paywalled
       @size = size
       @title = title
       @description = description
+      @public_key = public_key
     end
 
     def render_content?
-      customer? || owner?
+      customer? || owner? || valid_public_profile_access?
     end
 
     def small?
@@ -36,6 +37,10 @@ module Users
 
     def owner?
       @paywalled&.user == @user && @user.present?
+    end
+
+    def valid_public_profile_access?
+      @paywalled&.public_profile_key == @public_key
     end
   end
 end
--- a/app/javascript/src/clipboard.js
+++ b/app/javascript/src/clipboard.js
@@ -1,8 +1,8 @@
 class Clipboard {
   static copy({text, html}) {
     const data = [
-      this.clipboardItem({type: "text/plain", content: text}),
-      this.clipboardItem({type: "text/html", content: html})
+      this.clipboardItem({type: "text/plain", content: text})
+     // this.clipboardItem({type: "text/html", content: html})
     ]
 
     navigator.clipboard.write(data).then(
--- a/app/models/concerns/public_profile.rb
+++ b/app/models/concerns/public_profile.rb
@@ -1,36 +1,35 @@
 module PublicProfile
-
   extend ActiveSupport::Concern
-  
+
   def share_url(url)
-    update_public_profile_key if self.public_profile_key.blank?    
-    share_url = form_url(url, self.public_profile_key)
+    update_public_profile_key if public_profile_key.blank?
+    form_url(url, public_profile_key)
   end
 
   private
 
-    def form_url(url, query_param)
-      parsed = URI.parse(url)
-      query = if parsed.query
-              CGI.parse(parsed.query)
-            else
-              {}
-            end
-
-      query['key'] = query_param
-
-      parsed.query = URI.encode_www_form(query)
-      parsed.to_s
+  def form_url(url, query_param)
+    parsed = URI.parse(url)
+    query = if parsed.query
+      CGI.parse(parsed.query)
+    else
+      {}
     end
 
-    def update_public_profile_key
-      update!(public_profile_key: generate_token)
-    end
+    query["key"] = query_param
 
-    def generate_token
-      loop do
-        public_key = SecureRandom.hex(10)
-        break public_key unless self.class.find_by(public_profile_key: public_key)
-      end
+    parsed.query = URI.encode_www_form(query)
+    parsed.to_s
+  end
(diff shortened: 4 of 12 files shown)


app/javascript/src/clipboard.js
```diff
@@ -1,8 +1,8 @@
 class Clipboard {
   static copy({text, html}) {
     const data = [
-      this.clipboardItem({type: "text/plain", content: text}),
-      this.clipboardItem({type: "text/html", content: html})
+      this.clipboardItem({type: "text/plain", content: text})
+     // this.clipboardItem({type: "text/html", content: html})
```
sarvaiyanidhi, line 5: Had to comment this code to get copy clipboard functionality working in chrome browser.

sarvaiyanidhi:
@joemasilotti This PR is still in progress as need to add relevant test cases but can you please review the implemented code and provide feedback for any changes. Thanks

sarvaiyanidhi:
> Can you please merge in `main`? #666 should fix the JavaScript bug in Chrome.

Done.. Thanks @ferryphang for the fix. 

It is working fine now in chrome though Firefox it is still throwing below error

```
Error invoking action "click->clipboard#copy"

ReferenceError: ClipboardItem is not defined
```

joemasilotti:
I didn't test on Firefox – thanks for calling that out. It looks like the [ClipboardItem API](https://caniuse.com/?search=clipboarditem) isn't support on that browser.

However, [Mozilla's docs](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/Interact_with_the_clipboard) reference using `document.execCommand("copy")`, even though it is deprecated. [This gist](https://gist.github.com/n1k0/b17b5c248a3ee1df99acaae000eccae4) looks like it should work with that API, but it does require a bit of additional JavaScript.

@sarvaiyanidhi, can you test on Chrome for now? If not, I can take a pass at implementing the Firefox-specific behavior.

Either way, I'll start reviewing the PR now.

app/components/users/paywalled_component.rb
```diff
@@ -37,5 +38,9 @@ def customer?
     def owner?
       @paywalled&.user == @user && @user.present?
     end
+
+    def valid_public_profile_access?
+      @paywalled&.public_profile_key == @public_key && @paywalled&.public_profile_key.present?
+    end
```
joemasilotti, line 44: Let's move this to a method on the `PublicProfile` module.

app/models/concerns/public_profile.rb
```diff
@@ -0,0 +1,35 @@
...
+    parsed.to_s
+  end
+
+  def update_public_profile_key
+    update!(public_profile_key: generate_token)
+  end
+
+  def generate_token
+    loop do
+      public_key = SecureRandom.hex(10)
```
joemasilotti, line 31:
```suggestion
      public_key = SecureRandom.hex(4)
```

I know it creates a higher chance of collision but then it is 8 characters like the developer profile slug.

app/components/developers/primary_action_component.rb
```diff
@@ -15,5 +15,9 @@ def owner?
     def conversation
       Conversation.find_by(developer:, business:)
     end
+
+    def share_url
+      developer.share_url(developer_url(@developer))
```
joemasilotti, line 20:
```suggestion
      developer.share_url(developer_url(developer))
```

joemasilotti (changes requested):
Outside of missing tests this is looking great! I pushed a small design tweak and wrapped the button in a feature flag. I also made a few requests/questions but overall the architecture is solid. I really like the new concern.

sarvaiyanidhi:
Thank you for the feedback. Have gone through comments and will make necessary changes.
Will also add required test cases and update PR for further review.

app/models/concerns/public_profile.rb
```diff
@@ -0,0 +1,29 @@
...
+    polymorphic_url(self, key: public_profile_key)
+  end
+
+  def valid_public_profile_access?(resource, profile_key)
+    resource&.public_profile_key == profile_key && resource.public_profile_key.present?
+  end
+
+  private
+
+  def update_public_profile_key
```
joemasilotti, line 19:
```suggestion
  def generate_public_profile_key
```

app/models/concerns/public_profile.rb
```diff
@@ -0,0 +1,29 @@
...
+    include UrlHelpersWithDefaultUrlOptions
+  end
+
+  def share_url
+    update_public_profile_key if public_profile_key.blank?
+    polymorphic_url(self, key: public_profile_key)
+  end
+
+  def valid_public_profile_access?(resource, profile_key)
+    resource&.public_profile_key == profile_key && resource.public_profile_key.present?
```
joemasilotti, line 14:
```suggestion
    resource&.public_profile_key.present? && resource.public_profile_key == profile_key
```

test/components/developers/primary_action_component_test.rb
```diff
@@ -35,5 +35,11 @@ class PrimaryActionComponentTest < ViewComponent::TestCase
       render_inline PrimaryActionComponent.new(user:, developer:, business:)
       assert_text "Message"
     end
+
+    test "renders Share button to owner of profile" do
```
joemasilotti, line 39: This test also needs another case: when someone else is viewing the profile.

test/components/users/paywalled_component_test.rb
```diff
@@ -49,5 +49,22 @@ class PaywalledComponentTest < ViewComponent::TestCase
...
+    end
+
+    test "should show small CTA if profile access key does not match" do
+      user = users(:business)
+      developer = developers(:one)
+      developer.share_url
+      render_inline(PaywalledComponent.new(user:, paywalled: developer, size: :large, public_key: "randomkey")) { "Test text" }
+      assert_text I18n.t("users.paywalled_component.title")
+      assert_text I18n.t("users.paywalled_component.description")
+    end
```
joemasilotti, line 68: Let's also add an assertion that verifies "Test text" does _not_ appear here.

joemasilotti (changes requested):
This is looking great! I made a few small requests in the code that should be quick fixes.

Outside of those, two larger requests:

1. We need tests for `PublicProfile`. Feel free to add this to a new test class `Users::PublicProfile` (or similar). I'm OK with coupling the testing logic to the `User` model for now.
2. I'd like to change the URL from `/developers/:id?code=:code` to `/developers/:id/:code`. I think that reads cleaner and should only require a new route. But it needs to also work with `/developers/:id/edit` and such.

Let me know if you still have appetite to make these changes. I know I took a while to review the PR!

sarvaiyanidhi:
Thank you for the input. I will get the mentioned changes done soon and push code for further review.

joemasilotti:
This looks great! All that's left is the following:

> We need tests for `PublicProfile`. Feel free to add this to a new test class `Users::PublicProfile` (or similar). I'm OK with coupling the testing logic to the User model for now.

sarvaiyanidhi:
@joemasilotti Have added pending PublicProfile test cases but let me know if I need to update anything as I have just added basic checks in it. Most of the scenarios were covered in other related files so I was not sure what else to update in this.

PR #666, opened 2022-09-21
Fix clipboard bug

ferryphang:
Tested on Chrome, and Safari. One question though, I wonder why we need the HTMLvalue, if not i think we can use writeText


joemasilotti:
Works great, thanks for this @ferryphang! I made some code formatting updates (like removing the semicolons you added) and renamed a variable.

The HTML value is used for pasting directly into an email so it gets hyperlinked correctly.

PR #667, opened 2022-09-21
Redirect normal ID to hashid equivalent

### LINKED ISSUE #662: Redirect `/developers/:id` to `/developers/:hashid`
While folks are transitioning to the new developer profile URLs I want to make sure no one's links break.

**When the `:obfuscate_developer_urls` feature is enabled:**

- [ ] 302 redirect `/developers/:id` to `/developers/:hashid` with a flash message
- [ ] Update the sitemap (`config/sitemap.rb:24`) to use the developer's `hashid`

The flash message should read:

> Links to developer profiles have changed. Please update your bookmark to {new developer profile URL}.

rayhanw:
I ended up with a bunch of [`if` conditions within `developers_controller#show`](https://github.com/rayhanw/railsdevs.com/blob/68ec110034b7d35f7a8fc3a11b462226aba83537/app/controllers/developers_controller.rb#L43-L57) which I feel that it could be more graceful, but I'm not really sure how. Any guidance here is appreciated 🙏. 

closes #662 

## Pull request checklist

- [X] My code contains tests covering the code I modified


test/integration/developers_test.rb
```diff
@@ -57,14 +57,12 @@ class DevelopersTest < ActionDispatch::IntegrationTest
...
-    end
+    get developer_path(developer.id)
+    assert_redirected_to developer_path(developer.hashid)
 
     sign_in users(:developer)
-    assert_raises ActiveRecord::RecordNotFound do
-      get edit_developer_path(developer.id)
-    end
+    get developer_path(developer.id)
+    assert_redirected_to developer_path(developer.hashid)
```
rayhanw, line 65: Would I also be correct here to assume that it should no longer raise an error if users still access via normal ID? @joemasilotti

config/sitemap.rb
```diff
@@ -21,7 +21,8 @@
   add about_path, changefreq: "monthly", priority: 0.9
 
   Developer.visible.newest_first.find_each do |developer|
-    add developer_path(id: developer.id), changefreq: "always", priority: 0.8, lastmod: developer.updated_at
+    id = Feature.enabled?(:obfuscate_developer_urls, user: nil) ? developer.hashid : developer.id
+    add developer_path(id: id), changefreq: "always", priority: 0.8, lastmod: developer.updated_at
```
joemasilotti, line 25:
```suggestion
    add developer_path(id:), changefreq: "always", priority: 0.8, lastmod: developer.updated_at
```

app/controllers/developers_controller.rb
```diff
@@ -41,7 +41,18 @@ def update
...
+      if /\D+/.match?(params[:id])
+        @developer = Developer.find_by_hashid!(params[:id])
+      else
+        @developer = Developer.find(params[:id])
+        redirect_to @developer, status: 302, notice: t(".redirection", url: developer_url(@developer))
+      end
+    else
+      @developer = Developer.find(params[:id])
+    end
+
```
joemasilotti, line 55: I think we need some tests for this. There are a few edge cases that aren't covered and I'm having some trouble reasoning them about.

Maybe pulling it out to a `DeveloperFinder` object, or similar, could help make that a bit easier?

joemasilotti (changes requested):
Thanks for the PR! I left one code suggestion and a bigger question about the logic you wrote. Let me know if you have any questions.

app/controllers/developers_controller.rb
rayhanw: I extracted the logic to a `Developers::Finder` object and added some [tests for it](https://github.com/joemasilotti/railsdevs.com/pull/667/files#diff-39d8d4541629ba3c175e9180086b2d3c4291153cab51512fb185d0d1a2336878R1-R20).

For the task:
> 302 redirect /developers/:id to /developers/:hashid with a flash message

I modified the [tests here](https://github.com/joemasilotti/railsdevs.com/pull/667/files#diff-d18d5d631cf8cfd340082086aca016edb6da051bb625499cd2d6a0ee1874a3d3R60-R65) in `test/integration/developers_test.rb`

test/models/developers/finder_test.rb
```diff
@@ -0,0 +1,20 @@
+require "test_helper"
+
+class Developers::FinderTest < ActiveSupport::TestCase
```
joemasilotti, line 3: We need tests for when the feature is disabled, too. Right now we are only testing two of the three possibilities.

We can mock the feature with the following:

```ruby
Feature.stub(:enabled?, true) do
  # Do the stuff.
  # Feature.enabled? will be reset at the end of the block.
end
```

We should make sure to stub

test/integration/developers_test.rb
joemasilotti, line 65: I'm not sure we need both of these tests anymore – they are testing the same thing, right?

app/controllers/developers_controller.rb
```diff
@@ -41,7 +41,13 @@ def update
...
 
   def show
-    @developer = find_developer!
+    finder = Developers::Finder.new(id: params[:id]).call
+    @developer = finder.developer
+
+    if finder.should_redirect?
+      redirect_to @developer, status: 302, notice: t(".redirection", url: developer_url(@developer))
+    end
+
```
joemasilotti, line 50: Nice! I like how you pulled out the redirection question.

But I don't think we need the `#call` method. We could have `#developer` and `#should_redirect?` memoize so no matter which one you call first you get the same behavior.

For example,

```ruby
class Finder
  # ...

  def developer
    find_developer unless defined?(@developer)
    @developer
  end

  def should_redirect?
    find_developer unless defined?(@redirect)
    @redirect
  end

  private

  def find_developer
    # logic of #call
    # set @developer and @redirect
  end
end
```

joemasilotti (changes requested):

test/integration/developers_test.rb
rayhanw: yep you are correct

test/models/developers/finder_test.rb
rayhanw: Added the tests with the feature being disabled!

app/controllers/developers_controller.rb
rayhanw: Interesting, never thought to memoize this way. Cool stuff!

rayhanw:
It seems I accidentally tagged a random person in my commit 😅

joemasilotti:
Looks great - thanks for the changes! I need to do a bit more manual testing. But outside of that, can you replace the tests checking for false to `refute`? That’s a bit more concise. Thanks again!

rayhanw:
Updated! @joemasilotti

joemasilotti:
Looks great – merging in now! Thanks for your contribution.

PR #673, opened 2022-09-22
Unify 'Found a job?' with other mobile menu links

xmarlox:
## Pull request checklist


joemasilotti:
Ohh, good catch! Thanks for this.

Issue #652, opened 2022-09-16
Paywall location and timezone search, too


joemasilotti:
The goal of this issue is to ensure someone doesn't "find" a candidate by filtering so tightly that they show up on the first page of search results.

A PR addressing this issue should:

- [ ] Require an active subscription to filter by location and/or timezone
- [ ] This includes the UI (like text search) and updating the Pundit policy

For reference, check out how the search query works. We might be able to borrow paywall logic from that.

mansakondo:
I'd like to work on this one but I also volunteered for #653. So, I suppose that I should choose between the two, right?

joemasilotti:
Both are needed to launch the new feature – which do you have a preference of starting, first? If it helps, this one should be more cut-and-dry since there are no remaining design decisions (I don't think).

mansakondo:
I'd like to start with this one since there are fewer unknowns, but I'll still let you know if I have ideas for the CTA design.

mansakondo:
I'm experimenting with a design where I added a variant for the `PaywallComponent` to only display the lock icon.:
![Capture d’écran de 2022-09-22 15-37-36](https://user-images.githubusercontent.com/47113995/191762198-12fa3887-9717-43ed-a295-429068ada024.png)

What do you think?

mansakondo:
> Ohh, I like that design! How are you rendering the hover?

I'm glad that you like it! I'll open a draft PR right now to allow see the code. 

> Does it require any additional JavaScript or can we do it with CSS?

No additional JavaScript is required.

mansakondo:
The draft PR is open. There's some dirty code waiting to be refactored.

joemasilotti:
Got it – thanks for that!

Do you need an early set of eyes on the PR? Or would you like me to wait to review until it is marked as ready?

mansakondo:
> Got it – thanks for that!
> 
> Do you need an early set of eyes on the PR? Or would you like me to wait to review until it is marked as ready?

Though the code still needs to be refactored, I'm open to suggestions.

joemasilotti:
Closed via #674.

PR #674, opened 2022-09-22
Paywall location and timezone filters

mansakondo:
See #652.

## Pull request checklist

- [x] Require an active subscription to filter by location and/or timezone
- [x] This includes the UI (like text search) and updating the Pundit policy


mansakondo:
It seems like clicking on the CTA doesn't follow the link because the toggle prevents the default behavior of click events:
https://github.com/joemasilotti/railsdevs.com/blob/main/app/javascript/controllers/toggle_controller.js#L15

joemasilotti:
Maybe I missed something, but it looks like you need to standardize, not a test failure.

mansakondo:
> Maybe I missed something, but it looks like you need to standardize, not a test failure. 

Yeah, the CI didn't run the tests because of that. But if you run the tests against this branch you should see the following test fail.

mansakondo:
I fix the standardization issue. Now we see that the test failed the CI's details.

joemasilotti:
Gotcha, thanks for pushing that upstream. This is an interesting question.

Putting country search behind a paywall would remove most of programmatic SEO pages from #507 and leave us with only 10 (5 role levels x freelance or not). But keeping it means someone could more easily view a ton of profiles by manually selecting each country and viewing the first page of results.

Right now I'm leaning towards adding both paywalls, like the original issue stated, and removing the country parameter in `Developers::QueryPath`.

Also, let's make sure that if the feature flag is disabled then someone can continue to search by country. Thanks again!

mansakondo:
> Gotcha, thanks for pushing that upstream. This is an interesting question.
> 
> Putting country search behind a paywall would remove most of programmatic SEO pages from #507 and leave us with only 10 (5 role levels x freelance or not). But keeping it means someone could more easily view a ton of profiles by manually selecting each country and viewing the first page of results.
> 
> Right now I'm leaning towards adding both paywalls, like the original issue stated, and removing the country parameter in `Developers::QueryPath`.
> 
> Also, let's make sure that if the feature flag is disabled then someone can continue to search by country. Thanks again!

Ok, I see. Thanks for the feedback.

mansakondo:
I'm not sure how I should update this [test](https://github.com/joemasilotti/railsdevs.com/blob/700c54720991f02b69bbc483854dbdbebf33315d/test/models/developers/query_path_test.rb#L4) though. Since it's not an integration test I can't sign in as a subscriber to make this pass.

mansakondo:
Though I'm still not satisfied with the code yet, I'm gonna marked it as "Ready for review".

app/components/collapse_control_component.html.erb
```diff
@@ -1,7 +1,22 @@
...
+      <span class="ml-6 flex items-center">
+        <%= inline_svg_tag "icons/solid/minus_sm.svg", class: minus_icon_class, data: {"toggle-target": "element"}, title: "Open menu" %>
+        <%= inline_svg_tag "icons/solid/plus_sm.svg", class: plus_icon_class, data: {"toggle-target": "element"}, title: "Close menu" %>
+      </span>
+    <% end %>
+  <% end %>
+
+  <% if paywalled? && !customer? %>
+    <%= render Users::PaywalledComponent.new(user: user, paywalled: nil, size: :smallest) %>
+  <% end %>
```
joemasilotti, line 21: I don't like mixing logic of paywalls + customers with a generic control like this. Instead, I think we should refactor `Users::PaywalledComponent` to accept a slot for custom CTA content.

app/components/developers/query_component.html.erb
```diff
@@ -57,9 +57,9 @@
     <div data-controller="toggle accessibility" data-toggle-visibility-class="hidden">
       <div class="border-t border-gray-200 py-6">
         <h3 class="-mx-2 -my-3 flow-root">
-          <%= render CollapseControlComponent.new(t(".country.title"), collapsed: true) %>
+          <%= render CollapseControlComponent.new(t(".country.title"), collapsed: true, paywalled: {user: user}) %>
         </h3>
-        <div data-toggle-target="element" class="hidden">
+        <div<%= toggle_target_attribute %> class="hidden">
```
joemasilotti, line 62:
```suggestion
        <%= tag.div "data-toggle-target": toggle_target, class="hidden" do %>
```

app/components/developers/query_component.rb
```diff
@@ -65,5 +65,17 @@ def include_not_interested?
...
+      user&.permissions&.active_subscription?
+    end
+
+    def toggle_target_attribute
+      if Feature.enabled?(:paywalled_search_results)
+        " data-toggle-target=element" if customer?
+      else
+        ""
+      end
+    end
```
joemasilotti, line 79:
```suggestion
    def toggle_target_attribute
      "element" if Feature.enabled?(:paywalled_search_results) && customer?
    end
```

app/models/developers/query_path.rb
```diff
@@ -2,6 +2,12 @@ module Developers
   class QueryPath
     include UrlHelpersWithDefaultUrlOptions
 
+    attr_accessor :paywalled_search_results
```
joemasilotti, line 5:
```suggestion
    private attr_reader :paywalled_search_results
```

joemasilotti (changes requested):
I think that we are polluting the collapse component a bit with logic about the user/customer. See my inline comment for more details.

Let me know if you still have appetite to finish this – I know I took a while to get around to reviewing it.

app/components/collapse_control_component.html.erb
mansakondo: Yeah, I don't like that either. Using a slot to handle custom CTA content is a nice idea

mansakondo:
> I think that we are polluting the collapse component a bit with logic about the user/customer. See my inline comment for more details.
> 
> Let me know if you still have appetite to finish this – I know I took a while to get around to reviewing it.

All good. I simplified the logic. Let me know what you think.

test/models/developers/query_path_test.rb
```diff
@@ -2,7 +2,10 @@
 
 class Developers::QueryPathTest < ActiveSupport::TestCase
   test "generates unique URL paths for combinations of role levels, location, and freelance" do
-    paths = Developers::QueryPath.all
+    query_path = Developers::QueryPath.new
+    query_path.instance_variable_set(:@paywalled_search_results, false)
```
joemasilotti, line 6: Let's inject this via a a defaulted initializer instead of reaching in under test.

```ruby
class QueryPath

  private attr_reader :paywalled_search_results

  def initialize(paywalled_search_results: Feature.enabled?(:paywalled_search_results))
    @paywalled_search_results = paywalled_search_results
  end

  # ...
end
```

joemasilotti (changes requested):
Nice progress! But the collapse component still depends on the paywalled component, which I don't like. I'd rather see us render them both at a higher level to not couple them so tightly.

Does that make sense? If not, let me know and I can show you a scaffold of what I'm trying to explain.

mansakondo:
> Nice progress! But the collapse component still depends on the paywalled component, which I don't like. I'd rather see us render them both at a higher level to not couple them so tightly.
> 
> Does that make sense? If not, let me know and I can show you a scaffold of what I'm trying to explain.

I think I get it. You want an abstract slot that doesn't know about it's content, instead of a concrete slot, right?

joemasilotti:
An abstract slot on the paywall component, yep! Something like this (in pseudo code):

```ruby
<%= render PaywalledComponent.new do |paywall| %>
  <%= render CollapseComponent.new %>
  <% paywall.custom do %>
    <h3>Locations</h3>
  <% end %>
<% end %>
```

mansakondo:
> An abstract slot on the paywall component, yep! Something like this (in pseudo code):
> 
> ```ruby
> <%= render PaywalledComponent.new do |paywall| %>
>   <%= render CollapseComponent.new %>
>   <% paywall.custom do %>
>     <h3>Locations</h3>
>   <% end %>
> <% end %>
> ```

Ah ok, that makes sense.

mansakondo:
> An abstract slot on the paywall component, yep! Something like this (in pseudo code):
> 
> ```ruby
> <%= render PaywalledComponent.new do |paywall| %>
>   <%= render CollapseComponent.new %>
>   <% paywall.custom do %>
>     <h3>Locations</h3>
>   <% end %>
> <% end %>
> ```

I've been thinking about this and I couldn't find a way to implement the design with the abstract slot on the paywall component. The thing is that the CTA needs to be rendered inside the collapse control component and I can't really do that if there are no connections between the two components.

joemasilotti:
Can we render a "fake" collapse control with only the title and CTA in the slot?

mansakondo:
> Can we render a "fake" collapse control with only the title and CTA in the slot?

Sorry for the later awswer. I was reluctant to go that route but it's possible. If you don't mind having something like this then we can use this approach.
```erb
<%= render Users::PaywalledComponent.new(user: user, paywalled: nil, size: :extra_small) do |paywall| %>
  <%= render CollapseControlComponent.new(t(".timezone.title"), collapsed: collapse_timezone?) %>
  <% paywall.custom do %>
    <div class="flex items-center">
      <%= tag.button type: "button", class: "px-2 py-3 w-full flex items-center justify-between text-gray-400 hover:text-gray-500 text-sm" do %>
        <%= tag.span t(".timezone.title"), class: ["font-medium text-gray-900"] %>
      <% end %>

      <%= link_to pricing_path, class: "group flex items-center bg-gray-50 max-w-sm border border-gray-300 hover:border-gray-400 border-dashed rounded-lg p-1.5 px-2.5" do %>
        <%= inline_svg_tag "icons/solid/lock_closed.svg", class: "h-4 w-4 text-gray-400 group-hover:hidden", aria_hidden: true %>
        <%= inline_svg_tag "icons/solid/lock_open.svg", class: "h-4 w-4 text-gray-500 rotate-[-8deg] transition hidden group-hover:block", title: "You need a paid subscription to unlock this feature", aria_hidden: true %>
      <% end %>
    </div>
  <% end %>
<% end %>
```

joemasilotti:
Yeah, I think something like that could work!

We could probably move the `link_to` part into the component, too. The component could check if the `custom` slot is passed in and if so render it followed by the `link_to`.

That would DRY this up even more so the only repeated code is the styling of the `<button>`. And that could maybe even be simplified to `px-2 py-3 w-full text-gray-400 text-sm`.

mansakondo:
> Yeah, I think something like that could work!
> 
> We could probably move the `link_to` part into the component, too. The [component could check](https://viewcomponent.org/guide/slots.html#predicate-methods) if the `custom` slot is passed in and if so render it followed by the `link_to`.
> 
> That would DRY this up even more so the only repeated code is the styling of the `<button>`. And that could maybe even be simplified to `px-2 py-3 w-full text-gray-400 text-sm`.

Yes, we could DRY this even more. It also means that the slot will become more concrete and less abstract. So, I think that we should find a more concrete name for the slot. I know it's just a detail but I'm obsessed with naming :)

What do you think?

mansakondo:
I've been thinking about extracting the logic in another component called `Paywalled::CollapseComponent`. It might seem overkill but I think that this approach clarifies the logic by embracing the relationship between `PaywalledComponent` and `CollapseControlComponent` instead of trying to avoid it to avoid coupling. This way, we can have clarity without polluting the logic of `PaywalledComponent` and `CollapseControlComponent`.

joemasilotti:
Hmm. Hard to say without seeing the code to be honest. It still feels like we can get by with a custom slot but I'm open to ideas.

mansakondo:
> Hmm. Hard to say without seeing the code to be honest. It still feels like we can get by with a custom slot but I'm open to ideas.

I pushed the code at the same moment you asked for it :)

joemasilotti:
Ha, nice!

Digging in, I don't think this actually solves the problem of duplicated HTML. Looking at `app/components/users/paywalled/collapse_control_component.html.erb` and `app/components/collapse_control_component.html.erb` they look 90% the same.

I'm still set on using the approach I outlined above because it means we only render a collapse when the user can actually collapse. Let me know if you'd like to give it a go!

mansakondo:
I'm not sure how abstract the slot can be if we DRY this too much. If we include too much in the component the slot will become more concrete. If we don't include enough there will be duplication.

So, my question is should the slot be abstract or concrete? If that makes sense. 

Or I'm maybe I'm overthinking 🤔

mansakondo:
Unless I'm missing something, there seems to be a problem with this approach. Though we can check the presence of the slot with a predicate, the slot will still be part of the content. So, the slot will be rendered with the collapse control when the user has an active subscription.

joemasilotti:
I just pushed a POC that better represents what I was trying to say. Code is always better than hand-wavy comments, right!

Here, `Users::Paywalled::CollapseControlComponent` doesn't depend on anything. It can then be referenced _from_ the original `PaywalledComponent`. Then, in the containing view we can render with `component.custom_paywall` which decouples the components from each other.

Does this make sense? If so, please note that I only applied the full implementation to the desktop version of one of the filters. The other three might not be fully complete. Also, there's a non-localized string that I'd love to get rid of.

Let me know if you' like to resume work and take this over the finish line! And thank you for the back-and-forth. I think it's helped hone in on the right code structure for this feature.

mansakondo:
> I just pushed a POC that better represents what I was trying to say. Code is always better than hand-wavy comments, right!
> 
> Here, `Users::Paywalled::CollapseControlComponent` doesn't depend on anything. It can then be referenced _from_ the original `PaywalledComponent`. Then, in the containing view we can render with `component.custom_paywall` which decouples the components from each other.
> 
> Does this make sense? If so, please note that I only applied the full implementation to the desktop version of one of the filters. The other three might not be fully complete. Also, there's a non-localized string that I'd love to get rid of.
> 
> Let me know if you' like to resume work and take this over the finish line! And thank you for the back-and-forth. I think it's helped hone in on the right code structure for this feature.

That's way clearer now! I like how you refactor this!

I'm still committed to resuming work on this PR 👍🏾 Thank you for your patience!

joemasilotti:
Excellent! Glad to hear you are aligned and have appetite. Let me know when it's ready for re-review.

mansakondo:
Though you told me that you only implement one filter, it seems like they were all implemented when I looked at it. I only had to make a few tweaks, so I think it's ready.

app/components/collapse_control_component.html.erb
```diff
@@ -1,7 +1,9 @@
-<%= tag.button type: "button", data: {action: "toggle#toggle accessibility#toggleAriaExpanded", accessibility_target: "button"}, "aria-expanded": !collapsed?, class: "px-2 py-3 w-full flex items-center justify-between text-gray-400 hover:text-gray-500 text-sm" do %>
```
joemasilotti, line 1: This change can be reverted, we don't need the wrapping `flex` anymore.

app/components/developers/query_component.rb
```diff
@@ -78,5 +78,9 @@ def collapse_timezone?
     def utc_offsets
       Location.order(:utc_offset).distinct.pluck(:utc_offset)
     end
+
+    def customer?
+      user&.permissions&.active_subscription?
+    end
```
joemasilotti, line 84: This is unused and can be removed.

test/components/users/paywalled/collapse_control_component_test.rb
```diff
@@ -0,0 +1,12 @@
...
+require "test_helper"
+
+class Users::Paywalled::CollapseControlComponentTest < ViewComponent::TestCase
+  def test_component_renders_something_useful
+    # assert_equal(
+    #   %(<span>Hello, components!</span>),
+    #   render_inline(Users::Paywalled::CollapseControlComponent.new(message: "Hello, components!")).css("span").to_html
+    # )
+  end
+end
```
joemasilotti, line 12: This file can be removed.

joemasilotti (changes requested):
Looks great! Thanks for taking the time to review again.

I made three small requests which mostly involve deleting code. Once those are done this is ready for merge!

mansakondo:
I made the changes you requested. I had to update the styles when removed the unnecessary wrappers. So, let me know if you don't approve the changes.

joemasilotti:
Excellent! This looks great – thank you again.

PR #678, opened 2022-09-30
Set pagination page size to 10 (from 20) for /developers #650

xmarlox:
## Pull request checklist

### DIFF
--- a/test/support/helpers/stub_helper.rb
+++ b/test/support/helpers/stub_helper.rb
@@ -13,7 +13,6 @@ module StubHelper
       silence_warnings { object.send(:const_set, name, val) }
 
       yield
-
     ensure
       if original_values.key?(name)
         silence_warnings { object.send(:const_set, name, original_values[name]) }
--- a/test/support/helpers/stub_helper.rb
+++ b/test/support/helpers/stub_helper.rb
@@ -1,0 +1,25 @@
+module StubHelper
+  extend ActiveSupport::Concern
+
+  included do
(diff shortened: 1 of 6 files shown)


joemasilotti (changes requested):
Thanks for your contribution! I made a few requests in the comments. Let me know if you still have appetite to finish the PR – I know I took a while to review it.

xmarlox:
@joemasilotti just addressed your PR comments. let me know if there still need to change.
thanks! 😃

test/support/helpers/stub_helper.rb
```diff
@@ -0,0 +1,24 @@
+module StubHelper
+  extend ActiveSupport::Concern
+
+  included do
+    def stub_const(object, name, val = nil, &block)
```
joemasilotti, line 5: Wow, crazy that this isn't built in to Minitest! Did you explore any ways to do this without these 24 lines of code? Not that they aren't great, but it's a lot "just" to stub a constant.

Maybe we can refactor how the constant is defined or something?

kstratis: Hi, my 2 cents: 

* It seems there isn't such a method in minitest. Batteries are only included in RSpec 😉
* This helper works. However you can always swap it with [minitest-stub-const](https://rubygems.org/gems/minitest-stub-const)
* Since this is a concern, the `included` block is not really needed.

joemasilotti:
Closed via #733.

PR #679, opened 2022-10-06
Fix test settings for Rubymine

jkostolansky:
Using `minitest-reporters` with RubyMine is causing this issue:

> Current implementation of IntelliJ Minitest support conflicts with Minitest::Reporters. Please remove Minitest::Reporters.use! from your test code or suppress the invocation if ENV['RM_INFO'] is defined, then re-run your tests.

More information can be found in this [RubyMine blog post](https://blog.jetbrains.com/ruby/2021/04/improved-minitest-support-action-required/).

I don't love polluting the code with such conditional statements, but I don't know about a better solution and the RubyMine IDE is pretty popular, so I'm proposing this change for now.


joemasilotti:
Can you post screenshots of before and after this PR? I don't use RubyMine myself.

jkostolansky:
Sure. Before:

CleanShot 2022-10-17 at 23 59 31@2x

After:

CleanShot 2022-10-18 at 00 00 24@2x

Issue #621, opened 2022-08-30
Avatars won't render @2x on desktop Safari


joemasilotti:

jkostolansky:
Just to let you know, I'm investigating this issue. 🧐

PR #680, opened 2022-10-07
Bring back 1x and 2x avatars and remove lazy loading (#621)

jkostolansky:
I've reverted [this commit](https://github.com/joemasilotti/railsdevs.com/commit/d1f7cd3060345f9d15bf054f89138d7ff257103e) and fixed the media query. The original code used `or` operator:

```
"(-webkit-min-device-pixel-ratio: 2) or (min-resolution: 2dppx)"
```

But according to the [documentation](https://www.w3schools.com/tags/att_source_media.asp) the correct operator is `,` (comma):

```
"(-webkit-min-device-pixel-ratio: 2), (min-resolution: 2dppx)"
```

This solves the first issue - Safari is now correctly using the 2x image variant instead of 1x.

---

However, it doesn't solve the previous issue where a navigation in the app doesn't load the correct image, as has been mentioned in https://github.com/joemasilotti/railsdevs.com/pull/619.

I can reproduce this second problem even on the current deployed version of railsdevs. It's not related to Turbo, because turning Turbo off doesn't solve the issue. It's not related to the `<source srcset="...">` attribute, because it's causing the same problem even with the `` version.

https://user-images.githubusercontent.com/3950656/194542922-d8b2f395-aae5-45f5-ae03-20a3ed7c41da.mp4

From what I can tell, it's a Safari bug. It's related to the Lazy loading experimental feature in Safari. After turning this feature off in Safari (Develop > Experimental Features > Lazy image load), it works correctly. It has been introduced in [Safari 15.4](https://developer.apple.com/documentation/safari-release-notes/safari-15_4-release-notes).

Another way how to "fix" it is by adding eager loading to the avatar image itself:

```ruby
<%= image_tag ..., loading: "eager" %>
```

Problematic seems to be the Safari lazy loading feature in combination with Safari caching and how ActiveStorage serves images (via a redirect). It's not a problem for asset images (stored in `assets/images` folder), even with the `srcset` definition.

Because railsdevs doesn't have a ton of images and the image lazy loading is causing troubles, I propose turning it off for now.

Related issue: #621


config/application.rb
```diff
@@ -48,8 +48,5 @@ class Application < Rails::Application
 
     # Use default language as fallback if translation is missing
     config.i18n.fallbacks = true
-
-    # Defer loading of images until it reaches a calculated distance from the viewport
-    config.action_view.image_loading = "lazy"
```
joemasilotti, line 53: Why did you remove this configuration? If I had to guess, this line is what is causing your PR to work in Safari.

joemasilotti:
Got it. Thanks for diving in and the detailed explanation!

I'm going to deploy this now and see how much the PageSpeed changes. I'll report back here when I take another reading.

### PageSpeed (mobile)

* `https://railsdevs.com` - 98, 90, 97
* `https://railsdevs.com/developers` - 70, 74, 86
* `https://railsdevs.com/developers/1` - 72, 69, 71

jkostolansky:
Thanks for merging this PR! 👍

Not much experience with ActiveStorage in general, I prefer [Shrine](https://shrinerb.com/), but (hopefully) I can try to find some time and take a look at it. No promises, though, so if you want it sooner, you can create the issues.

joemasilotti:
No rush, feel free to get to it when you can. Thank you again for the help!

PR #687, opened 2022-10-11
RailsDevs French Version

5andu:
## Pull request checklist


config/application.rb
```diff
@@ -41,7 +41,7 @@ class Application < Rails::Application
     config.i18n.load_path += Dir[Rails.root.join("config", "locales", "**", "*.{rb,yml}")]
 
     # Permitted locales available for the application
-    config.i18n.available_locales = [:en, :es, :"zh-TW", :"pt-BR"]
```
joemasilotti, line 44: `fr:` needs to be `:fr` here.

app/components/time_component.html.erb
```diff
@@ -1,3 +1,3 @@
 <%= time_tag time do %>
-  <%= time_ago_in_words time %> ago
```
joemasilotti, line 2: "ago" still needs to appear in the UI.

joemasilotti (changes requested):
Thanks for this! I requested two changes and then we should be good to merge.

app/components/time_component.html.erb
5andu: I made it so ago appears, but it will be replaced by "auparavant" in French :)

5andu:
Done 👌

(I made "ago" appear, but I replaced it by "auparavant" in French)

Le sam. 15 oct. 2022 à 16:15, Joe Masilotti ***@***.***> a
écrit :

joemasilotti:
Thanks so much for this!

PR #701, opened 2022-10-20
Multiple Tailwind CSS configs (for the redesign)

### LINKED ISSUE #690: Split Tailwind CSS into two configs
To ensure the existing design doesn't conflict with the redesign, we can use [multiple Tailwind CSS config files](https://tailwindcss.com/blog/tailwindcss-v3-2#multiple-config-files-in-one-project-using-config) switched on via a feature flag.

The new config will contain the new colors and fonts.

joemasilotti:
This PR closes #690 by introducing a second Tailwind CSS configuration for the redesign. It also adds the branded colors: navy, salmon, teal, and green!

image

I went with "light" vs. "dark" for color names instead of numbers because there aren't 9 of each color. And I didn't want to have to guess if the "light" version was 200 or 100 for different colors.

To enable the redesign set the `REDESIGN` environment variable. You can do this when running the server locally via:

```bash
$ ENABLE_REDESIGN=1 bin/dev
```

Note that none of these colors are actually being used yet. So when enabling the redesign the site will actually look _less_ colorful than before! (Because red, yellow, blue, etc. no longer exist according to Tailwind.)

## Pull request checklist

### DIFF
--- a/Procfile.dev
+++ b/Procfile.dev
@@ -2,4 +2,5 @@ web: bin/rails server -p 3000
 worker: bundle exec sidekiq
 js: yarn build --watch
 css: yarn build:css --watch
+redesign: yarn build:redesign --watch
 stripe: stripe listen --forward-to localhost:3000/pay/webhooks/stripe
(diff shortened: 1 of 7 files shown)


joemasilotti:
I still have to figure out if Heroku will automatically pick up the new yarn script `build:redesign` or if I have to tell it to do so. Otherwise, even if we enable the redesign in production the stylesheet won't exist.

Issue #703, opened 2022-10-21
Ensure Heroku builds `css:redesign` when deploying


joemasilotti:
#701 closed #690 but it looks like Heroku isn't picking up on the new `css:redesign` yarn script.

@afomera [noted on Twitter](https://twitter.com/afomera/status/1583493594647625734?s=20) that cssbundling-rails and jsbundling-rails add custom Rake files that tell Heroku to run `build` and `build:css`. So maybe we need to do something similar?

joemasilotti:
Closed via #704.

PR #704, opened 2022-10-21
Tell Heroku to build the redesign CSS, too

joemasilotti:
Fix attempt for #703.

### DIFF
--- a/lib/tasks/redesign.rake
+++ b/lib/tasks/redesign.rake
@@ -1,0 +1,12 @@
+namespace :css do
+  desc "Build your CSS bundle for the redesign"
+  task :redesign do
+    unless system "yarn install && yarn build:redesign"
+      raise "Command css:redesign failed, ensure yarn is installed and `yarn build:redesign` runs without errors"
+    end
+  end
+end
+
+if Rake::Task.task_defined?("assets:precompile")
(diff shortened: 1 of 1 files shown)


lib/tasks/redesign.rake
```diff
@@ -0,0 +1,12 @@
+namespace :css do
+  desc "Build your CSS bundle for the redesign"
+  task :redesign do
+    unless system "yarn install && yarn build:redesign"
+      raise "Command css:redesign failed, ensure yarn is installed and `yarn build:redesign` runs without errors"
+    end
```
natematykiewicz, line 6: `system` is able to raise a more specific and verbose error if you pass `exception: true` to it.

https://rubyapi.org/3.1/o/kernel#method-i-system

```suggestion
  system "yarn install && yarn build:redesign", exception: true
```

Rails recently made a similar change to what I'm suggesting. https://github.com/rails/rails/commit/24fbbe06d214e587ece9e9292dcc9013f28d214f

joemasilotti: Good to know! Thanks for sharing.

I copy-pasted this from cssbundling-rails and jsbundling-rails. But for custom `system` calls I'll make sure to use `exception: true` in the future.

PR #710, opened 2022-10-22
Fix tailwind postcss setup #709

### LINKED ISSUE #709: Custom CSS isn't getting included
## Describe the bug

Custom CSS from `app/assets/stylesheets/tailwindcss/overrides.cs` and `app/assets/stylesheets/components/pagy.css` doesn't seem to be getting included in the output CSS file anymore.

## To reproduce

1. Checkout the latest code from `main`
2. Run `yarn build:css`
3. View `app/assets/builds/application.css`

## Expected behavior

`.pagy-nav` from `pagy.css` and `code::before` from `overrides.css` should both be present.

## Actual behavior

Neither custom CSS behavior is present in the output file.

## Additional context

This was most likely caused by one of the recent PRs related to the redesign. I'm guessing #700, #701, or maybe #708.

jkostolansky:
The `--postcss` parameter tells the Tailwind CLI to use our custom Postcss configuration.

The new `@config` directive doesn't work for some reason. However, specifying the config file directly as a CLI parameter seems to work.

I've also added the missing `text-sm` to the `redesign.config.js`, it's used by Pagy.

Fixes #709

### DIFF
--- a/app/assets/stylesheets/application.tailwind.css
+++ b/app/assets/stylesheets/application.tailwind.css
@@ -1,8 +1,6 @@
-@config "./../../../tailwind.config.js";
-
 @import "tailwindcss/base";
 @import "tailwindcss/components";
 @import "tailwindcss/utilities";
-@import "tailwindcss/overrides";
+@import "./tailwindcss/overrides.css";
 
-@import "components/pagy";
+@import "./components/pagy.css";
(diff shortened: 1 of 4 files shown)


joemasilotti:
Ah! Great catch. I removed the `--postcss` option to get `@config` working. Not realizing that we actually have a custom config.

PR #717, opened 2022-10-26
Github Actions changes

jkostolansky:
Changes in this PR:

- Removed the unnecessary `strategy.matrix` because there was only one node version specified.
- Updated versions of some older actions (`checkout`, `cache`, `setup-node`, `codeql-action`).
- Updated `node` to `16` to match the version in the `.node-version` file.
- Updated `postgres` to `13` because this version is mentioned in the `README` file. If you are using something else in production, let me know.
- Added the argument `--frozen-lockfile` to the `yarn install` ([docs](https://classic.yarnpkg.com/lang/en/docs/cli/install/#toc-yarn-install-frozen-lockfile)).
- Split the `Run tests` step into multiple smaller steps (`Build assets`, `Prepare database`, `Run tests`).
- Joined the two jobs into one because they were not running in parallel anyway. This should remove some overhead.

If there was any reason for the previous state and/or you want to keep it as it was (e.g. two jobs), just let me know.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -2,9 +2,9 @@ name: CI
 
 on:
   pull_request:
-    branches: [ "*" ]
+    branches: ["*"]
   push:
-    branches: [ main ]
+    branches: [main]
 
 concurrency: ci-${{ github.ref }}
 
@@ -18,7 +18,7 @@ jobs:
     services:
       postgres:
         image: postgres:13
-        ports: [ "5432:5432" ]
+        ports: ["5432:5432"]
         env:
           POSTGRES_USER: postgres
           POSTGRES_PASSWORD: password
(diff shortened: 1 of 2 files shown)


joemasilotti (commented):
Thanks for this! I'm not super up to date on GitHub Actions so I appreciate someone else telling me what to do. :D

Two small requests to match the style of Ruby code in the repo:

1. Can you remove the spaces inside of brackets (`[]`)?
2. Can you double quote strings (instead of singe quotes)?

jkostolansky:
Sure! Done. There are some Github expressions still using single quotes, because:

> You must use single quotes (`'`) around the string. Wrapping with double quotes (`"`) will throw an error.

Source: https://docs.github.com/en/actions/learn-github-actions/expressions#literals

joemasilotti:
Got it, thanks!

Issue #647, opened 2022-09-15
Publicly shareable URLs polish


joemasilotti:
The goal of this issue is to clean up the UX around #646.

A PR addressing this issue should:

- [ ] When the “Share” button is clicked show a modal explaining how the public page works
- [ ] A Cancel buttons closes the modal
- [ ] A “Copy link” button copies the link to the clipboard

~This issue is blocked by #646.~ This issue is no longer blocked.

Reminder to include a disclaimer that the developer's email address will be shared if they share their profile via this method.

sarvaiyanidhi:
Hi @joemasilotti I would like to work further on this issue.. can you please assign it to me?

joemasilotti:
Sure thing! All yours @sarvaiyanidhi. Thanks again for all the help. :)

joemasilotti:
Closed via #718.

PR #718, opened 2022-10-27
Publicly shareable URLs polish 

sarvaiyanidhi:
Implemented modal to share public URL of the developer. More details in issue #647 

## Pull request checklist

- [ ]  ~~My code contains tests covering the code I modified~~


sarvaiyanidhi:
Hi @joemasilotti Have implemented basic functionality but only thing pending is to close modal on click of copy link. Please have a look unto implementation and let me know if you see any other changes to be done.

app/controllers/public_profiles_controller.rb
```diff
@@ -0,0 +1,14 @@
...
+
+  def new
+    @developer = developer
+  end
+
+  private
+
+  def developer
+    finder = Developers::Finder.new(id: params[:developer_id])
+    finder.developer
```
joemasilotti, line 12: We can combine these lines.

```suggestion
    Developers::Finder.new(id: params[:developer_id]).developer
```

app/controllers/public_profiles_controller.rb
```diff
@@ -0,0 +1,14 @@
+class PublicProfilesController < ApplicationController
+  before_action :authenticate_user!
+
+  def new
+    @developer = developer
```
joemasilotti, line 5: We also need to authorize that the current user owns the developer profile. We can create a new Pundit policy in `app/policies` or use the existing one but pass in `:update?` as a custom method. Your call!

app/views/public_profiles/new.html.erb
```diff
@@ -0,0 +1,39 @@
...
+        <button data-action="clipboard#copy clipboard#toggle" type="button" class="w-full inline-flex justify-center rounded-md border border-transparent shadow-sm px-4 py-2 bg-gray-600 text-base font-medium text-white hover:bg-gray-700 hover:cursor-pointer focus:outline-none focus:ring-2 focus:ring-offset-2 focus:ring-gray-500 sm:ml-3 sm:w-auto sm:text-sm">
+          <%= inline_svg_tag "icons/solid/document-duplicate.svg", class: "-ml-1 mr-2 h-5 w-5", data: {clipboard_target: "toggleable"}, aria_hidden: true %>
+          <%= inline_svg_tag "icons/solid/check.svg", class: "-ml-1 mr-2 h-5 w-5 hidden", data: {clipboard_target: "toggleable"}, aria_hidden: true %>
+          <%= t(".copy_link") %>
+        </button>
+            <% end %>
+             <%= link_to t(".cancel"), developer_path(@developer), class: "mt-3 w-full inline-flex justify-center rounded-md border border-gray-300 shadow-sm px-4 py-2 bg-white text-base font-medium text-gray-700 hover:bg-gray-50 focus:outline-none focus:ring-2 focus:ring-offset-2 focus:ring-gray-500 sm:mt-0 sm:w-auto sm:text-sm" %>
+        </div>
```
joemasilotti, line 32: It looks like the indentation is off a bit here.

joemasilotti, line 32: It looks like the indentation is off a bit here.

app/views/public_profiles/new.html.erb
```diff
@@ -0,0 +1,39 @@
...
+        <div class="mt-8 text-center sm:ml-10 sm:text-left">
+          <p class="text-sm text-gray-400 inline-flex"><%= inline_svg_tag "icons/outline/information_circle.svg", class: "-ml-1 mr-2 h-5 w-5" %><%= t(".disclaimer") %></p>
+        </div>
```
joemasilotti, line 35: Let's remove the disclaimer. I updated the copy to make it clear.

joemasilotti (commented):
RailsDevs hasn't updated to Heroicons v2, please use [v1](https://v1.heroicons.com) for the share icon. Also, please use the "solid" version, which is recommended for buttons.

Other than that this is looking great! I made a few requests/comments and pushed one commit. Let me know when the "close on submit" is done and I can re-review.

sarvaiyanidhi:
Thank you for the feedback. Will complete pending changes and update code as per suggestion soon.

app/controllers/public_profiles_controller.rb
sarvaiyanidhi: I have updated existing developer_policy.rb file with new method share_profile? and authorizing in controller. Not sure about passing update? custom method mentioned in above comment.. Can you confirm if current implementation looks fine?

joemasilotti (changes requested):
This is looking awesome! A few small requests and we should be good to merge.

1. Let's delay closing the modal after copying to clipboard. That will let the checkmark appear and make it obvious that copying "worked".
2. Let's move the modal closing to a new Stimulus controller. It doesn't feel right in the clipboard one. This also allows us to move the delay to a [defaulted `value`](https://stimulus.hotwired.dev/reference/values#default-values) on a new controller.
3. Please add tests for the new `DeveloperPolicy` method.

Finally, I would like the "Hire me" button when viewing a valid public profile to link to emailing the developer. Would you like to add that to this PR, too? If not, I can create a new issue/PR for it.

joemasilotti:
Sounds great all around! Looking forward to it :D

sarvaiyanidhi:
@joemasilotti PR is ready for review again. Let me know if it is good to merge or need any more changes. Thanks

joemasilotti (approved):

joemasilotti:
I made a few tiny tweaks but everything else is great! Merging this in now – thanks for your help on these issues.

PR #670, opened 2022-09-21
Making sidebar of /developers sticky on Desktop

5andu:
## Pull request checklist

- [x] My code contains tests covering the code I modified
- [x] I linted and tested the project with `bin/check`
- [x] I added significant changes and product updates to the [changelog](CHANGELOG.md) 
👉 I don't consider this as a significant change (this is my first commit on an open source project 🙌 )

PR #725, opened 2022-11-21
Revert "Making sidebar of /developers sticky on Desktop"

joemasilotti:
Reverts joemasilotti/railsdevs.com#670 from @5andu.

Unfortunately when the sidebar is taller than the screen you can't get to the submit button without scrolling the entire developer list.


5andu:
Yes my bad, I just realized it too! :)

Issue #726, opened 2022-11-22
Update to Pay v6.0.4


joemasilotti:
Currently, there is some gnarly code to work around paused subscriptions. My mental model didn't line up with Stripe's when I first allowed folks to "pause" their subscription.

I thought pausing meant you should _not_ have access to the product. Stripe thinks the opposite. (I'm not 100% sure why this is the case, but this isn't the time for that.)

Today Pay released [v6.0.0](https://github.com/pay-rails/pay/releases/tag/v6.0.0) (and v6.0.4) which consolidates all the pausing behavior. Updating to this version and reworking `Businesses::Permission` will help clean up the existing (and no longer needed) code.

A PR addressing this issue will:

- [ ] Update to Pay v6.0.4
- [ ] Remove the custom "paused" logic from `Businesses::Permission`
- [ ] Ensure that:
    - [ ] A paused subscription has access until their current, active billing period ends
    - [ ] A paused subscription loses access after their active billing period ends
    - [ ] An active subscription has access
    - [ ] An inactive subscription does not have access

sarvaiyanidhi:
Hi @joemasilotti I would like to work on this issue.. Can you please assign it to me?

sarvaiyanidhi:
Hi @joemasilotti I have updated Pay to v6.0.3 and also added related migration for it.

In Progress - https://github.com/joemasilotti/railsdevs.com/compare/main...sarvaiyanidhi:railsdevs.com:update-pay-6.0.3

Now I see that if we pause subscription it adds current_period_start and current_period_end value in database which I understand needs to be referred to keep access active till current_period_end date.

I was accordingly trying to clean existing code in this file https://github.com/joemasilotti/railsdevs.com/blob/main/app/models/businesses/permission.rb#L34 by removing reject pause logic.

What I understood that active_or_paused scope present in this file https://github.com/pay-rails/pay/blob/master/app/models/pay/subscription.rb will automatically handle and checks current_period_end but I don't see it been checked in the scope.

So do we need to add logic in permission file to check current_period_end or am I missing anything?

joemasilotti:
Thanks for diving in!

It looks like those fields are being used in the Stripe subscription model, not the generic one. Ideally, with this we don't need any custom "paused" logic because it all exists in Pay.

https://github.com/pay-rails/pay/blob/master/lib/pay/stripe/subscription.rb

---

[...]

PR #735, opened 2022-11-29
Update Pay Gem and clean pause logic for the subscriptions

sarvaiyanidhi:
This PR updates Pay gem to 6.0.3 and Stripe gem to 8.0.0. It also cleans existing logic of rejecting pause subscriptions as it is now handled in new version to keep subscriptions till current period ends. More details in issue #726 

## Pull request checklist

## Deploy checklist
- [ ] Merge and deploy this PR
- [ ]  Run heroku run rake sync:subscriptions -a rails-devs


db/schema.rb
```diff
@@ -10,9 +10,8 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.0].define(version: 2022_11_24_025600) do
+ActiveRecord::Schema[7.0].define(version: 2022_11_26_042109) do
   # These are extensions that must be enabled in order to support this database
-  enable_extension "pg_stat_statements"
```
joemasilotti, line 15: Please add this line back in.

test/models/businesses/permission_test.rb
```diff
@@ -13,13 +13,20 @@ class Businesses::PermissionTest < ActiveSupport::TestCase
...
     refute permission.active_subscription?
   end
 
+  test "active_subscription? if paused and current_period_end has grace period" do
+    customer = pay_customers(:one)
+    pay_subscriptions(:full_time).update!(pause_behavior: "void", pause_resumes_at: nil, current_period_end: 1.day.from_now, pause_starts_at: 1.day.from_now)
+    permission = Businesses::Permission.new(customer.subscriptions)
+    assert permission.active_subscription?
+  end
+
```
joemasilotti, line 29: I think we need one more test for the following:

```ruby
pay_subscriptions(:full_time).update!(pause_behavior: "void", pause_resumes_at: nil, current_period_end: 1.day.from_now, pause_starts_at: 1.week.ago)
permission = Businesses::Permission.new(customer.subscriptions)
assert permission.active_subscription?
```

If I understand correctly, this is an important test. It makes sure someone who paused (1 week ago) but is still on the grace period (expiring tomorrow) still has an active subscription.

db/schema.rb
sarvaiyanidhi: Sorry not sure how it got removed. Have added back.

test/models/businesses/permission_test.rb
sarvaiyanidhi: As far as I understand if current_period_end is 1 day from now so pause_stars_at will also be automatically set as same so it will also be 1 day from now and subscription access will be paused accordingly.

If we check here pause_start_date gets sets as per current_period_end date https://github.com/pay-rails/pay/blob/master/test/pay/stripe/subscription_test.rb#L219

so I think above mentioned test is conflicting if we set current_period_end from 1 day from now and pause_starts_date to one week before.

Also, when I paused ongoing subscription from stripe portal I observed that puase_starts_date in database was set same as that of current_period_end.

Let me know your thoughts if I am assuming it correctly.

joemasilotti:
Thanks for this @sarvaiyanidhi! I'm planning on diving in tomorrow. I will update the CHANGELOG accordingly.

Can you please update to Pay 6.1.0 that was just released? If not, no worries, I can do it.

sarvaiyanidhi:
Sure... Will update and push the changes.

joemasilotti:
Digging in I'm confused about one scenario. Their period ends tomorrow but the pause started yesterday. I _think_ this accurately represents a grace period – e.g. someone paused but still has time left in their paid subscription.

```ruby
pay_subscriptions(:full_time).update!(pause_behavior: "void", pause_resumes_at: nil, current_period_end: 1.day.from_now, pause_starts_at: 1.day.ago)
permission = Businesses::Permission.new(customer.subscriptions)
assert permission.active_subscription?
```

I added this test and it is failing. But from my understanding it should be passing.

Am I missing something or is this a bug?

sarvaiyanidhi:
> ```ruby
> pay_subscriptions(:full_time).update!(pause_behavior: "void", pause_resumes_at: nil, current_period_end: 1.day.from_now, pause_starts_at: 1.day.ago)
> permission = Businesses::Permission.new(customer.subscriptions)
> assert permission.active_subscription?
> ```

Here I think pause_starts_at will also be 1.day.from_now as it sets as per current_period_end.

So, if you try Stripe portal in test mode for any existing account which has on going subscription and mark it as pause you will see that in database pause_starts_at will be set as per current_subscription_end data and not today's date.

joemasilotti:
Ah, so the scenario I created isn't possible! Good to know, thanks again for diving in.

I have a test subscription that will "expire" tomorrow but is paused. Once I can test that I'll merge in this PR.

sarvaiyanidhi:
> I have a test subscription that will "expire" tomorrow but is paused. Once I can test that I'll merge in this PR.

Just one query here, was this subscription paused with this new code in place as otherwise pause_starts_at field might not be set in database for the subscription and may not work as expected. Otherwise we have to sync subscription to set those values.

joemasilotti:
Good catch. I pulled down your PR, ran the migrations, then set up 4 new subscriptions to test. So I think it should be a valid run!

joemasilotti:
Everything is looking good – thank you again for this @sarvaiyanidhi! ❤️

PR #743, opened 2022-12-14
Revamp CTA Design with Blurred Out Developers

### LINKED ISSUE #653: Revamp CTA design
The goal of this issue is to improve the design (and increase conversion) of the search results CTA.

- [ ] Replace CTA design with a static list of developers blurred out
- [ ] Overlay the copy and button on top of this list

Note that this will probably require a few design iterations. If you're up for it, feel free to give it a shot!

This is blocked on #651.

izaguirrejoe:
## Pull request checklist

Closes #653. This PR overlays the CTA over a blurred list of static developers. 

Screenshot 2022-12-13 at 5 35 42 PM

### Static List of Developers
This PR uses `Developer.available.first(3)` for the backdrop of the CTA.

**Pros:**
- Simple
- Probably not going to drop below 3

**Cons:**
- Backdrop same for all visits to /developers
- User data in DOM
- Additional query on page

I've considered creating a mock `PaywalledDeveloper` class with fake data and using those instead, let me know what you think. 

### Possible accessibility issues
It can be a little difficult to read the text overlayed the blurred backdrop. I've added text shadow to help alleviate this. I played around with adding an additional background over only the text portion of the CTA, but the results weren't as clean. 

### Bump WebDrivers
I bumped Webdrivers gem to 5.2.0. The system tests wouldn't run on my M1 Mac unless I updated Webdrivers to 5.2.0. Please see https://github.com/SeleniumHQ/selenium/issues/11066#issuecomment-1265694239.

### Developers Test
Removed `count: 1` from developers integration test. It's possible for the same developer to appear twice on the page, once under the CTA.


joemasilotti:
> I've considered creating a mock `PaywalledDeveloper` class with fake data and using those instead, let me know what you think.

Let's go with that. There isn't any value in showing "real" developers (in my opinion) underneath the blur. We can then choose/create avatars that are simple and won't be obvious you are seeing the same three every time.

> I bumped Webdrivers gem to 5.2.0. The system tests wouldn't run on my M1 Mac unless I updated Webdrivers to 5.2.0. Please see https://github.com/SeleniumHQ/selenium/issues/11066#issuecomment-1265694239.

Can you pull that out into a different PR? Then I can merge it in sooner.

> Removed count: 1 from developers integration test. It's possible for the same developer to appear twice on the page, once under the CTA.

Showing a static blur should remedy this.

izaguirrejoe:
Will do! PR already opened to bump webdrivers.

izaguirrejoe:
Sorry for the delay. I ended up creating a `Developers::PaywalledDeveloper` class with a generator function to create fake Developers, with corresponding `Developers::PaywalledCardComponent` and tests.

I used the `Faker` gem to create fake bio and fake hero. I played around with using Faker to create fake avatars too, but they looked very cartoon-y. I instead made developer_avatar_urls from SeedsHelper public and using that instead, looks a lot cleaner. 

I made the `PaywalledCardComponent` always display Available Now to incentivize the CTA, could randomize it as well.

Let me know what you think.

joemasilotti:
This is looking good! But the font is still difficult to read and we definitely need to fix that. Can you share what you had for the solid background? Or any other ideas here?

Also, it looks like I can sometimes get 2 or 3 of the same exact profile photo. I'm also a bit worried that if the services goes down there will be no profile photos.

This is making me think we want to go back to 3 static images that appear for every search result. What do you think?

izaguirrejoe:
Sounds good, I'll switch to 3 static profile pic. I'll push the update tomorrow. 👍🏽

izaguirrejoe:
I've done the following:

- Added 3 paywalled developer images to assets/paywalled_developers
- Modified the PaywalledDeveloper generator function to use these images, shuffled. The generator can take a number greater than 3, but it starts to cycle through the images. I'm keeping the generator in case someone wants to go further in developing the PaywalledDeveloper class, for now it's used in only developers/index.html.erb with `generate(3)`.
- Undid changes to SeedsHelper and removed references in PaywalledDeveloper
- Update tests

I haven't yet pushed the branch with the added background underneath the CTA test. Here's a screenshot of what it looks like with bg-white/50. Can be adjusted to be more/less transparent.
Screenshot 2023-01-04 at 1 17 30 PM

joemasilotti:
This is looking great! Nice work.

A few more requests before I dive into the code again:

1. Can you ensure that there is a broader representation of people in the three photos?
2. Let's add a bit more padding to the white background and round the corners.
3. As long as the `3` is defaulted then having `#generate` take a param sounds good to me.

izaguirrejoe:
1. I've added 5 more photos to ensure a broader representation of people, for a total of 8 profiles.
2. I've added padding and rounded corners. I played around with adding a border as well, another option. The commit I just pushed is borderless.
3. The default for `generate` is 1, which allows for `PaywalledDeveloper.generate`, feels a little more natural. Now that the number of photos is up to 8 I don't think cycling over 9+ PaywalledDevelopers will be needed. If so, we can continue to add more photos until the desired limit.

joemasilotti:
Hey @izaguirrejoe, can you please allow maintainers to add commits to your branch? I have some design changes I'd like to get in before merging. Thanks!

izaguirrejoe:
I believe "Allow edits by maintainers" was already checked. I unchecked it and checked it again to be sure.

joemasilotti:
Awesome, thanks! Looks like it is working now.

I ended up reverting back to showing 3 random "real" developers instead of the seeded ones. Nothing was wrong with your code. But using seeded devs required way too much scaffolding for me to feel comfortable shipping it.

I tweaked the design a bit, too. Thanks again for your contribution!

image

izaguirrejoe:
Looks good! 👍🏽

PR #748, opened 2022-12-31
Randomize developers on home page

jbennett:
This randomizes the developers who are added to the home page, but caches them so that it stay stable for 1 day. This is coming from discussion #721.

I don't currently see any caching being using in the app, but you do have Sidekiq. If you like this solution, are you wanting to using Redis in production?

### DIFF
--- a/app/controllers/home_controller.rb
+++ b/app/controllers/home_controller.rb
@@ -1,9 +1,19 @@
 class HomeController < ApplicationController
   def show
-    @developers = Developer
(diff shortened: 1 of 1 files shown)


joemasilotti:
Thanks for the PR @jbennett! I'm going to close this for now as I don't think that randomizing developers is the right approach. Even if they are cached.

But if I do decide to go that route I'll make sure to re-open this. Thanks again!

PR #751, opened 2023-01-17
Check for deleted developers in conversation

### LINKED ISSUE #747: Admins viewing conversations with a deleted developer raise an error
## Describe the bug

Folks on RailsDevs are able to delete their account. If you view a conversation with that person the `User` is nil and it raises an error.

## To reproduce

1. Reseed your local database via `bin/rails db:seed`
2. Delete the first developer in the console via `Developer.first.destroy`
3. Sign in as an admin via `admin@example.com` and `password`
4. View a conversation including that developer from the admin dropdown
 
## Expected behavior

The business's conversations and individual conversation appear without issue.

## Actual behavior

[An exception is raised](https://app.honeybadger.io/fault/93997/18b7f61b0e6f84863bc85a2fc3660c02).

## Additional context

Two ways we could potentially fix this issue:

1. Cascading deletions for the user and deleting all of their conversations, like notifications.
2. Making the `User` a "tombstone" and showing "Deleted account" with a generic avatar for the conversations.

I'm open to either depending on your appetite!

troyizzle:
I haven't wrote some test but wanted to make sure this is what you were looking to implement.
Let me know what you think @joemasilotti 
4DJyP2l
e3XEBtA

## Pull request checklist

Closes #747


joemasilotti:
Pretty much, yeah! Let's change the copy to `(deleted account)`.

I am also pretty sure the messages and conversations will cascade delete. (Can you double check that for me?)

If that's correct then you won't be able to click into a conversation, so the link will need to be removed in a list.

troyizzle:
I don't believe the messages get deleted, just the notifications.


Let me know what you think!

joemasilotti:
You are correct! Messages/conversations are _not_ deleted when the `User` is deleted.

Please proceed with implementing the change request above and adding some tests. Thank you!

troyizzle:
> You are correct! Messages/conversations are _not_ deleted when the `User` is deleted.
> 
> Please proceed with implementing the change request above and adding some tests. Thank you!

I think this should be ready now!

app/models/conversation.rb
```diff
@@ -18,6 +18,10 @@ def self.find_by_inbound_email_token!(token)
     where("lower(inbound_email_token) = ?", token).first!
   end
 
+  def deleted_business_or_developer?
+    developer.nil? | business.nil?
```
joemasilotti, line 22:
```suggestion
    developer.nil? || business.nil?
```

...right?

joemasilotti (changes requested):

joemasilotti:
If I'm a business and view a conversation with a deleted developer it still says "Conversation with #{business}" in the <h1> and <title>.

troyizzle:
Thanks Joe!
Should the `deleted_account` i18n key be moved into something more common, like at the root, its being used a couple different times.

Hope this fixes that title problem
image

joemasilotti:
That would be great! My rule is if it’s used in 3 or more places to extract it to a common area.

joemasilotti (changes requested):
Almost there! The `<title>` tag on a conversation still needs to be updated.

joemasilotti:
Awesome, thank you! Going to merge in now.

There might be a refactor down the line with the repeated check... if you have appetite feel free to give it a pass. But this looks good for now.

PR #754, opened 2023-01-19
Focus in mobile filters

DamonClark:
This PR is a first attempt at fixing: https://github.com/joemasilotti/railsdevs.com/issues/541

**Resolved:** When "Filters" button focused, screen readers should not announce "collapsed"
**In Progress:** Focus should be trapped inside modal while modal is open (should not move back to main page)

I ran  `bin/check` but it throws an error while it was in progress: (seems like it didn't want to complete the last of the tests)
`objc[71328]: +[__NSCFConstantString initialize] may have been in progress in another thread when fork() was called.`

My modifications were adding javascript file and adding a class to a few `.html.erb` files. I can add tests if needed, but may need some direction on where they should be included. 

Didn't update changelog yet but I can if requested. 

I created a draft pull request, since this is definitely not ready yet.

## Pull request checklist


joemasilotti:
I don't think that test failure is related to your changes. That looks like Objective-C code to me! Regardless, I wouldn't worry about tests for this improvement.

Looking at the code, I have two questions/requests:

1. Do we _need_ to have all of those classes added? Can we add it once on the outer form? Or target the child elements of a specific form in the Stimulus controller? It feels like a lot of additional changes - and ones that might be forgotten if we add a new filter.
2. I'd prefer using the Stimulus target API instead of class names if we can get that to work.

DamonClark:
@joemasilotti yea I'm struggling to get anything to work on this. It just seems like it won't initiate the logic in my stimulus controller not matter what I try. I really wish I could figure this on out on my own, but I may need som assistance on this one.

joemasilotti:
Thanks for the update! I'm going to close this out for now. I spent some time diving in and I'm not happy with my implementation.

Issue #759, opened 2023-01-26
Filter by developer badges


joemasilotti:
Developers can now have up to two badges on their profile: recently active and source contributor. It makes sense to be able to filter by these when searching.

A PR addressing this issue should:

- [ ] Add a new filter option with a checkbox for each badge

The new filter option should be placed in between the search and role level filters and match the UI and UX of the other checkboxes. It should initially be expanded.

Two implementation details to watch out for:

* Make sure the filtering works across desktop and mobile.
* Make sure clicking the label toggles the associated checkbox on desktop and mobile.

Also, I'm not interested in caching the recently active status. This will occur in #758.

sarvaiyanidhi:
Hi @joemasilotti I would like to work on this feature... Can you please assign it to me?

joemasilotti:
Sure thing - thank you!

joemasilotti:
Closed via #763.

PR #763, opened 2023-02-01
Filter by developer badges

sarvaiyanidhi:
Implementation for filtering developers with recently_active and source_contributor badges.. More details in issue #759 

Please note that badges list is right now defined in developer model but have added todo to move to badge model later on.

## Pull request checklist

### DIFF
--- a/app/components/developers/query_component.html.erb
+++ b/app/components/developers/query_component.html.erb
@@ -11,6 +11,24 @@
       <% end %>
     </div>
 
+     <div data-controller="toggle accessibility" data-toggle-visibility-class="hidden">
+      <div class="border-t border-gray-200 py-6">
+        <h3 class="-mx-2 -my-3 flow-root">
+          <%= render CollapseControlComponent.new(t(".badge.title")) %>
+        </h3>
+        <div data-toggle-target="element">
+          <div class="space-y-4 pt-6">
+            <%= form.collection_check_boxes(:badges, badges, :first, :last, {include_hidden: false}) do |b| %>
+              <div class="flex items-center">
+                <%= b.check_box checked: badge_selected?(b.object), class: "h-4 w-4 border-gray-300 rounded text-gray-600 text-gray-600 focus:ring-gray-500" %>
+                <%= b.label class: "ml-3 text-gray-500 text-sm" %>
+              </div>
+            <% end %>
+          </div>
+        </div>
+      </div>
+    </div>
+
     <div data-controller="toggle accessibility" data-toggle-visibility-class="hidden">
       <div class="border-t border-gray-200 py-6">
         <h3 class="-mx-2 -my-3 flow-root">
--- a/app/components/developers/query_component.html.erb
+++ b/app/components/developers/query_component.html.erb
@@ -11,23 +11,25 @@
       <% end %>
     </div>
 
-     <div data-controller="toggle accessibility" data-toggle-visibility-class="hidden">
-      <div class="border-t border-gray-200 py-6">
-        <h3 class="-mx-2 -my-3 flow-root">
-          <%= render CollapseControlComponent.new(t(".badge.title")) %>
-        </h3>
-        <div data-toggle-target="element">
-          <div class="space-y-4 pt-6">
-            <%= form.collection_check_boxes(:badges, badges, :first, :last, {include_hidden: false}) do |b| %>
-              <div class="flex items-center">
-                <%= b.check_box checked: badge_selected?(b.object), class: "h-4 w-4 border-gray-300 rounded text-gray-600 text-gray-600 focus:ring-gray-500" %>
-                <%= b.label class: "ml-3 text-gray-500 text-sm" %>
-              </div>
-            <% end %>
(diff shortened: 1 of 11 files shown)


sarvaiyanidhi:
Filter UI in desktop and mobile view

Desktop

badge_filter_desktop

Mobile

badge_filter_mobile

app/components/developers/query_component.html.erb
```diff
@@ -11,6 +11,24 @@
       <% end %>
     </div>
 
+     <div data-controller="toggle accessibility" data-toggle-visibility-class="hidden">
```
joemasilotti, line 14:
```suggestion
    <div data-controller="toggle accessibility" data-toggle-visibility-class="hidden">
```

app/components/developers/query_mobile_component.html.erb
```diff
@@ -18,6 +18,22 @@
           <% end %>
         </div>
 
+         <div data-controller="toggle" data-toggle-visibility-class="hidden" class="border-t border-gray-200 px-4 py-6">
```
joemasilotti, line 21:
```suggestion
        <div data-controller="toggle" data-toggle-visibility-class="hidden" class="border-t border-gray-200 px-4 py-6">
```

app/models/developer.rb
```diff
@@ -12,6 +12,9 @@ class Developer < ApplicationRecord
   FEATURE_LENGTH = 1.week
   RECENTLY_ACTIVE_LENGTH = 1.week
 
+  # TODO: Move to Badge model once new model is in place
```
joemasilotti, line 15:
```suggestion
  # TODO #758: Cache developer badges to a new model 
```

app/models/developer.rb
```diff
@@ -62,6 +65,10 @@ class Developer < ApplicationRecord
     joins(:location).where(locations: {country: countries})
   end
 
+  scope :filter_by_recently_active, -> { where("developers.updated_at >= ?", RECENTLY_ACTIVE_LENGTH.ago) }
+
+  scope :filter_by_source_contributor, -> { where("source_contributor >= ?", true) }
+
   scope :actively_looking_or_open, -> { where(search_status: [:actively_looking, :open, nil]) }
   scope :available, -> { where(available_on: ..Time.current.to_date) }
   scope :available_first, -> { where.not(available_on: nil).order(:available_on) }
```
joemasilotti, line 74: Let's remove the empty lines and sort the group alphabetically.

app/models/developer.rb
```diff
@@ -62,6 +65,10 @@ class Developer < ApplicationRecord
     joins(:location).where(locations: {country: countries})
   end
 
+  scope :filter_by_recently_active, -> { where("developers.updated_at >= ?", RECENTLY_ACTIVE_LENGTH.ago) }
```
joemasilotti, line 68:
```suggestion
  scope :recently_active, -> { where("developers.updated_at >= ?", RECENTLY_ACTIVE_LENGTH.ago) }
```

app/models/developer.rb
```diff
@@ -62,6 +65,10 @@ class Developer < ApplicationRecord
...
+
+  scope :filter_by_source_contributor, -> { where("source_contributor >= ?", true) }
```
joemasilotti, line 70:
```suggestion
  scope :source_contributor, -> { where("source_contributor >= ?", true) }
```

app/queries/developer_query.rb
```diff
@@ -90,9 +96,22 @@ def empty_search?
       role_levels.empty? &&
       search_query.blank? &&
       countries.blank? &&
+      badges.blank? &&
       !include_not_interested
   end
 
+  def badges_filter_records
+    badges.map { |badge| send(badge) } if badges.any?
```
joemasilotti, line 104: `#send` feels a little dangerous here. Especially because it will be coming directly from user input.

Can we modify this to either a) not use `#send` or b) filter the input to guarantee it is either `recently_active` or `source_contributor` and nothing else?

joemasilotti (changes requested):
This is looking great! I made a few requests inline but they are mostly style requests.

There is one question on using `#send` that might require a bit of thinking, though.

Outside of these, can you please wrap the UI in a feature flag called `:badge_filter`? I'm thinking of adding some tooltips on each of these but want to get your code merged in first.

app/queries/developer_query.rb
sarvaiyanidhi: You are correct, given the second thought it is not good idea to use send for such input. Will modify to restrict to only two options

sarvaiyanidhi:
Hi Joe.. Have addressed all points as per feedback.  Let me know if it is good to go or need anymore changes. Thanks

joemasilotti:
Looking good! Let’s change the feature to be enabled everywhere but production.

sarvaiyanidhi:
@joemasilotti Have pushed the requested changes. Thanks

joemasilotti:
Looks great! I'm going to merge this in now.

Thanks again for the contribution, @sarvaiyanidhi!

joemasilotti (approved):

PR #773, opened 2023-02-05
Add association between referral and referrer

### LINKED ISSUE #736: Rework modeling of referral program
RailsDevs has a very basic referral program. Folks refer businesses to the site and 10% of any generated revenue is shared with the referrer. This includes monthly fees _and_ hiring fees.

Currently, I manually create a 6 character referral code for each person. Then they can apply that to URLs to get credit for a referral, e.g. `https://railsdevs.com?ref=ABCDEF`.

As it stands we can't associate a `Referral` record with a user. Which will make #728 a bit more difficult.

Let's rework some of the modeling to better match this diagram.

image

In text, this means that a `Referral` is still created when someone signs up with a `ref` URL param present. The `code` is still stored on the `Referral` model.

But while creating this record we should also check to see if the code matches any existing `ReferralCode` records. If so, associate them together.

`ReferralCode` records belong to a `User` and have a `code` string as well. This string cannot be changed. For now we can create the code manually in the database (e.g. no UI).

Note that a referral can occur without a `ReferralCode` or associated referrer. This means the relationship from `Referral` to `ReferralCode` is optional.

When creating a new `ReferralCode` we _might_ also want to see if any existing `Referral` objects need to be associated with it. But let's skip this for now – I can handle it manually as the program gets going. And I'd rather not have that code exist after it needs to.

PatrickMcSweeny:
Closes #736 

## Pull request checklist

- [X] My code contains tests covering the code I modified

### DIFF
--- a/app/models/concerns/referrable.rb
+++ b/app/models/concerns/referrable.rb
@@ -1,0 +1,14 @@
+module Referrable
+  extend ActiveSupport::Concern
+
+  included do
+    has_one :referral, dependent: :destroy, foreign_key: "referred_user_id"
+    has_one :referring_user, through: :referral
+    alias_method :referred_by, :referring_user
+
+    has_many :referrals, foreign_key: "referring_user_id", dependent: :destroy
+    has_many :referred_users, through: :referrals
+
+    validates :referral_code, uniqueness: true, allow_nil: true
+  end
+end
--- a/app/models/concerns/referrable.rb
+++ b/app/models/concerns/referrable.rb
@@ -2,11 +2,11 @@ module Referrable
   extend ActiveSupport::Concern
 
   included do
-    has_one :referral, dependent: :destroy, foreign_key: "referred_user_id"
+    has_one :referral, foreign_key: "referred_user_id", dependent: :destroy
     has_one :referring_user, through: :referral
     alias_method :referred_by, :referring_user
 
-    has_many :referrals, foreign_key: "referring_user_id", dependent: :destroy
+    has_many :referrals, foreign_key: "referring_user_id", dependent: :nullify
     has_many :referred_users, through: :referrals
 
     validates :referral_code, uniqueness: true, allow_nil: true
--- a/app/models/referral.rb
+++ b/app/models/referral.rb
@@ -2,13 +2,13 @@ class Referral < ApplicationRecord
   belongs_to :referred_user, class_name: "User"
   belongs_to :referring_user, class_name: "User", optional: true
 
-  before_create :associate_referrer
+  before_create :associate_referring_user
 
   validates :code, presence: true
 
   private
 
-  def associate_referrer
+  def associate_referring_user
     referring_user = User.find_by(referral_code: code)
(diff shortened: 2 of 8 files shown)


joemasilotti (changes requested):
I don't want any referral-based columns to exist on the `users` table, so I think we still need another model. The diagram in the issue outlined a `ReferralCode` which `belongs_to :user` and I think we still want that.

This will rework the associations on the `User` model to `has_one :referral_code`, `has_one :referrer`, and `has_many :referrals` with the necessary `through:` options.

We can also add a unique constraint to the `code` on the new table in the database, along with the existing Rails validation.

PatrickMcSweeny:
Currently there is a `has_one :referral` association on the User model so I think it would be confusing to have a `has_many :referrals` association.

joemasilotti:
Good point. I think that it should actually be `has_one :referral` and `has_many :referees`.

PatrickMcSweeny:
@joemasilotti I updated the commit

joemasilotti:
Thanks for the updates! I pulled down the code and I think I glossed over a major fallacy with this approach: it is terribly hard to reason around what is going on! Not your code, but the data modeling that I proposed.

Thanks to your initial commit I'm leaning back towards getting rid of the `ReferralCode` model entirely. Instead, a `Referral` can have an optional relationship with the referrer. I think that should simplify things quite a bit.

I also think that I haven't given the naming conventions of this enough though. Here's what I'd love the naming to be:

* A user has one referrer - the person who referred them
* A user has many referrals - the people they've referred
* A referral belongs to a referee (of class "User")
* A referral belongs to an _optional_ referrer (of class "User")

Which I _think_ means we need to add `where` clauses to the relationship references on the user model. Otherwise, it would always include records where the user was a referrer _or_ referee.

---

Does that make sense? Or am I overthinking this?

Let me know if you'd like to give it another pass. And no worries if not – I understand I've jumped around a lot with this one!

PatrickMcSweeny:
That comes pretty close to the commit I originally made. The main difference is that the relationship I used for the referrals made the user is referrals_made.

PatrickMcSweeny:
I pushed the original commit to the remote branch. Take a look and let me know if it needs any adjustments.

joemasilotti:
Awesome, thank you again!

I pulled out the referral-related code to a module and tweaked the names a bit more. I think that in theory "referrer" and "referee" make sense but in practice I have a hard time wrapping my head around which is which.

I like where this `Referrable` model ended up, especially the alias.

```ruby
module Referrable
  extend ActiveSupport::Concern

  included do
    has_one :referral, foreign_key: "referred_user_id", dependent: :destroy
    has_one :referring_user, through: :referral
    alias_method :referred_by, :referring_user

    has_many :referrals, foreign_key: "referring_user_id", dependent: :nullify
    has_many :referred_users, through: :referrals

    validates :referral_code, uniqueness: true, allow_nil: true
  end
end

user.referred_by # => User
user.referred_users # => [User]
```

Issue #758, opened 2023-01-26
Cache developer badges to a new model


joemasilotti:
Developers can now have up to two badged on their profile and more are coming soon. Before we add anything new I'd like a single place to reference and cache the existing ones.

A PR addressing this issue should:

- [ ] Create a new `Developers::Badge` model and table
- [ ] Cache the "recently active" status to this new model
- [ ] Ensure the status is updated when appropriate
- [ ] Migrate `developers.source_contributor` to the new table
- [ ] Ensure the admin form works with the new modeling

troyizzle:
I'd like to take this on!

joemasilotti:
All yours! Thank you!

troyizzle:
Couple questions,

I'm thinking it would be creating a table like this
 ```
    create_table :developers_badges do |t|
      t.belongs_to :developer
      t.boolean :recently_active, null: false, defualt: false
      t.boolean :source_contributor, null: false, default: false
      t.timestamps
    end
```

Backfill the data, should this been done through a lib task?
```
Developer.all.find_each do |developer|
  developer.badge.create!(
    recently_active: developer.recently_active?
    source_contributor: developer.source_contribuitor?
  )
end
```

For updating the data, I was thinking of a cron job on a set period of time
```
Developer
  .joins(:developer_badge)
  .where(updated_at: >= Developer::RECENTLY_ACTIVE_LENGTH.ago)
  .update_all(recently_active: false)
```
Also would need to update them to be active, so I was thinking of maybe on `ApplicationController` updating it

Some sudo code but would like to get your feedback!

joemasilotti:
The table looks good to me. And I imagine we will need to add/update more fields in the future so the cron job also sounds like a good idea.

The backfill can live in the dedicated backfill task. And the cron job can be a different task perhaps - not sure the best spot for that yet. I think the job can be run once per day and still get us what we want.

I know it is pseudo code, but we also want to make sure we are updating the recently active field to `true` when necessary. This PR should ensure that those calls are in place if they aren't already.

---

We will also probably need to split this into 2 PRs. One to add the migrations and migrate the data. And a second to switch over to the new code path. Otherwise we might deploy without the data being backfilled yet.

joemasilotti:
Doing some Spring cleaning and this issue is pretty stale. I'm going to close it out for now. But feel free to open a PR if you have something ready!

PR #774, opened 2023-02-06
Add developer responsiveness badges

### LINKED ISSUE #757: High response rate badge
Another way to help developers stand out on the platform is to highlight their response rate. Businesses are more inclined to reach out to someone who responds to every message than someone who doesn't.

I'm defining "response rate" as the number of conversations a developer has responded (at all) divided by the total number of conversations a developer is participating in.

- [ ] Add a badge to developer profiles if their response rate is over 90%
- [ ] Replace that badge with a different one if their response rate is a perfect 100%
- [ ] If needed, ensure the badge status is cached efficiently with no N+1 queries
- [ ] Stretch goal: Give developers 6 hours to respond to a new message before _dropping or removing_ their badge

For the badges, please use the following design/copy for a first pass.

* `90% response rate` in indigo
* `100% response rate` in purple

fig:
This PR adds a "responsiveness" badge to a developer's profile.
closes #757

## Pull request checklist

## Requirements from #757 
 - [x] Add a badge to developer profiles if their response rate is over 90%
 - [x] Replace that badge with a different one if their response rate is a perfect 100%
 - [x] If needed, ensure the badge status is cached efficiently with no N+1 queries
       Requires review by someone with better database fu.
 - [x] Stretch goal: Give developers 6 hours to respond to a new message before dropping or removing their badge
        Too follow

I'm not totally happy with the text used in the badges so any suggestions for alternatives would be gratefully appreciated.
Current text: "Good responder" / "Perfect responder"

I have chosen to implement this feature by writing a concern to be included in the `Developer` model. The concern utilises the existing `Stats::Conversation` class to calculate developer response rate.

## Sample rendering

image


joemasilotti:
Thanks for the submission!

Let's use the badge titles I outlined in the issue for now: "100% response rate" and "90% response rate".

Also, I won't be able to merge this until the third todo item is completed. Without caching the values we end up with a nasty N+1 on `/developers`:

```
web      |   Conversation Count (0.3ms)  SELECT COUNT(*) FROM "conversations" WHERE "conversations"."developer_id" = $1 AND "conversations"."developer_blocked_at" IS NULL AND "conversations"."business_blocked_at" IS NULL  [["developer_id", 26]]
web      |   ↳ app/models/stats/conversation.rb:10:in `sent'
web      |   Conversation Count (0.3ms)  SELECT COUNT(*) FROM "conversations" WHERE "conversations"."developer_id" = $1 AND "conversations"."developer_blocked_at" IS NULL AND "conversations"."business_blocked_at" IS NULL  [["developer_id", 25]]
web      |   ↳ app/models/stats/conversation.rb:10:in `sent'
web      |   Conversation Count (0.3ms)  SELECT COUNT(*) FROM "conversations" WHERE "conversations"."developer_id" = $1 AND "conversations"."developer_blocked_at" IS NULL AND "conversations"."business_blocked_at" IS NULL  [["developer_id", 20]]
web      |   ↳ app/models/stats/conversation.rb:10:in `sent'
web      |   Conversation Count (0.3ms)  SELECT COUNT(*) FROM "conversations" WHERE "conversations"."developer_id" = $1 AND "conversations"."developer_blocked_at" IS NULL AND "conversations"."business_blocked_at" IS NULL  [["developer_id", 19]]
web      |   ↳ app/models/stats/conversation.rb:10:in `sent'
web      |   Conversation Count (0.3ms)  SELECT COUNT(*) FROM "conversations" WHERE "conversations"."developer_id" = $1 AND "conversations"."developer_blocked_at" IS NULL AND "conversations"."business_blocked_at" IS NULL  [["developer_id", 18]]
web      |   ↳ app/models/stats/conversation.rb:10:in `sent'
web      |   Conversation Count (0.4ms)  SELECT COUNT(*) FROM "conversations" WHERE "conversations"."developer_id" = $1 AND "conversations"."developer_blocked_at" IS NULL AND "conversations"."business_blocked_at" IS NULL  [["developer_id", 17]]
web      |   ↳ app/models/stats/conversation.rb:10:in `sent'
web      |   Conversation Count (0.4ms)  SELECT COUNT(*) FROM "conversations" WHERE "conversations"."developer_id" = $1 AND "conversations"."developer_blocked_at" IS NULL AND "conversations"."business_blocked_at" IS NULL  [["developer_id", 16]]
web      |   ↳ app/models/stats/conversation.rb:10:in `sent'
web      |   Conversation Count (0.3ms)  SELECT COUNT(*) FROM "conversations" WHERE "conversations"."developer_id" = $1 AND "conversations"."developer_blocked_at" IS NULL AND "conversations"."business_blocked_at" IS NULL  [["developer_id", 15]]
web      |   ↳ app/models/stats/conversation.rb:10:in `sent'
web      |   Conversation Count (0.3ms)  SELECT COUNT(*) FROM "conversations" WHERE "conversations"."developer_id" = $1 AND "conversations"."developer_blocked_at" IS NULL AND "conversations"."business_blocked_at" IS NULL  [["developer_id", 14]]
web      |   ↳ app/models/stats/conversation.rb:10:in `sent'
```

Let me know if you have appetite to take care of that. If no, no worries! I can take over.

fig:
All noted.

After sleeping on this, I'm not loving the dependency on `Stats::Conversation` during the request cycle. Maybe adding a database column for `response_rate` might be a better way to go architecturally. I envision a background job queued to update this column a configurable time (e.g. 12 hours) after a 'cold message' is received.

This would solve the N+1 and keep the use of `Stats::Conversation` to a minimum and in a background job. Essentially, we will cache the `response_rate` in the db.

Your thoughts?

joemasilotti:
That's very similar to what I was thinking, too!

We can roll ahead and cache this directly in the `developers` table for now. Then when #758 is implemented we can migrate over the data.

Or, we can hold off and wait until that is merged in first. Your call!

joemasilotti:
It will generate on additional db query per _list_ of developers, instead of one for every badge for every developer.

So, for each paginated page of 10 developers it will add 1 additional query.

If we had all of the current and upcoming badge logic calculated on the fly it would generate 40 additional queries (4 badges x 10 developers).

fig:
- The solution now adds a new column `response_rate` to the `developers` table.
- A newly created conversation triggers a background job to update the Developer's response rate after a configurable length of time.
- Badge text fixed.

A migration is included in this PR
A rake task is provided to backfill the new table column (`rake backfills:developer_response_rate`).

app/controllers/cold_messages_controller.rb
```diff
@@ -51,6 +52,14 @@ def require_signed_hiring_agreement!
...
   end
 
+  def update_developer_response_rate
+    return unless response.redirect?
+
+    UpdateDeveloperResponseRateJob
+      .set(wait: Rails.application.config.developer_response_grace_period)
+      .perform_later(developer)
+  end
+
```
joemasilotti, line 62: This feels like a round about way to queue the job. Let's instead add this to `Messages::Notifications` or inline to the controller.

lib/tasks/backfills.rake
```diff
@@ -1,11 +1,17 @@
...
-      last_notification = conversation.messages.last.notifications_as_message.last
-      if last_notification&.unread?
-        conversation.update(user_with_unread_messages: last_notification.recipient)
+  task developer_response_rate: :environment do
+    begin
+      ActiveRecord::Migration.check_pending!
+    rescue ActiveRecord::PendingMigrationError => e
+      if e.message.include?("add_response_rate_to_developers")
+        puts e.message
+        exit 1
```
joemasilotti, line 9: Let's remove this and let the job fail on its own when the migration hasn't been run.

lib/tasks/backfills.rake
```diff
@@ -1,11 +1,17 @@
...
       end
     end
+    Developer.find_each(batch_size: 50) do |developer|
+      UpdateDeveloperResponseRateJob.perform_now(developer)
+      sleep(0.01)
```
joemasilotti, line 14: Why sleep?

app/jobs/update_developer_response_rate_job.rb
```diff
@@ -0,0 +1,11 @@
+class UpdateDeveloperResponseRateJob < ApplicationJob
+  queue_as :default
+
+  GRACE_PERIOD = Rails.application.config.developer_response_grace_period
+
+  def perform(developer)
+    conversations = developer.conversations.where("created_at < ?", GRACE_PERIOD.ago)
+    replied_rate = Stats::Conversation.new(conversations).replied_rate
+    developer.update_column(:response_rate, (replied_rate * 100).round.floor(-1))
```
joemasilotti, line 9: I don't think this logic is correct. Isn't it only querying for conversations started within the last 12 hours?

Instead we need something that _ignores_ all conversations created in the last 12 hours and calculates the response rate of the others.

joemasilotti (changes requested):
Looking good!

image

I'm now realizing that 90 vs. 100 might not matter very much. Let's consolidate them. If a developer has a 90%+ response rate show a badge that says "High response rate".

Also, it looks like the background job is updating the developer's `updated_at` column which is used to show the "Recently active" badge. Let's update the job to not update that column for now with a comment explaining why. This will be removed in a future PR when we start to cache the badges in their own model.

I also made some inline questions/comments.

app/controllers/cold_messages_controller.rb
fig: I didn't like it here either, but it didn't seem to fit in with the Notifications either. It possibly deserves it's own class?
I've moved it in with the Notifications for now, as suggested.

lib/tasks/backfills.rake
fig: Sleep to throttle mass Database insertions between batches of 50.
I'm not a db expert, so I don't totally understand the reasoning or the choice of value, but I've read that this is a GoodIdea™.
Refs:
https://world.hey.com/paulosilva/ruby-on-rails-backfilling-data-136adb66
https://github.com/ankane/strong_migrations#backfilling-data

app/jobs/update_developer_response_rate_job.rb
fig: `developer.conversations.where("created_at < ?", GRACE_PERIOD.ago)`
Fetch conversations where `created_at` is less than (i.e earlier than) the given time value.

I've added tests to confirm the logic.

fig:
I've changed the implementation to have a single "High Response Rate" badge as suggested.

I noticed that "recently active" was relying on the `updated_at` column. The background job is calling `#update_column` which updates the attribute directly in the database issuing an UPDATE SQL statement. I chose this method because updated_at/updated_on are not updated. One downside is that validations are also skipped. I've add a comment to change this to a regular `#update!` at a later date.

I noticed that `test/components/developers/badge_component_test.rb` wasn't actually doing what it claimed.
It was possible to delete the lines setting the attributes of the developer and the tests would still pass.
I have renamed the test file to `test/components/developers/badges_component_test.rb` and refactored the tests within to exercise the Developers::BadgesComponent. This class utilises the BadgeComponent for rendering and so now the whole chain is tested.

lib/tasks/backfills.rake
joemasilotti: Got it. TIL, thanks!

app/jobs/update_developer_response_rate_job.rb
joemasilotti: Ah! You're right. I read that backwards. Thanks for clarifying - and with tests to prove it.

joemasilotti:
What's even weirder is that when I run `bin/rails rake db:reset` in development I get the same error. But when starting a new conversation from the UI the job enqueues without issue.

Honestly, I'm not really sure what's going on! I'll have to dig in more.

fig:
This is what I have learnt in the last few hours.

- Unless specified otherwise, the data seeding task will use the :async queue adaptor.
- We are specifying the :inline adaptor in our seeds file, so this will be honoured.
- The :inline adaptor does not allow scheduling jobs so raises an error when trying to seed Conversations because of the `.set(wait: ...)` method.
- If we specify :async only for the UpdateDeveloperResponseRateJob,
  (`UpdateDeveloperResponseRateJob.queue_adapter = :async` in seeds.rb), the db will seed, but will enqueue jobs in our dev environment.

joemasilotti:
Got it! Thanks for identifying all the moving parts - that's helpful.

Ignoring the implementation details for a second... when running the seeds ideally the column is populated right away. Meaning, the job kicks off immediately.

Is it possible to achieve that? Perhaps via running the job conditionally async/sync depending on the environment? Or something more graceful in the config or options passed to the job?

fig:
One option might be to set  `config.developer_response_grace_period` to a very low value (5secs) in development and have the job run :async during seeds. The low grace period won't be an issue as our seeded developers will be responding almost instantaneously and the update job will run as soon as a dev server is fired up. My only concern is that there will be jobs left dormant in redis if a server instance is not started after seeding. After saying that, I believe the response rate update job is idempotentish (multiple calls within a short time-frame will produce the same result) so it might not be an issue.

db/seeds/conversations.rb
```diff
@@ -4,15 +4,18 @@
 conversation = Conversation.find_or_create_by!(developer:, business:)
 SeedsHelper.create_message!(conversation:, sender: business, body: "Your XP looks perfect.", cold_message: true)
 SeedsHelper.create_message!(conversation:, sender: developer, body: "Let's chat over Zoom.")
+UpdateDeveloperResponseRateJob.perform_now(developer)
```
joemasilotti, line 7: I'm confused as to why this needs to be performed outside of the normal creating of messages. Should the job get kicked off "in 0 seconds" when running the seeds?

test/jobs/update_developer_response_rate_job_test.rb
```diff
@@ -0,0 +1,81 @@
...
+
+    # update the developer's response rate
+    UpdateDeveloperResponseRateJob.perform_now(@developer.reload)
+  end
+
+  test "should update developer response rate from the returned value" do
+    replied_rate = 0.42
+    expected_response_rate = 42
+    mock = Minitest::Mock.new
+    mock.expect :replied_rate, replied_rate
```
joemasilotti, line 35: Why a song and dance with the mock instead of checking for `50`? We clearly create both a replied to and not replied to conversation in the setup of this test, so no risk for mystery guests.

test/jobs/update_developer_response_rate_job_test.rb
```diff
@@ -0,0 +1,81 @@
...
+
+    Stats::Conversation.stub :new, mock do
+      UpdateDeveloperResponseRateJob.perform_now(@developer.reload)
+      assert_equal expected_response_rate, @developer.reload.response_rate
+    end
+    assert_mock mock
+  end
+
+  test "should not take into account unreplied conversations created during grace period" do
+    before = @developer.reload.response_rate
```
joemasilotti, line 45: We can use the following for this test.

```ruby
assert_no_changes @developer.reload.response_rate do
  # ...
end
```

test/jobs/update_developer_response_rate_job_test.rb
```diff
@@ -0,0 +1,81 @@
...
+  test "should increase response rate if replied to within grace period" do
+    before = @developer.reload.response_rate
+
+    business = Business.create!(business_attributes)
+    conversation = Conversation.create!(developer: @developer, business: business)
+    Message.create!(conversation: conversation, sender: business, body: "First contact")
+    Message.create!(conversation: conversation, sender: @developer, body: "First reply")
+    UpdateDeveloperResponseRateJob.perform_now(@developer.reload)
+
+    assert_operator @developer.reload.response_rate, :>, before
```
joemasilotti, line 64: The relative assertions are a little hard to reason about. Let's we instead use `assert_changes @developer.reload.response_rate, from: X, to: Y`. I think that's a bit clearer.

fig:
- Set the response grace period to zero in production and use the :async adaptor to avoid the seeding issue.
- Added a call to the update job after seeding conversations — this ensures that a freshly seeded development environment starts with appropriate developer badges.
- Fixed an issue with the response rate logic whereby conversations replied to within the grace period were not being taken into account.

The test files (especially [update_developer_response_rate_job_test.rb](https://github.com/joemasilotti/railsdevs.com/pull/774/files#diff-0a8de7757c0bf4b2a5663c45cfa5cc208860c0440eff73d64949d1f9b9bc2ad9) ) are probably not very pretty. I usually use RSpec, so feel free to refactor them, or open an issue to get them tidied up. I believe their functionality is sound but, again, they could probably use a closer look by someone proficient in MiniTest.

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### February
 
+* February 27 - Add a badge for developers with a High Response Rate
```
```suggestion
* February 28 - Add a badge for developers with a high response rate #774 @fig
```

joemasilotti (changes requested):
Great idea using `0.seconds`! Just a few more minor requests and this is ready to merge.

app/models/conversation.rb
```diff
@@ -68,8 +68,6 @@ def first_reply?(developer)
     messages.where(sender: developer).one? && latest_message.sender == developer
   end
 
-  private
-
   def developer_replied?
     messages.from_developer.any?
```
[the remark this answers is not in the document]
fig: Ah yes! Good spot.

db/seeds/conversations.rb
fig: I thought it would, but it didn't appear to be doing so.
I shall conduct further experiments.

test/jobs/update_developer_response_rate_job_test.rb
fig: This is my brain trying to translate RSpec to MiniTest. 🙃
I'll see if I can make it clearer.

db/seeds/conversations.rb
fig: Working as expected now without the call to the job.

app/models/conversation.rb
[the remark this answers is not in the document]
fig: Test added for 'Conversation#developer_replied?'

test/jobs/update_developer_response_rate_job_test.rb
fig: I have done an almost complete rewrite of the `test/jobs/update_developer_response_rate_job_test.rb` 
I've learn't a lot. I see more MiniTest in my future! :crystal_ball:

test/jobs/update_developer_response_rate_job_test.rb
fig: TIL! Thank you.
Done.

test/jobs/update_developer_response_rate_job_test.rb
fig: Another thing I didn't know until today. Thank you.

joemasilotti (approved):
Looks great! Nice work @fig. Thanks for all the updates.

PR #776, opened 2023-02-09
Tu cache developer badge

troyizzle:
This should handle creating the migration for #758.

- [x] My code contains tests covering the code I modified
- [x] I linted and tested the project with `bin/check`
- [ ] I added significant changes and product updates to the [changelog](CHANGELOG.md)

PR #777, opened 2023-02-09
Change logic for badges to be handle from Developers::Badge

troyizzle:
This should finish up #758, there is one thing to do.

Updating the recently active status to false, I believe you mentioned to do this on a cron job but I'm not sure where to put it. Any suggestions?


app/controllers/application_controller.rb
```diff
@@ -45,4 +46,12 @@ def set_variant
       request.variant = :redesign
     end
   end
+
+  def update_recently_active
+    return unless current_user&.developer
+    return unless current_user.developer&.badge
+    return if current_user.developer.badge.recently_active?
+
+    UpdateDeveloperRecentlyActiveJob.perform_now(current_user.developer.id, true)
```
joemasilotti, line 55: I think we still need to save _when_ the developer was last active. This is now the only place we are updating the column, which means it will never be set to false.

Also, if we are going to use a job let's keep the logic there and pass in `current_user&.developer_id`. Then the job can check if the developer exists and if so set the timestamp on the badge.

app/controllers/application_controller.rb
```diff
@@ -11,6 +11,7 @@ class ApplicationController < ActionController::Base
   around_action :set_locale
   before_action :redirect_suspended_accounts
   before_action :set_variant
+  before_action :update_recently_active
```
joemasilotti, line 14: I'd like to change this trigger to only occur when:

1. The developer sends a message
2. The developer updates their profile (even with no changes)

app/jobs/create_developer_badge_job.rb
```diff
@@ -0,0 +1,13 @@
+class CreateDeveloperBadgeJob < ApplicationJob
```
joemasilotti, line 1: Let's remove this job, test, and backfill in this PR since it will be merged in after #776.

app/models/developer.rb
```diff
@@ -27,13 +24,16 @@ class Developer < ApplicationRecord
...
   has_many :hired_forms, class_name: "Hired::Form", dependent: :destroy
+  has_one :badge, dependent: :destroy, class_name: "Developers::Badge"
   has_one :location, dependent: :destroy, autosave: true
   has_one :role_level, dependent: :destroy, autosave: true
   has_one :role_type, dependent: :destroy, autosave: true
   has_one_attached :cover_image
 
   has_noticed_notifications
 
+  delegate :source_contributor?, :recently_active?, to: :badge
```
joemasilotti, line 35: What's the purpose of this? We are already delegating in the `Developers::BadgeComponent`.

app/models/developer.rb
```diff
@@ -74,8 +76,8 @@ class Developer < ApplicationRecord
   scope :featured, -> { where("featured_at >= ?", FEATURE_LENGTH.ago).order(featured_at: :desc) }
   scope :newest_first, -> { order(created_at: :desc) }
   scope :profile_reminder_notifications, -> { where(profile_reminder_notifications: true) }
-  scope :recently_active, -> { where("developers.updated_at >= ?", RECENTLY_ACTIVE_LENGTH.ago) }
-  scope :source_contributor, -> { where("source_contributor >= ?", true) }
+  scope :recently_active, -> { joins(:badge).where(developers_badges: {recently_active: true}) }
```
joemasilotti, line 79: Let's create a scope on the badge model to check for recently active (see my comment above about storing the timestamp).

test/fixtures/developers_badges.yml
```diff
@@ -0,0 +1,9 @@
+_fixture:
```
joemasilotti, line 1: We can delete this file, right? We have the fixture above.

joemasilotti (changes requested):

joemasilotti:
@troyizzle, a heads up that #774 was just merged which adds a new badge.

Feel free to integrate this into your backfill PR and this one if you want.

Or, don't and we can do it later. But wanted to give you a heads up. I'm happy either way!

joemasilotti:
Hey again @troyizzle, do you still have appetite to finish this PR? Let me know either way and we can figure out next steps!

joemasilotti:
Hey again @troyizzle. Hate to be a pain but this feature is blocking a few other pieces of work in the pipeline.

I'm going to take it over in a day or two if I don't hear from you.

joemasilotti:
Closing because the PR has grown quite stale and I haven't heard back from OP in over a month.

PR #780, opened 2023-02-13
Add "referred by" to admin emails

### LINKED ISSUE #729: Add "referred by" to admin emails
Admins receive an email when a user creates a business or developer profile. I'd like to know if that user was referred by someone.

- [ ] Add "Referred by #{referrer.email}" to the bottom of these emails, if present

The emails are `AdminMailer#new_business` and `AdminMailer#new_developer`.

This relies on the modeling built out by #736 to get the referrer's email address.

PatrickMcSweeny:
Closes #729 

## Pull request checklist


test/fixtures/referrals.yml
```diff
@@ -0,0 +1,9 @@
+one:
```
PatrickMcSweeny, line 1: I added these fixtures so the "referred by" text will show up in the mailer previews.

joemasilotti: Hmm, shouldn't they appear in the database seeds then?

app/models/business.rb
```diff
@@ -17,6 +18,8 @@ class Business < ApplicationRecord
 
   alias_attribute :name, :contact_name
 
+  delegate :email, to: :referring_user, prefix: true, allow_nil: true
```
joemasilotti, line 21: Ohh, very nice! But let's abstract these two methods to a concern that can be included in `Developer` and `Business`.

app/views/admin_mailer/new_business.text.erb
```diff
@@ -1,3 +1,7 @@
 <%= @business.contact_name %> (<%= @business.user.email %>) added their business!
 
 <%= notification_url(@notification.record) %>
+
+<% if @business.referring_user_email.present? %>
+  Referred by <%= @business.referring_user_email %>
+<% end %>
```
joemasilotti, line 7: This will render two spaces in front of the text on a plain-text email. Let's remove the indentation in these two emails.

joemasilotti (changes requested):
Looking good! I made a few inline requests.

app/models/business.rb
PatrickMcSweeny: In order to do that the `belongs_to :user` association would have to be moved to the new concern. Are you OK with that?

joemasilotti: Why is that?

PatrickMcSweeny: I got the following error when I tried creating a concern without that association:

"Cannot have a has_many :through association 'Business#referring_user' which goes through 'Business#user' before the through association is defined."

The only options I can think of would be to move the "include" statements after the association is defined in the models, or to move the assocation into the new concern.

joemasilotti: Ah, gotcha. OK, I don't think it makes sense to pull that association into a concern because it is core to the developer/business model.

One more idea, though. Can we define the association _twice_? Once in the model and again in the concern? If so maybe that could work. If not, then lets skip the refactor and roll with what you originally proposed.

PatrickMcSweeny: I tried defining the association in both the model and the concern, and the error kept happening.

joemasilotti: Gotcha. Let’s punt on the reactor then. Thanks for diving in!

test/fixtures/referrals.yml
PatrickMcSweeny: I guess I misunderstood that. I removed the fixtures.

app/views/admin_mailer/new_business.text.erb
PatrickMcSweeny: Done

app/models/business.rb
PatrickMcSweeny: I reverted the changes

joemasilotti:
Looks like all that’s left is the seed vs. fixture question!

PatrickMcSweeny:
I deleted the fixtures and the previews still work, so I think that has been resolved.

joemasilotti:
Great! Can you add them to the seed file? That way we can test in the mailer preview without having to manually create a referral. Thanks!

PatrickMcSweeny:
I added referrals to the seeds.

joemasilotti:
Excellent! I tweaked some loading in the seeds and made sure the mailer previews always pick up those records.

Going to merge when CI passes - thanks again for your contribution!

PR #782, opened 2023-02-16
[DRAFT]: Update file_upload_controller.js to use Media Stream

KonnorRogers:
I have no idea if this is actually an issue, but a path you could explore is using MediaSource.

According to MDN: 

> Warning: If you still have code that relies on createObjectURL() to attach streams to media elements, you need to update your code to set srcObject to the MediaStream directly.

https://developer.mozilla.org/en-US/docs/Web/API/URL/createObjectURL#using_object_urls_for_media_streams

I have not tested this, but could be a basis for someone else, I had a few minutes to spare to help.

## Pull request checklist

### DIFF
--- a/app/javascript/controllers/file_upload_controller.js
+++ b/app/javascript/controllers/file_upload_controller.js
@@ -52,7 +52,7 @@ export default class extends Controller {
       }
     }
       
-    img.src = URL.createObjectURL(mediaSource);
+    this.#createPreviewFallback(file)
   }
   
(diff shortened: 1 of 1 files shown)


joemasilotti:
Thanks for the WIP, @KonnorRogers! It doesn't seem like any browser outside of Safari has support for this just yet. So I'm not worried about losing support in new updates.

But if someone submits a working PR then I will happily merge it in.

Issue #779, opened 2023-02-10
Notify admins when a user's GitHub company or organization changes


joemasilotti:
Continuing #756 but with GitHub...

I'm exploring another way to keep businesses honest and pay the required hiring fee. One idea is to track developers on ~LinkedIn~ GitHub and see if their job status changes.

Documented in the linked issue above, the GitHub API can fetch a lot of information for free. I'm most interested in "company" and "organizations". These most correlate to a developer getting a new job.

A PR addressing this issue should:

- [ ] Send an email to admins once per day of any GitHub changes for developers

This is left intentionally vague as there are a ton of ways to do this. I imagine a rough outline would be running a daily background job to fetch the relevant information, comparing it to existing records, and sending an email with the changes.

I asked about how folks handle this [on Twitter](https://twitter.com/joemasilotti/status/1622367065389805568?s=20&t=9wysN7G7_9Iuf1zFH7OafQ) and got a variety of responses. The most viable (in my opinion) is working with [`ActiveModel::Dirty`](https://api.rubyonrails.org/classes/ActiveModel/Dirty.html) in some way. That means we can rely on the framework to do the "diff" and not calculate it manually.

I'm open to ideas and questions!

PatrickMcSweeny:
I looked that API documentation and it appears that the endpoint mentioned in the linked issue only returns information for the user that is currently authenticated with the API. [This endpoint](https://docs.github.com/en/rest/orgs/orgs?apiVersion=2022-11-28#list-organizations-for-a-user) returns publicly available organization memberships for a user, but this would not cover private information about organizations. Another possibility would be to use [this endpoint](https://docs.github.com/en/rest/orgs/orgs?apiVersion=2022-11-28#get-an-organization] which provides a URL to list public members of that organization.

I guess the major sticking point is how big of an issue relying on publicly available information about organization memberships would be. If @joemasilotti feels that it is worthwhile to move forward on this, I would be willing to work on it.

joemasilotti:
Great point! I think that company plus public organization info is better than nothing right now. But let's test that assumption.

Want to throw together a throw-away PR that tallies how many folks have their company and/or an organization listed on GitHub? I can run it against production and share info here.

Based on that we can decide to move forward with a real PR that persists data, notifies admins, is tested, etc.

PatrickMcSweeny:
Sure, I can do that. Would you like me to work with actual user data? I could probably get by with a list of GitHub usernames

[...]

PR #783, opened 2023-02-18
Create class to find developers' organizations

PatrickMcSweeny:
This is research for #779. @joemasilotti I think this should be enough to get you started, but let me know if you need anything more.


joemasilotti:
Thanks for this! Can you please add company to the query and some output? Ideally, in the following format:

```
Developer profiles

Total: XXX
With GitHub organization(s): XXX
With company: XXX
With both: XXX
```

Thanks! Once that's in I'll run the numbers and see if it makes sense to implement.

PatrickMcSweeny:
@joemasilotti Done

joemasilotti:
OK, I finally got my GitHub access token working! Here are the results:

```
Developer profiles
Total: 828
With GitHub organization(s): 188
With company: 362
With both: 122
```

So we are looking at over 40% with either one. Which feels high enough to validate my assumption. Let's move forward with the original issue!

I'll close this PR and we can continue conversation back in #779.

PR #785, opened 2023-02-21
[DRAFT] Extract general text into locale files to make forking easier

arminghofrani:
Following our conversation on Twitter about taking all occurrences of the string "RailsDevs" (and other details, such as the founder's email) and putting them inside locale files as variables. The idea is to make it easier to fork the project, change the name (and other details) and reduce merge conflicts when pulling in commits from your repo.

So far, I've only extracted the string "RailsDevs" from short strings - let me know your thoughts on the approach.

To proceed with making the repo easier to fork, the question is how to proceed with longer strings. Take the following line from `app/views/developer_mailer/welcome.text.erb` as an example:

> I'm Joe, the founder of RailsDevs. Thanks for adding your profile!

It doesn't make sense to turn this into:

> I'm <%= t("general.organization.founder.name") %>, the founder of <%= t("general.project.name") %>. Thanks for adding your profile!

Because forks may choose to rewrite this text anyway. A potential option is to extract the entire text into locale files as such:

> I'm %{founder_name}, the founder of %{project_name}. Thanks for adding your profile!

But that defeats the purpose of using email/view templates (although it may work for longer strings in other bits of the codebase).

What are your thoughts?

## Pull request checklist

### DIFF
--- a/app/components/nav_bar/guest_component.html+redesign.erb
+++ b/app/components/nav_bar/guest_component.html+redesign.erb
@@ -7,8 +7,8 @@
             <!-- Logo -->
             <div class="flex-shrink-0">
               <%= link_to root_path, class: "shrink-0 flex items-center text-white fill-current" do %>
-                <span class="sr-only"><%= t("general.project.name") %></span>
-                <%= inline_svg_tag "logo.svg", class: "h-10 w-auto", alt: "#{t("general.project.name")} logo", title: "#{t("general.project.name")} logo" %>
+                <span class="sr-only"><%= Rails.configuration.strings.project_name %></span>
+                <%= inline_svg_tag "logo.svg", class: "h-10 w-auto", alt: "#{Rails.configuration.strings.project_name} logo", title: "#{Rails.configuration.strings.project_name} logo" %>
               <% end %>
             </div>
 
--- a/app/components/nav_bar/guest_component.html+redesign.erb
+++ b/app/components/nav_bar/guest_component.html+redesign.erb
@@ -7,8 +7,8 @@
             <!-- Logo -->
             <div class="flex-shrink-0">
               <%= link_to root_path, class: "shrink-0 flex items-center text-white fill-current" do %>
-                <span class="sr-only">RailsDevs</span>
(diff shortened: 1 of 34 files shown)


joemasilotti:
Great question!

I'm assuming the goal with this PR is to templatize as much as possible to make forking as easy as possible. If so, my vote is moving the founder's name and project name somewhere and leaving the mailer.

> I'm <%= t("general.organization.founder.name") %>, the founder of <%= t("general.project.name") %>. Thanks for adding your profile!

This means someone can fork the project and get up and running without having to change anything but some config. They are always free to customize the mailer content later down the road.

---

Thinking on this a bit more I think might also want a configuration that exists outside of the locales. "RailsDevs" for example, will never be translated. Neither will the founder's name (right?).

My vote is to put these in `config/project.yml` or similar and load them similar to emails in `config/application.rb`. Then everything that might be translated leave in the locale files.

What do you think?

dctanner:
Hey chaps!

It'd actually be great if we can replace `I'm <%= t("general.organization.founder.name") %>, the founder of <%= t("general.project.name") %>` which something like `<%= t("general.project.email_intro") %>`, because in our case we have two founders and we might just want to put "Hi, it's FirstName from ProjectName here." or something more general.

In workingin.ai we also had to replace Developer, mainly in the locale strings. So it would be good to have this defined in one place too.

dctanner:
I also noticed that in the locale files, things like RailsDevs, Developer etc are mentioned a lot in the strings. Can we make this a variable in the local strings too?

joemasilotti:
@dctanner, for the welcome email, do you see yourself changing the rest of the email? Or only the introduction?

dctanner:
@joemasilotti I think just the introduction. Maybe we can think of it as an email intro variable like an email signature is. 

We're actually adding in some more changes over the next week or so. We'll add a comment when this work is done and the PR is in a better state to hopefully merge 🙏

joemasilotti:
Sounds great all around, thanks for the update!

joemasilotti:
Hey @dctanner and @arminghofrani, I'm doing a bit of Spring cleaning on the repo and closing old/stale PRs.

If you still want to see this merged let me know when you have time to wrap this up!

dctanner:
Had to put things on hold right now so you are welcome to rm and we can resubmit in future if need be.

PR #790, opened 2023-02-28
Remove database delivery for business and developer welcome notifications

### LINKED ISSUE #789: Option to hide certain notifications from the UI
`Developers::WelcomeNotification` and `Businesses::WelcomeNotification` only exist to trigger an email. I'd like to hide these from showing up as notifications in the dropdown and `/notifications`.

A PR addressing this issue should _either_:
- [ ] Add a flag to `notifications` to show in the UI or not; respect flag when viewing notifications
- [ ] Create a system/approach to sending emails

I might be overthinking it, the second one might be ripping out a lot of code. The choice is yours!

gathuku:
## Pull request checklist

Remove database delivery method for notification we don't want them to appear on the UI but send with other delivery methods like email.

closes #789


gathuku:
@joemasilotti do you have some tests for checking an emails is enqueued after `developer.save_and_notify`

joemasilotti:
I wouldn't think so since they would be queued to run on a background job when creating the notification. We will want to add them in.

gathuku:
@joemasilotti without the database delivery method the record will be `nil`

https://github.com/joemasilotti/railsdevs.com/blob/2ce462409934bd423ad53a34b26d70d85db11f91/app/mailers/developer_mailer.rb#L17

Is there another simple way you can get the notification instance at this level?

Only the developer and recipient params will be available.

```
(byebug) params
{:developer=>#<Developer id: 1, name: "Concepcion Fritsch", available_on: "2022-02-25", hero: "Molestiae ut dolores vel.", bio: "Maiores ducimus repellendus. Et architecto iusto. ...", website: nil, github: nil, twitter: nil, created_at: "2022-03-12 07:23:39.500642000 +0000", updated_at: "2022-03-12 07:23:39.574053000 +0000", user_id: 1, linkedin: nil, search_status: "open", preferred_min_hourly_rate: nil, preferred_max_hourly_rate: nil, preferred_min_salary: nil, preferred_max_salary: nil, textsearchable_index_col: "'a':30 'architecto':9 'aut':27,31 'autem':32 'comm...", featured_at: nil, profile_reminder_notifications: true, stack_overflow: nil, public_profile_key: nil, source_contributor: false, mastodon: nil, availability_status: "now">, :recipient=>#<User id: 1, email: "angela@example.com", created_at: "2022-03-12 07:23:36.732180000 +0000", updated_at: "2022-07-14 16:25:32.600803000 +0000", admin: false, authentication_token: [FILTERED], suspended: false, referral_code: nil>, :record=>nil}
```

joemasilotti:
Nice! So glad I have those mailer validators. They have caught so many potential issues.

There will no longer be a notification to reference. So the mailer will need to get the developer/business as a param directly. `DeveloperMailer#celebration_promotion` and `BusinessesMailer#developer_profiles` are using that pattern if it helps to model after something.

gathuku:
@joemasilotti i think we can get rid of `@notification` variable plus this link. Which other link url can we replace with https://railsdevs.com ?

https://github.com/joemasilotti/railsdevs.com/blob/2ce462409934bd423ad53a34b26d70d85db11f91/app/views/developer_mailer/welcome.html.erb#L2

joemasilotti:
`https://railsdevs.com` works.

gathuku:
Did something like this to get the subject. Check if it works well
```
  subject = Developers::WelcomeNotification.new.title
```

gathuku:
> Did something like this to get the subject. Check if it works well
> 
> ```
>   subject = Developers::WelcomeNotification.new.title
> ```

Just remove it 4bc539fce2b3ea415ff7a2eaef6b844fba48b284 😄

joemasilotti (approved):

joemasilotti:
Sorry for the rewrite... I realized we were doing too much working around Noticed. I ripped out the notifications and instead am using a good ol' mailer directly to send the email.

Thanks for your contribution, @gathuku!

Issue #806, opened 2023-03-16
Archive conversations


joemasilotti:
The conversations index page can get overwhelming after you've had a lot of contact with folks. I'd like a way to make it easier to follow a few important threads without getting bogged down by the rest.

A PR addressing this issue should:

- [ ] Add an Archive button to every conversation
- [ ] When clicked, remove the conversation from `/conversations`
- [ ] If any conversations are archived, add a link to the bottom labeled "Archived conversations"
- [ ] When clicking this link, visit `/conversations/archived` and show the archived ones
- [ ] Each conversation on this screen should have a Unarchive button that moves it back to `/conversations`

JuzerShakir:
I would like to try this out!

vinhmai570:
Hi Joe. Can I receive this issue?

joemasilotti:
@JuzerShakir, you commented first so this is all yours!

JuzerShakir:
**Here's the status of this issue till now:**
[Screencast from 17-03-23 07:43:28 PM IST.webm](https://user-images.githubusercontent.com/10782481/225930374-4469917c-c62a-4795-94cb-3f21948e4ea6.webm)

I believe I have achieved all the tasks mentioned in this Issue! Hurray! 

But it actually doesn't work as expected as you can see in the video! I created `archived?` attribute as a boolean in the `conversation` table to achieve this task. But there's a flow with this logic as I released just now.

As conversations involve 2 users, so if a user sets a conversation as archived then the other side of the users' conversation is also set as `archived`. So when a message is received by a user then that conversation is in the `/archived`.

I have been about the solution but I am not understanding how should I resolve this. The only way I think is I need to undo all changes and redo them again. I spent hours on this only to know that this logic is not worth it. I need help @joemasilotti, what should I do?

joemasilotti:
This is looking good! I'll probably tweak the design of the button but the flow looks great.

I think we will need to have two fields on `Conversation`. Look at the schema and note how we use `developer_blocked_at` and `business_blocked_at`. Let's model this the same way: `developer_archived_at` and `business_archived_at`.

You could also look at `BlocksController` to see how we update the correct field.

JuzerShakir:
Yes! That's a nice idea! The `developer_blocket_at` & `business_blocked_at` are `datetime` type fields whose values are by default `nil` but when they're `touch`ed in the controller then their value is updated with the current timestamp!

[...]

PR #813, opened 2023-03-18
Archive conversations

JuzerShakir:
## Pull request checklist
This PR resolve [issue](https://github.com/joemasilotti/railsdevs.com/issues/806):

- [x] Add an Archive button to every conversation
- [x]  When clicked, remove the conversation from /conversations
- [x]  If any conversations are archived, add a link to the bottom labeled "Archived conversations"
- [x]  When clicking this link, visit /conversations/archived and show the archived ones
- [x]  Each conversation on this screen should have an Unarchive button that moves it back to /conversations

An additional task was resolved:
- [x] After the sender sends a message & the receiver has the conversation in `archive` then it should be added to his `conversation` path.


db/schema.rb
```diff
@@ -296,7 +297,7 @@
     t.string "processor_id"
     t.boolean "default"
     t.jsonb "data"
-    t.datetime "deleted_at", precision: nil
+    t.datetime "deleted_at"
```
joemasilotti, line 300: Let's revert this change as it is relevant to the PR.

db/schema.rb
```diff
@@ -333,8 +334,8 @@
     t.string "processor_plan", null: false
     t.integer "quantity", default: 1, null: false
     t.string "status", null: false
-    t.datetime "trial_ends_at", precision: nil
-    t.datetime "ends_at", precision: nil
+    t.datetime "trial_ends_at"
+    t.datetime "ends_at"
```
joemasilotti, line 338: Same for this change.

joemasilotti (changes requested):
This looks like it is working pretty well so far!

### Turbo Streams

I'm seeing some weird issues where if I navigate back after archiving a conversation then the UI doesn't reflect the real state. I think this is because of the Turbo Stream. Let's remove that for now and do plain-old HTML updates. We can progressively enhance in a future PR.

### `ArchivedConversationsController`

But I'm not sure if I like that everything got tacked on to `ConversationsController`. Can you please move all the new logic to a new controller, `ArchivedConversationsController`, instead?

This will remove the non-RESTful routes and clean up a lot of the methods on the controller since the new one _only_ deals with archiving. I think the views, too!

You should be able to copy the routes and logic from `ReadNotificationsController` pretty closely. I think that handles everything except a `delete` route.

### Translations

Also, where did the non-English translations come from? If you are fluent in each of those languages please leave them. But if they were put through Google Translate then please remove them.

### Auto-unarchiving

I love this - good idea! I think we can simplify this by unarchiving it for both folks involved. That will ensure if I send a message from an archived conversation that it unarchives for me, too.

### Then...

Once all of these items are addressed I'll do a more in-depth review of the code. Nice work so far!

db/schema.rb
JuzerShakir: About those, I am wondering why was that removed since I didn't remove it manually. This was removed when I ran `rails db:migrate` when I was initializing this project. I don't know why! 

Will revert it! 👍🏽

JuzerShakir:
> I'm seeing some weird issues where if I navigate back after archiving a conversation then the UI doesn't reflect the real state

Ohh! I am not able to understand that, navigate back to which path does the view don't update? 
But nevertheless, if you want me to remove `turbo_stream` I will remove it and I believe instead of that you want me to implement `redirect to` each time a conversation is archived or unarchived, right?

> I'm not sure if I like that everything got tacked on to `ConversationsController`.
I knew you would say something like this because I have noticed that in other controllers too as most of the logic is in the models. I tried my best to keep the controller as skinny as possible. And I think the idea of shifting the logic of archiving to its own controller is a refactoring on the next level!

> If you are fluent in each of those languages please leave them
Naah! I ain't! I will remove it! The reason I gave it because if user changes the language mode of the archive page to other than English then it would raise an error, so to avoid that I gave Google Translation of it wherever it was necessary. Will remove them! 

> I love this - good idea! 
Thank You! ❤️ 





Thanks for all feedbacks! 😄

JuzerShakir:
Hey @joemasilotti, I have just pushed the changes you requested!

I am awaiting your feedback!

app/controllers/archived_conversations_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+    if @conversation.unarchived_by?(current_user)
+      @conversation.archive_from(user_type)
+      load_all
+      redirect_to conversations_path
+    else
+      @conversation.update("#{user_type}_archived_at": nil)
+      @conversations = load_archived
+      redirect_to archived_conversations_path
+    end
+  end
```
joemasilotti, line 21: If we instead use POST/DELETE routes, similar to `ReadNotificationsController`, we can clean up even more of this logic.

app/controllers/archived_conversations_controller.rb
```diff
@@ -0,0 +1,37 @@
...
+    @conversations = current_user.conversations.send("unarchived_by_#{user_type}")
+    @archived_any = current_user.conversations.send("archived_by_#{user_type}").exists?
+  end
+
+  def load_archived
+    current_user.conversations.send("archived_by_#{user_type}")
+  end
+
+  def user_type
+    current_user.business ? "business" : "developer"
```
joemasilotti, line 35: This logic isn't perfect. A user could have a business _and_ a developer but be messaging from their developer "account".

Instead, we can use `conversation#business?(current_user)` and its pair to determine which field to update.

app/controllers/archived_conversations_controller.rb
joemasilotti: `BlocksController` might also offer some tips for how to move some logic to the model.

app/controllers/archived_conversations_controller.rb
joemasilotti: Or check out `BlocksController`!

app/controllers/messages_controller.rb
```diff
@@ -40,6 +41,11 @@ def sender
     end
   end
 
+  def unarchive_conversation
+    conversation.update(business_archived_at: nil) if conversation.business_archived_at.present?
+    conversation.update(developer_archived_at: nil) if conversation.developer_archived_at.present?
```
joemasilotti, line 46: We can do both of these at the same time in one `UPDATE`, no need to check the value beforehand.

app/models/conversation.rb
```diff
@@ -13,6 +13,10 @@ class Conversation < ApplicationRecord
 
   scope :blocked, -> { where.not(developer_blocked_at: nil).or(Conversation.where.not(business_blocked_at: nil)) }
   scope :visible, -> { where(developer_blocked_at: nil, business_blocked_at: nil) }
+  scope :archived_by_business, -> { where.not(business_archived_at: nil).order(business_archived_at: :desc) }
+  scope :unarchived_by_business, -> { where(business_archived_at: nil).order(updated_at: :desc) }
+  scope :archived_by_developer, -> { where.not(developer_archived_at: nil).order(developer_archived_at: :desc) }
+  scope :unarchived_by_developer, -> { where(developer_archived_at: nil).order(updated_at: :desc) }
```
joemasilotti, line 19: I think that these scopes aren't the right abstraction because they are requiring you to use `#send`. I think we might be able to consolidate to 2 scopes and pass in the user - `archived_by_user` and `unarchived_by_user`, maybe?

config/locales/en.yml
```diff
@@ -180,6 +180,12 @@ en:
       search:
         cta: Search
         help: Filter users by name, company, or email
+  archived_conversations:
+    index:
+      empty_state:
+        body: Archived conversations will show up here.
+        title: No conversations
+      title: Archived Conversations
```
joemasilotti, line 188:
```suggestion
        title: No archived conversations
      title: Archived conversations
```

db/schema.rb
```diff
@@ -10,9 +10,8 @@
 #
 # It's strongly recommended that you check this file into your version control system.
 
-ActiveRecord::Schema[7.0].define(version: 2023_03_13_230358) do
+ActiveRecord::Schema[7.0].define(version: 2023_03_18_042319) do
   # These are extensions that must be enabled in order to support this database
-  enable_extension "pg_stat_statements"
```
joemasilotti, line 15: Please add this line back in, too.

app/views/conversations/_conversation.html.erb
```diff
@@ -1,24 +1,30 @@
...
-    </div>
-  <% end %>
-</li>
+      <div class="mt-1">
+        <p class="line-clamp-2 text-sm text-gray-600"><%= conversation.messages.last.body %></p>
+      </div>
+    <% end %>
+  </li>
+  <div class="text-right pt-2.5">
+   <%= button_to current_page?(conversations_path) ? "Archive" : "Unarchive",
```
joemasilotti, line 27: Please move these strings to the locale file.

joemasilotti (changes requested):
Getting closer! I think there are still a few areas for improvement: using more CRUD in the controller and cleaning up the scopes on the model.

Let me know if you have any questions.

app/controllers/archived_conversations_controller.rb
JuzerShakir: Done ✅ 

But the `create` action doesn't work in a RESTFUL style. With routes set as this:
```
    resources :archived_conversations, only: %i[index create destroy], path: "/conversations/archived"
```
when making a `post` request, it shows this error:
```
Started POST "/1/conversations/archived"
ActionController::RoutingError (No route matches [POST] "/1/conversations/archived"):
```
So I have created the `create` route manually:
```
    post "conversations/archived/:id", to: "archived_conversations#create"
```
And this works fine!

joemasilotti: I think we need two routes. One nested under conversations for CREATE and DELETE. And another to view the list at `/conversations/archived`.

app/views/conversations/_conversation.html.erb
JuzerShakir: Means? Do you mean to shift the logic of the button to `index` files of their respective controllers? If yes, then how would we have access to the `id` for each of the `@conversations`?

joemasilotti: No, move “Archive” and “Unarchive” to `en.locale.yml`.

JuzerShakir: Sorry! But I am not able to follow you. Could you elaborate on what exactly you want me to do?

joemasilotti: Plain text shouldn't exist in views, it should be moved to a translation file so different languages can use their own translation of the string.

https://guides.rubyonrails.org/i18n.html

So for this, you would add the following to `config/locales.en.yml`:

```yml
conversations:
  conversations:
    archive: Archive
    unarchive: Unarchive
```

Then in the view, line 27 becomes:

```erb
<%= button_to current_page?(conversations_path) ? t(".archive") : t(".unarchive") # ... %>
```

Does that help?

JuzerShakir: Acha you meant the text! I thought you were talking about the logic of the `archive` & `unarchive` buttons and when you said in the `locales` I was confused because that's used for translation purposes! Thanks a lot for the detailed explanation! 

Done ✅ 😄

app/controllers/archived_conversations_controller.rb
JuzerShakir: Ok so the workaround for this as you said is to use the `business?` or `developer?` method to know the type of user which is accessible by `@conversation`. Great! But I will not be able to use that method on `@conversations` or in the `index` action! So I created this logic:

```
  def user_type
    if current_user.business
      :business_archived_at
    elsif current_user.developer
      :developer_archived_at
    end
  end
```
Is this something you were looking for?

joemasilotti: Oh good point. Unfortunately that logic still has the same flaw as before - a user can have a business but be involved in a conversation from the developer’s side.

app/controllers/messages_controller.rb
JuzerShakir: Done ✅

app/models/conversation.rb
JuzerShakir: Done ✅

config/locales/en.yml
JuzerShakir: Done ✅

db/schema.rb
JuzerShakir: Done ✅

app/controllers/archived_conversations_controller.rb
JuzerShakir: The `delete` route works perfectly fine but not `create`. 

> One nested under conversations for CREATE 

Do you want me to move the `create` logic in `ConversationsController`? I don't think so & if that's the case then I haven't understood properly!

app/controllers/archived_conversations_controller.rb
JuzerShakir: Exactly! It's exactly the same logic as before but with a different return value.

I gave a little thought to it & I came up with this logic inside the `index` action of `ConversationsController`:
```
  def index
    if current_user.business
      @conversations = current_user.business.conversations.unarchived_by_user(:business_archived_at)
      @archived_any = current_user.business.conversations.archived_by_user(:business_archived_at).exists?
    else
      @conversations = current_user.developer.conversations.unarchived_by_user(:developer_archived_at)
      @archived_any = current_user.developer.conversations.archived_by_user(:developer_archived_at).exists?
    end
  end

File: app/controllers/messages_controller.rb (hunk 5 above)
Comment (JuzerShakir) [reply app/controllers/messages_controller.rb:46]: Here's the alternatives to those methods:
```
  scope :archived_by_user, -> (dev_or_business) { where("#{dev_or_business} IS NOT NULL").order("#{dev_or_business} DESC") }
  scope :unarchived_by_user, -> (dev_or_business) { where("#{dev_or_business} IS NULL").order("#{dev_or_business} DESC") }
```

File: app/models/conversation.rb (hunk 6 above)
Comment (JuzerShakir) [reply app/models/conversation.rb:19]: Here's the alternatives to those methods:
```
  scope :archived_by_user, -> (dev_or_business) { where("#{dev_or_business} IS NOT NULL").order("#{dev_or_business} DESC") }
  scope :unarchived_by_user, -> (dev_or_business) { where("#{dev_or_business} IS NULL").order("#{dev_or_business} DESC") }
```

File: app/controllers/messages_controller.rb (hunk 5 above)
Comment (JuzerShakir) [reply app/controllers/messages_controller.rb:46]:
```
    conversation.update(business_archived_at: nil, developer_archived_at: nil)
```

File: app/models/conversation.rb (hunk 6 above)
Comment (joemasilotti) [reply app/models/conversation.rb:19]: That string interpolation directly in a query scares me. Is there no way to add this to the User model so we have the scope already?

File: app/controllers/archived_conversations_controller.rb (hunk 4 above)
Comment (joemasilotti) [reply app/controllers/archived_conversations_controller.rb:35]: But if a user has both they would never see their developer conversations. 

I recognize this is a problem in the existing logic of the app. User vs. Accounts and all that. I’m not sure the best approach just yet.

File: app/controllers/archived_conversations_controller.rb (hunk 3 above)
Comment (joemasilotti) [reply app/controllers/archived_conversations_controller.rb:21]: Create should live under Archived not base Conversations. What isn’t working?

Comment (JuzerShakir) [reply app/controllers/archived_conversations_controller.rb:21]: How's this for a RESTful Resource?
```
    resources :archived_conversations, only: %i[index destroy], path: "/conversations/archived"
    resources :archived_conversations, only: %i[create], path: "/conversations/archived/:id"
```

File: app/controllers/archived_conversations_controller.rb (hunk 4 above)
Comment (JuzerShakir) [reply app/controllers/archived_conversations_controller.rb:35]: Ohh! So If I am understanding this correctly, assuming a user has both a `developer` & `business` profiles, do you want to load all the unarchived conversations in `/conversation` for both developer & business scopes right? And the same for the archived conversation?

File: app/models/conversation.rb (hunk 6 above)
Comment (JuzerShakir) [reply app/models/conversation.rb:19]: Exactly! I knew you would say this, so here's a different approach I came up with:

In the `Conversation` Model:

```
  scope :archived_by_user, (lambda do |user|
    user.business ? where.not(business_archived_at: nil) : where.not(developer_archived_at: nil)
  end)

  scope :unarchived_by_user, (lambda do |user|
    user.business ? where(business_archived_at: nil) : where(developer_archived_at: nil)
  end)
```
Instead of passing the right attribute as an argument and then using string interpolation to extract results, I thought why not just pass in the user as an argument to it. This avoids string interpolation in the scope, removes the extra logic from both the controller, and makes code more readable.

Now, of course, this doesn't solve the issue of a user having both `developer` & `business` types but we can extend that logic into this in the future.

File: app/controllers/archived_conversations_controller.rb (hunk 4 above)
Comment (JuzerShakir) [reply app/controllers/archived_conversations_controller.rb:35]: Okk! I am not sure I am still following the details here! So I want to understand this step by step:

> For example, User#archived_conversations should return conversations where the user's developer profile has archived something or the user's business has archived something.

What if the user is both a developer & business? Would you like to show archived conversations for both the developer & business in `/conversations/archived` page?

File: app/models/conversation.rb (hunk 6 above)
Comment (JuzerShakir) [reply app/models/conversation.rb:19]: To be honest, when I read the issue of this PR I thought this task wouldn't take more than a couple of hours, but boy I was wrong! 😄

File: app/controllers/archived_conversations_controller.rb (hunk 4 above)
Comment (joemasilotti) [reply app/controllers/archived_conversations_controller.rb:35]: Yes. `/conversations` should show all non-archived conversations for the user. And `/conversations/archived` should show all archived conversations for the user. Both regardless of which, if either/any, business/developer profile they have.

File: app/models/conversation.rb (hunk 6 above)
Comment (joemasilotti) [reply app/models/conversation.rb:19]: It shouldn't take more than a few hours. The whole developer vs. business thing is complicating the issue a ton. Which is no fault of yours! But is definitely calling out a bigger problem in the data modeling.

Comment (joemasilotti) [reply app/models/conversation.rb:19]: Interestingly enough, there is a single person with both developer and business conversations. Me :)

Another approach to all of this would be to have an `Archive` model joins between conversations and business/developer. I wonder if that would simplify the logic of these queries?

If not, or we don't want to explore that path right now, I'm happy to hold off on this issue until #237 is finally completed. What do you think?

Comment (JuzerShakir) [reply app/models/conversation.rb:19]: > Interestingly enough, there is a single person with both developer and business conversations. Me :)

Haha! But I believe that in the future maybe you have users who are both developers and businesses. And for that I think I have the solution and it would be just an extra check to the existing scope that I showed:

```
  scope :unarchived_by_user, (lambda do |user|
    if user.developer && user.business
      where(business_archived_at: nil, developer_archived_at: nil).order(updated_at: :desc)
    elsif user.business
      where(business_archived_at: nil).order(updated_at: :desc)
    elsif user.developer
      where(developer_archived_at: nil).order(updated_at: :desc)
    end
  end)
```
And we can share this same logic to the `archived_by_user` scope.

I will see how I can refactor this but wouldn't this solve our query, at least for now? What do you think?

Comment (joemasilotti) [reply app/models/conversation.rb:19]: I don't think that logic will work the way we want to.

Just because a user has a business account doesn't mean they are the business associated with every conversations. Meaning, I can have a business account but be involved with a conversation as a developer. If _that_ business archives the conversation then it would be considered archived by your code.

Comment (JuzerShakir) [reply app/models/conversation.rb:19]: Well! How about this?
```
  scope :archived_by_business, -> { where.not(business_archived_at: nil) }
  scope :unarchived_by_business, -> { where(business_archived_at: nil) }
  scope :archived_by_developer, -> { where.not(developer_archived_at: nil) }
  scope :unarchived_by_developer, -> { where(developer_archived_at: nil) }

  scope :unarchived_by_user, (lambda do |user|
    if user.developer && user.business
      unarchived_by_business.merge(unarchived_by_developer)
    elsif user.business
      unarchived_by_business
    elsif user.developer
      unarchived_by_developer
    end
  end)
```

Comment (joemasilotti) [reply app/models/conversation.rb:19]: Again, I don't think that logic will work. We need to check _each_ conversation to see if the user is on the business side or developer side to see which "archived at" timestamp is set or not.

I think that this comment is as close as we've gotten to getting that to work. Have you explored this?

https://github.com/joemasilotti/railsdevs.com/pull/813#discussion_r1142159773

Comment (JuzerShakir) [reply app/models/conversation.rb:19]: In the above query, as we can see we're giving or passing the `current_user` twice, in order to refactor that, and as you suggested, I have implemented that logic in the `User` model like this:
```
# app/models/user.rb

  def archived_conversations
    if business && developer
      conversations.where.not(business_archived_at: nil).merge(where.not(developer_archived_at: nil))
    elsif business
      conversations.where.not(business_archived_at: nil)
    elsif developer
      conversations.where.not(developer_archived_at: nil)
    end
  end

  def unarchived_conversations
    if developer && business
      conversations.where(business_archived_at: nil).merge(where(developer_archived_at: nil))
    elsif business
      conversations.where(business_archived_at: nil)
    elsif developer
      conversations.where(developer_archived_at: nil)
    end
  end
```

Now in our controllers we can call for those conversations like this:
```
current_user.unarchived_conversations
```
OR
```
current_user.archived_conversations
```

Comment (joemasilotti) [reply app/models/conversation.rb:19]: I think you correctly understand the goal but are missing an annoyingly tiny nuance that makes the logic need some fine tuning.

Let's take a step back and dive into how conversations are queried. Here's what currently exists on `main` in the `User` model.

```ruby
# app/models/user.rb

has_many :conversations, ->(user) {
  unscope(where: :user_id)
    .left_joins(:business, :developer)
    .where("businesses.user_id = ? OR developers.user_id = ?", user.id, user.id)
    .visible
}
```

Note that we have to match an `OR` clause to make sure this user's developer _or_ business profile is part of the conversation.

Now, in this PR, a conversation can be archived by a business and/or a developer. But it is only archived to the user if they are _that_ business or developer.

So we need to make sure that "archived" shows conversations the user archived _as a business_ or _as a developer_. Said another way, if I message you as a business you are the developer. If I archive the conversation it is _not_ archived for you.

Let's assume you also have a business profile. Then _only_ checking for the presence of `business_archived_at` means it would show as archived for you. Which is wrong. I archived it!

Which is where this code (from the comment above) comes into place. It lines up the conversations a user has as a developer with the developer archive column and the same for business.

```ruby
has_many :conversations, ->(user) {
  conversations_as_developer.merge(conversations_as_business)
}

has_many :archived_conversations, ->(user) {
  archived_conversations_as_developer.merge(archived_conversations_as_business)
}

private

has_many :conversations_as_developer, ->(user) {
  unscope(where: :user_id)
    .left_joins(:developer)
    .where(developers: {user_id: user.id})
    .where(developer_archived_at: nil)
    .visible
}

has_many :conversations_as_business, ->(user) {
  unscope(where: :user_id)
    .left_joins(:business)
    .where(businesses: {user_id: user.id})
    .where(business_archived_at: nil)
    .visible
}

has_many :archived_conversations_as_developer, ->(user) {
  unscope(where: :user_id)
    .left_joins(:developer)
    .where(developers: {user_id: user.id})
    .where.not(developer_archived_at: nil)
    .visible
}
has_many :archived_conversations_as_business, ->(user) {
  unscope(where: :user_id)
    .left_joins(:business)
    .where(businesses: {user_id: user.id})
    .where.not(business_archived_at: nil)
    .visible
}
```

Does that make more sense?

JuzerShakir: Wow! That's a detailed explanation! I never thought of the scenario where a business could also message other businesses and this is where the logic I created breaks, right?

Sorry I wasn't able to reply back to you as our holy month of Ramadan has begun so I will be much busier than usual but I will make sure to keep in touch with you every 3-4 days. 

I will be experimenting with the logic you have given soon and revert to you if I have any doubts or queries. I was also thinking maybe we can create a different model as you suggested earlier to keep track of which `user_id` has archived and which hasn't. Your thoughts?

joemasilotti: > I never thought of the scenario where a business could also message other businesses and this is where the logic I created breaks, right?

Not exactly, only a developer and a business can message each other. But a user might have a business profile but be acting as a developer in a specific conversations. I hope that distinction makes sense!

You know... I wonder if a `Conversations::Block` record would simplify a ton of this logic. Do you want to explore that path or continue trucking with what you have?

And have a generous Ramadan! Let me know if you'd like to pause on this or pass it off. I'm happy with whatever you choose but I need to make sure work continues as other folks will want to pick up stale issues.

JuzerShakir: > A user can be a business type but a developer to a conversation.

That makes so much sense now to me. I am open to any ideas you have because I don't think my logic could work in this scenario & you knew it but I couldn't understand the root of it.

Also I want to complete this PR as quick as possible. I believe this has taken much more time than expected. So yes, do let me know what you have in mind for such a problem. 

Thanks for the wishes! Happy weekend! 🙂

joemasilotti: @kaspth proposed a refactor a while back that converts "blocking" conversations to its own model. We could adopt that same approach for "archiving" conversations.

https://github.com/joemasilotti/railsdevs.com/pull/455

What I like about this approach is that it is per _user_ not developer/business. Which should make a lot of querying much simpler!

JuzerShakir: > @kaspth proposed a refactor a while back that converts "blocking" conversations to its own model. We could adopt that same approach for "archiving" conversations.
> 
> #455 

I did read it but I couldn't understand how could I use that logic here.

So I read your previous replies to come up with this solution:

*`conversations_controller.rb`*

```
  def index
    @conversations = current_user.conversations
    @archived_any = current_user.archived_conversations.exists?
  end

```

*`app/models/user.rb`*

```
  def archived_conversations
    Conversation.archived_conversations_as_developer(self).union(Conversation.archived_conversations_as_business(self))
  end

  def conversations
    Conversation.conversations_as_developer(self).union(Conversation.conversations_as_business(self))
  end
```

*`app/models/coversation.rb`*

```
  scope :conversations_as_developer, ->(user) {
    unscope(where: :user_id)
    .left_joins(:developer)
    .where(developers: {user_id: user.id})
    .where(developer_archived_at: nil)
    .visible
  }

  scope :conversations_as_business, ->(user) {
    unscope(where: :user_id)
    .left_joins(:business)
    .where(businesses: {user_id: user.id})
    .where(business_archived_at: nil)
    .visible
  }

  scope :archived_conversations_as_developer, ->(user) {
    unscope(where: :user_id)
    .left_joins(:developer)
    .where(developers: {user_id: user.id})
    .where.not(developer_archived_at: nil)
    .visible
  }

  scope :archived_conversations_as_business, ->(user) {
    unscope(where: :user_id)
    .left_joins(:business)
    .where(businesses: {user_id: user.id})
    .where.not(business_archived_at: nil)
    .visible
  }
```

You must be wondering where the `union` method comes from. That comes from the gem I installed [`active_record_union`](https://github.com/brianhempel/active_record_union). The `merge` method doesn't return the results we're looking for.

Thoughts?

kaspth: heyo, chiming in since I was mentioned. I still find the business/developer split, so I just wanted to show a sketch that tries to blend the two — and then how I'd do archiving off of that established relationship. It probably doesn't work yet, since the domain modeling here between conversations and whose the sender/recipient is still murky to me. But nonetheless, here's a sketch:

```ruby
class Account < ApplicationRecord
  delegated_type :record, types: %w[ Business Developer ]

  has_many :memberships, dependent: :destroy
  has_many :users, through: :memberships
end

class User::Membership < ApplicationRecord
  belongs_to :account
  belongs_to :user
end

class User < ApplicationRecord
  has_many :memberships
  has_many :accounts, through: :memberships # So businesses could have multiple users attached
  delegate :businesses, :developers, to: :accounts

  has_many :conversations, through: :accounts
  has_many :archivings
end

class Conversation < ApplicationRecord
  belongs_to :account
  has_many :archivings, dependent: :destroy

  scope :archived_with, ->(user) { joins(:archivings).where(archivings: { account: user.accounts }) }
  scope :visible_to,    ->(user) { left_outer_joins(:archivings).where(archivings: { account: user.accounts }) }

  def business?(user)
    user.businesses.include? account.record
  end

  def developer?(user)
    user.developers.include? account.record
  end
end

class Conversation::Archiving < ApplicationRecord
  belongs_to :conversation
  belongs_to :account
  belongs_to :user
end

class Conversations::ArchivingsController < ApplicationController

  def index
    @conversations = Conversation.archived_with(current_user)
  end

  def create
    conversation.archivings.create!(account:, user: current_user)
    redirect_to conversations_url, notice: "Conversation archived"
  end

  def destroy
    conversation.archivings.destroy_by!(account:)
    redirect_to conversations_url, notice: "Conversation unarchived"
  end

  private

  def conversation
    @conversation ||= Conversation.find(params[:id])
  end

  def account
    @account ||= current_user.accounts.find(params[:account_id])
  end
end

class ConversationsController < ApplicationController

  def index
    @conversations = Conversation.visible_to(current_user)
    @archived_any = current_user.archivings.exists?
  end

  def show
    authorize @conversation = Conversation.find(params[:id])

    @conversation.mark_notifications_as_read(current_user)

    @message = @conversation.messages.new
  end
end

# Potential alternative sketch where an archiving and a block are really considered the same type of model,
# because we're connecting an action on a conversation between a user and an account.
# Didn't get far with this.
class Conversation::Involvement < ApplicationRecord
  belongs_to :conversation
  belongs_to :account, default: -> { conversation&.account }
  belongs_to :user
  
  enum kind: %i[ block archive ].index_by(&:itself)
end

class Conversations::InvolvementsController < ApplicationController
  def create
  end
end
```

joemasilotti: Thanks @kaspth! Once RailsDevs has proper accounts (instead of only users) this looks pretty ideal to me. Unfortunately, there is a lot more work to get us there right now.

This remodeling also starts to immediately open up interesting product questions. Does a user or account archive a conversation? What if a team wants to share a workflow and message developers together? What about blocks? Who _owns_ one of those? Interesting stuff!

@JuzerShakir, two questions:

1. What is the difference between a union and a merge? Could an `or` work just the same? I'm not sure I understand the nuances here.
2. I'm starting to think we might want to postpone this PR until #410 is finally live. Someone is starting to work on that now but I don't have a good timeline on when it might be done.

What do you think?

JuzerShakir: Answering  your questions:

**Could an `or` work the same?**
Yes! It does. `union` & `or` are quite similar where both are used to combine the results of multiple queries. I used the `or` method in our `scopes` but since it performs a `left_outer_join`, it shows the following error:
Screenshot from 2023-03-26 11-31-37

-----

**What is the difference between a union and a merge?**
Applying `merge` doesn't return any errors but it doesn't return the results we are looking for because when we call `merge` on a relation object as an argument, **ActiveRecord will combine the conditions of both relations into a single SQL query**. For example, consider the following two relation objects:
```
relation1 = User.where(age: 20)
relation2 = User.where(city: 'New York')
```
If we want to combine these two relation objects into a single relation object that contains all the users who are 20 years old and live in New York, we can use the `merge` method. But in our case, we don't want to combine the conditions but evaluate and return the queries separately and then return a single `ActiveRecord::Relation` object containing the results of both queries.

So in order to combine the results of those queries into a single result I could use `+` method which would give results that we expect but it would return an `Array` object instead of an `ActiveRecord::Relation` where we wouldn't be able to use any ActiveRecord methods or make SQL queries.

And that's where the `union` method comes in. The `union` method is used to combine the results of two or more queries into a single relation object with duplicates removed. The columns in each query must be in the same order and have the same data types for the `union` to work correctly. And the `active_record_union` gem enables us to use the `union` method to return the results we're looking for.

I hope this clears your doubt and explains my reasoning well for using the `union` query.

joemasilotti: That does - thanks for the detailed explanation!

What are your thoughts on the second question?

JuzerShakir: Yeah sorry @joemasilotti, about that, I think you should do what you think is best. But I have a question, if this PR gets postponed will I be assigned to other issues that interest me for railsdev.com?

joemasilotti: OK, cool. Let's postpone this issue for now. I'll mark it blocked until #843 is done.

And yes, 100%. You are free to pick up any issue that is available. I'm also happy to pay you the bounty for this one since you put in so much work already. [Send me an email](mailto:joe@railsdevs.com) to coordinate.

joemasilotti:
Closed until #843 is released. There are too many complex queries and edge cases if we don't know if the current user is acting as a business or developer.

app/models/conversation.rb
JuzerShakir: Thanks, a lot @joemasilotti ❤️

PR #816, opened 2023-03-18
Add Tab Ordering and AutoFocus on sign in and Registration Page

aj-abhishek-jadav:
This PR address:

- [x] Set the tab order on the sign in screen to email then password then forgot password
- [x] Set the tab order on the registration screen to email then password then password confirmation
- [x] Autofocus on the email field for Sign in, Registration, Forgot Password and Resend confirmation page

_Note: Tab Ordering on registration screen to email then password then password confirmation is already maintained because of its location in source code (these fields are in ascending order as per code) & hence changes are not required_

### <ins> Screen Recording of Autofocus on mentioned pages

https://user-images.githubusercontent.com/36496890/226115603-188d1edc-8aaf-4357-b4ab-5868cd8bf2a4.mov

### <ins> Screen Recording of Tab Ordering on Sign in page

https://user-images.githubusercontent.com/36496890/226115920-3445bc81-261a-48e6-b7ce-cbec1af5ba5e.mov

### DIFF
--- a/app/views/devise/confirmations/new.html.erb
+++ b/app/views/devise/confirmations/new.html.erb
@@ -15,7 +15,7 @@
         <div>
           <%= form.label :email, class: "block text-sm font-medium text-gray-700" %>
           <div class="mt-1">
-            <%= form.email_field :email, autocomplete: "email", required: true, class: "appearance-none block w-full px-3 py-2 border border-gray-300 rounded-md shadow-sm placeholder-gray-400 focus:outline-none focus:ring-gray-500 focus:border-gray-500 sm:text-sm" %>
+            <%= form.email_field :email, autocomplete: "email", required: true, class: "appearance-none block w-full px-3 py-2 border border-gray-300 rounded-md shadow-sm placeholder-gray-400 focus:outline-none focus:ring-gray-500 focus:border-gray-500 sm:text-sm", autofocus: true %>
           </div>
         </div>
 
--- a/app/views/devise/passwords/new.html.erb
+++ b/app/views/devise/passwords/new.html.erb
@@ -17,7 +17,7 @@
         <div>
           <%= form.label :email, class: "block text-sm font-medium text-gray-700" %>
           <div class="mt-1">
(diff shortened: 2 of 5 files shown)


joemasilotti:
Hey @aj-abhishek-jadav, is this ready for review? If not, please convert it to a draft PR and mark as ready for review when you'd like me to take a look. Thanks!

aj-abhishek-jadav:
Hi @joemasilotti, I m ready with PR. Request you to please Review and let me know if changes are needed.
Btw this is my 1st Open source contribution. Thank you so much for giving this opportunity 😍 

Thanks ❤️

joemasilotti (approved):

brunoprietog:
Hi, I'm sorry to tell you that this change introduces an accessibility issue. The tab order should always be the order in which the elements are defined. If a screen reader user presses tab after the sign in button, he should go to English link, and if he presses shift tab after the field to enter his email, he should go to create an account link. Pressing tab from the top of the page to the end of the page should reflect the logical order. If a user is on the link to create an account and presses tab, he would expect to find the field to enter the email, but instead, he will skip the entire login form and go directly to the English link.

The tabindex attribute should very rarely have values other than 0 or -1. The 0 is used so that elements that are not focusable by the browser by default, such as a div, for example, are focusable by the keyboard and are added to the corresponding tab order. The -1, serves so that an element is keyboard focusable, but is removed from the tab order. This is useful to make a poppup menu, for example, where you want that the elements of the menu are focusable, but that only one of them is in the tab order, the one that is active at that moment.

Actually the solution in this case is that these elements do not have the tabindex attribute at all. If the email field has the autofocus attribute, it will focus automatically. The problem is that when you press tab, it will focus on the forgot password link instead of the password field. The best thing to do in these cases is to change the design, perhaps placing the link in another place that makes sense with the tab order.

I hope this is useful!

joemasilotti:
Thanks for chiming in! Can you please provide a link showing that breaking "tab order should always be the order in which the elements are defined" is an accessibility issue?

brunoprietog:
Sure, there is. The second is a rule of axe-core, the best accessibility linter. Also, I tell you this as a blind user and developer.

<https://webaim.org/techniques/keyboard/tabindex>

<https://dequeuniversity.com/rules/axe/4.4/tabindex>

joemasilotti:
Wow, TIL! Thanks for the links and bringing this up.

Would you like to open a PR that removes the tab indexes from the form? No pressure - if not I'll take care of it myself this week.

PR #819, opened 2023-03-18
Use upsert_all in Specialities seeds

fractaledmind:
## Pull request checklist


joemasilotti:
Thanks @fractaledmind! Unfortunately it looks like I would need to change the YAML file to be:

```yml
- name: AI
- name: AWS
# ...
``

Which, for now, really breaks the readability in my opinion. And isn't worth the trade-off for a slightly quicker database seed.

If/when more columns are added to `specialties` and the YAML file _must_ be hashes then I'd like to merge this in!

PR #821, opened 2023-03-19
Refactor DeveloperQuery

### LINKED ISSUE #805: Refactor `DeveloperQuery` to be closed for modification
`DeveloperQuery` is becoming increasingly more difficult to work with. There is a _lot_ of logic in that class.

I'd like to refactor it to to receive an array of `QueryItem` objects. This way, each query item can handle its own scoping, sorting, and filtering. This should make it easier to extend in the future when new query options are required, like #803.

soumyashree25:
closes #805 

## Pull request checklist

### DIFF
--- a/app/queries/developers/query_item_builder.rb
+++ b/app/queries/developers/query_item_builder.rb
@@ -2,34 +2,15 @@ class Developers::QueryItemBuilder
 
   attr_reader :query_items, :options
 
-  QUERY_ITEM_OPTIONS = %i[sort countries]
-
   def initialize(options = {})
+    options[:countries] = [] unless options.has_key?(:countries)
+    options[:sort] = nil unless options.has_key?(:sort)
     @options = options
     @query_items = []
-  end
 
-  def build_query_items
-    add_sorting_query
-    add_countries_filter_query
-
-    query_items
-  end
-
-  def add_sorting_query
-    query_items << Developers::SortQueryItem.new(options[:sort])
-  end
-
-  def add_countries_filter_query
-    query_items << Developers::CountriesQueryItem.new(options[:countries])
-  end
-
-  # defines query item methods like sort_query_item, countries_query_item
-  # which can give us the respective query item object
-  # example: query_item_builder.sort_query_item
-  QUERY_ITEM_OPTIONS.each do |method_initial, _|
-    define_method "#{method_initial}_query_item" do
-      query_items.detect { |query_item| query_item.type.has_key? method_initial }
+    options.each do |query_item_type, query_item_value|
+      query_item_type = query_item_type.to_s.capitalize
+      query_items << "Developers::#{query_item_type}QueryItem".constantize.new(query_item_value)
     end
   end
 end
--- a/app/queries/developers/query_item_builder.rb
+++ b/app/queries/developers/query_item_builder.rb
@@ -1,0 +1,35 @@
+class Developers::QueryItemBuilder
+
+  attr_reader :query_items, :options
+
+  QUERY_ITEM_OPTIONS = %i[sort countries]
+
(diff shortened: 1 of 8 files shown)


soumyashree25:
Hi @joemasilotti ,  I have extracted sort and countries filter from `DeveloperQuery` as of now in this PR.
Can you review once, I want to make sure this PR is aligned with your expectation.

app/queries/developers/query_item_builder.rb
```diff
@@ -0,0 +1,35 @@
...
+  QUERY_ITEM_OPTIONS = %i[sort countries]
+
+  def initialize(options = {})
+    @options = options
+    @query_items = []
+  end
+
+  def build_query_items
+    add_sorting_query
+    add_countries_filter_query
```
joemasilotti, line 14: I think that these two should be passed in as parameters to `initialize`. That way we can pass in any combination we want - and future queries only require changes to the new query item and not this class.

app/queries/developers/query_item_builder.rb
```diff
@@ -0,0 +1,35 @@
...
+  end
+
+  # defines query item methods like sort_query_item, countries_query_item
+  # which can give us the respective query item object
+  # example: query_item_builder.sort_query_item
+  QUERY_ITEM_OPTIONS.each do |method_initial, _|
+    define_method "#{method_initial}_query_item" do
+      query_items.detect { |query_item| query_item.type.has_key? method_initial }
+    end
+  end
```
joemasilotti, line 34: This feels like it is breaking the point of the refactor. If the public interface exposes the different filters by name then anyone using the query needs to know about reach filter.

soumyashree25: @joemasilotti since methods like `sort`, `countries` etc are called on developer query object in many places (example: https://github.com/joemasilotti/railsdevs.com/pull/821/files#diff-9b36d228956c34411c77b9a8f1fc0f9831f453e3c4fb52b050673af85a6ee584R14) , somehow we have to expose those either in developer query or query builder. 

Any thoughts on how to achieve that if we don't get access to `sort_query_item` or `countries_query_item` ?

joemasilotti: One approach is for the query to expose an array of query items publicly. Then, the view can iterate over those to generate the necessary form elements.

The query item can indicate what type of field it requires (text box, select, checkbox, etc.) along with the copy and notes needed for the UI. It can also expose a method that decides if a collapse should be expanded or not.

Does that make sense?

soumyashree25: > One approach is for the query to expose an array of query items publicly. Then, the view can iterate over those to generate the necessary form elements.
> 
@joemasilotti made adjustments to countries filter and sort query items, let me know if this is fine.

 
> The query item can indicate what type of field it requires (text box, select, checkbox, etc.) along with the copy and notes needed for the UI. It can also expose a method that decides if a collapse should be expanded or not.
> 
@joemasilotti can you please explain a bit more what do you mean by indicating tags like checkbox, select, etc ?
Should we let the query item return its own html  and them embed it to `app/components/developers/query_component.html.erb` ?

joemasilotti: Thanks for the update!

In an ideal world we could pass in a bunch of `QueryItem` duck types and a single `SortItem` into `DeveloperQuery` and that would be that. The HTML could be generated based entirely by iterating over these items and inferring the type.

But the more I think about it the more I'm wondering if a) I'm overcomplicating it and b) we are rebuilding ransack.

I appreciate the work you put into this PR. But I think I'd like to explore ransack next. I'd still love to award a bounty - [send me an email](mailto:joe@railsdevs.com) and we can figure it out.

PR #822, opened 2023-03-20
Fixes record duplication in DeveloperQuery

bschrag620:
I believe the current implementation of the specialty filters in `DeveloperQuery` is incorrectly returning multiple the same developer when they have multiple matches to the selected specialties. Take this from the test case as an example:
```
hotwire_developer = create_developer(specialty_ids: [turbo.id, stimulus.id])
records = DeveloperQuery.new(specialty_ids: [turbo.id, stimulus.id]).records
records.length => 2
records == [hotwire_developer, hotwire_developer] => true
```

This PR is adding a `with_specialties` scope that accepts a or many specialty ids and uses a subquery to retrieve a single record, even if the record has many matches in specialties. The specialty filter test case was extended to check for uniqueness as well.

* changes the where clause used for the specialties filter so that the same user isn't returned multiple times

## Pull request checklist

### DIFF
--- a/app/models/developer.rb
+++ b/app/models/developer.rb
@@ -74,6 +74,16 @@ class Developer < ApplicationRecord
   scope :newest_first, -> { order(created_at: :desc) }
   scope :profile_reminder_notifications, -> { where(profile_reminder_notifications: true) }
   scope :visible, -> { where.not(search_status: :invisible).or(where(search_status: nil)) }
+  scope :with_specialty_ids, ->(specialty_ids) {
+    where_sql = <<-SQL
+      exists (
+        select 1 from specialty_tags
+        where specialty_tags.specialty_id in (?)
+          and specialty_tags.developer_id = developers.id
+      )
+    SQL
+    where(where_sql, Array.wrap(specialty_ids))
+  }
 
   def visible?
     !invisible?
--- a/test/queries/developer_query_test.rb
+++ b/test/queries/developer_query_test.rb
@@ -181,6 +181,7 @@ class DeveloperQueryTest < ActiveSupport::TestCase
     assert_includes records, stimulus_developer
     refute_includes records, react_developer
     refute_includes records, developer
+    assert_equal records.length, records.uniq.length
   end
 
   test "filtering developers by their bio or hero does not includes all if business has an active subscription" do
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 ### March
 
+* March 20 - Update developer query to not duplicate records with multiple matching specialties #822 @bschrag620
 * March 18 - Add developer specialties #803
 * March 16 - Don't retry emails for invalid/unsubscribed recipients #798 @zzJZzz
 * March 15 - Fix Preview image doesn't change when select new avatar in profile page #787 @ryy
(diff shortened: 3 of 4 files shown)


app/models/developer.rb
```diff
@@ -74,6 +74,16 @@ class Developer < ApplicationRecord
...
   scope :profile_reminder_notifications, -> { where(profile_reminder_notifications: true) }
   scope :visible, -> { where.not(search_status: :invisible).or(where(search_status: nil)) }
+  scope :with_specialties, ->(specialty_ids) {
+    where_sql = <<-SQL
+      exists (
+        select 1 from specialty_tags
+        where specialty_tags.specialty_id in (?)
+          and specialty_tags.developer_id = developers.id
+      )
+    SQL
```
bschrag620, line 84: By using `where exists`, a developer record is returned if they have any of the specialties. This works because not only are we getting the specialty_tags that are in the specialty_id array, we are also restricting the specialty_tags to the developers in the outer query.

test/queries/developer_query_test.rb
```diff
@@ -181,6 +181,7 @@ class DeveloperQueryTest < ActiveSupport::TestCase
     assert_includes records, stimulus_developer
     refute_includes records, react_developer
     refute_includes records, developer
+    assert_equal records.length, records.uniq.length
```
bschrag620, line 184: Ensure there is no record duplication. Prior to the change above, the addition of this test was causing a failure in the test suite because the `hotwire_developer` was selected twice since they had more than 1 matching specialty.

app/models/developer.rb
```diff
@@ -74,6 +74,16 @@ class Developer < ApplicationRecord
   scope :newest_first, -> { order(created_at: :desc) }
   scope :profile_reminder_notifications, -> { where(profile_reminder_notifications: true) }
   scope :visible, -> { where.not(search_status: :invisible).or(where(search_status: nil)) }
+  scope :with_specialties, ->(specialty_ids) {
```
joemasilotti, line 77:
```suggestion
  scope :with_specialty_ids, ->(specialty_ids) {
```

joemasilotti (changes requested):
Thanks for this! I can see in the tests how the old query is returning duplicates. But if I filter by "Automated testing" in the UI I only see one of each matching dev.

Any idea why that might be happening?

bschrag620:
> Any idea why that might be happening?

I'll look into it. I'm not sure I quite understand the issue just yet.

joemasilotti:
Thanks! If it helps...

The failing test makes me think we should be seeing duplicate developers if they match a queried specialty _and_ have additional ones.

But I'm not seeing that in development.

image

bschrag620:
> The failing test makes me think we should be seeing duplicate developers if they match a queried specialty _and_ have additional ones.
> 
> But I'm not seeing that in development.

Where I _think_ you should see the duplicates is when selecting more than 1 specialty AND the developer matches all selected specialties. So in the screenshot above, if you select AWS, Android, and Automated Testing I would anticipate the `DeveloperQuery` returning Core Developer 3 times.

That said, I'm not able to get the form to render properly in my local environment. The checkboxes are rendering with the name `specilaty_ids[]` instead of `developer[specialty_ids][]`. So when Applying the filters, the `specialty_ids` are seen as unpermitted since they aren't properly namespaced behind `developer.

I'll see if I can hack around with the naming to at least verify the duplication before digging in much further.

joemasilotti (approved):

joemasilotti:
I offer a $25 bounty for every issue/PR submitted to RailsDevs. Can you [email me](mailto:joe@railsdevs.com) the best way to pay you?

PR #823, opened 2023-03-20
Text search also searches specialities

nnahgodspower122:
updated developers model to include specialties

### DIFF
--- a/app/models/developer.rb
+++ b/app/models/developer.rb
@@ -24,6 +24,7 @@ class Developer < ApplicationRecord
   has_one :referring_user, through: :user
   has_many :conversations, -> { visible }
   has_many :messages, -> { where(sender_type: Developer.name) }, through: :conversations
+  has_many :specialty, through: :specialty_tags
   has_many :hired_forms, class_name: "Hired::Form", dependent: :destroy
   has_one :location, dependent: :destroy, autosave: true
   has_one :role_level, dependent: :destroy, autosave: true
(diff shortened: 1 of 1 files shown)


app/models/developer.rb
```diff
@@ -24,6 +24,7 @@ class Developer < ApplicationRecord
   has_one :referring_user, through: :user
   has_many :conversations, -> { visible }
   has_many :messages, -> { where(sender_type: Developer.name) }, through: :conversations
+  has_many :specialty, through: :specialty_tags
```
bschrag620, line 27: I think this should be pluralized. You may also need to supply a `source`. I'm a bit rusty on the exact association definition here.
```
has_many :specialties, through: :specialty_tags, source: :specialty
```

joemasilotti: I'm not sure why this is needed at all. Up at the top of this file we have `HasSpecialties` which contains `has_many :specialties, -> { visible }, through: :specialty_tags, dependent: :destroy`.

Can you please remove this line and update the query to use the existing relationship?

app/models/developer.rb
```diff
@@ -43,7 +44,7 @@ class Developer < ApplicationRecord
   validates :name, presence: true
   validates :response_rate, numericality: {greater_than_or_equal_to: 0, less_than_or_equal_to: 100}
 
-  pg_search_scope :filter_by_search_query, against: [:bio, :hero], using: {tsearch: {tsvector_column: :textsearchable_index_col}}
+  pg_search_scope :filter_by_search_query, against: [:bio, :hero], associated_against: {specialty: :name}, using: {tsearch: {tsvector_column: :textsearchable_index_col, prefix: true}}
```
joemasilotti, line 47: Nice! Do we need to update the index at all for this work?

joemasilotti (changes requested):

app/models/developer.rb
nnahgodspower122: hi @joemasilotti  I dont think we need to update the index

joemasilotti: Excellent! Thanks for confirming.

joemasilotti (approved):
Perfect!

joemasilotti:
Thanks for the contribution, @nnahgodspower122!

Can you [email me](mailto:joe@railsdevs.com) and let me know how you'd like to receive the bounty? I can do GitHub Sponsorships, Venmo, PayPal, Zelle, and Wise.

PR #829, opened 2023-03-23
Fix seeds after migration

### LINKED ISSUE #826: Fresh database reset and migration breaks seeds
## Describe the bug

Migrating a new database then running the seeds raises an error.

## To reproduce

1. `bin/rails db:drop db:create db:migrate db:seed`

## Expected behavior

The database is created, migrated, and seeded without issue.

## Actual behavior

An error is raised:

```
NoMethodError: undefined method `referral_code' for #<User id: nil, email: "admin@example.com", created_at: nil, updated_at: nil, admin: true, authentication_token: nil>
Did you mean?  referral_ids
               referral_ids=
/Users/joemasilotti/workspace/projects/railsdevs/rails/db/seeds/admins.rb:10:in `<top (required)>'
/Users/joemasilotti/workspace/projects/railsdevs/rails/db/seeds.rb:4:in `load'
/Users/joemasilotti/workspace/projects/railsdevs/rails/db/seeds.rb:4:in `seed'
/Users/joemasilotti/workspace/projects/railsdevs/rails/db/seeds.rb:16:in `<top (required)>'
<internal:/Users/joemasilotti/.rbenv/versions/3.1.2/lib/ruby/3.1.0/rubygems/core_ext/kernel_require.rb>:85:in `require'
<internal:/Users/joemasilotti/.rbenv/versions/3.1.2/lib/ruby/3.1.0/rubygems/core_ext/kernel_require.rb>:85:in `require'
-e:1:in `<main>'
Tasks: TOP => db:seed
(See full trace by running task with --trace)
```

## Additional context

`bin/rails db:reset` works without issue. IIRC, this loads the schema directly and doesn't walk through every migration.

jkostolansky:
This PR fixes #826 - two issues preventing running the following command:
```
bin/rails db:drop db:create db:migrate db:seed
```

---

### Issue 1: Using ActiveRecord models in migrations

The rake task `db:seed` works without issue, but the combination `db:migrate db:seed` throws an error. This means there is some residual cached data available during the seeding process originating in the migrations.

There are multiple places in the migrations where ActiveRecord models are used directly. This is not a good practice for various reasons - and one of them is that after "touching" (using) a model, some column information is lazy loaded and cached. And if the corresponding database table is modified in the following migrations, the cached model doesn't know about it. That's why the method [`reset_column_information`](https://apidock.com/rails/ActiveRecord/Base/reset_column_information/class) exists.

I fixed it by defining and using temporary models directly in these migrations.

Commit: https://github.com/joemasilotti/railsdevs.com/commit/a799c2c6a30d24fc80ace62cd362ef6305a93885

---

### Issue 2: Trying to schedule future jobs with `inline` queue adapter

After fixing the previous issue, the task `db:migrate db:seed` produces the following error:
```
NotImplementedError: Use a queueing backend to enqueue jobs in the future.
railsdevs.com/app/models/concerns/messages/notifications.rb:36:in `update_developer_response_rate'
```

The `queue_adapter` is set to `inline` in the [`seeds.rb`](https://github.com/joemasilotti/railsdevs.com/blob/e2d1e6459fda6a53926041b68d0ca87f36a8c6a2/db/seeds.rb#L8-L13) file for all jobs except the `UpdateDeveloperResponseRateJob` - this one is set to `async` adapter. This job is scheduled during the seeding process using `set(wait:).perform_later`.

However, it seems that there are more jobs scheduled in the future with the mentioned one - related (possibly) to ActiveStorage, ActionMailer, Noticed etc. They use `inline` adapter, what is producing the `NotImplementedError`.

The issue occurs - once again - only when running both `db:migrate db:seed`. I don't know exactly why, it could be related to some loading magic in Rails.

The only problematic job (scheduled in the future during seeding) is `UpdateDeveloperResponseRateJob`. And in the development environment, the `wait` parameter is set to [`0.seconds`](https://github.com/joemasilotti/railsdevs.com/blob/e2d1e6459fda6a53926041b68d0ca87f36a8c6a2/config/environments/development.rb#L90) . So if the seeds are run only in development, we can easily fix it by setting the `wait` parameter to `nil` instead of `0.seconds`. This will call the `enqueue` method instead of the `enqueue_at` which is not implemented in the `inline` adapter ([related code](https://github.com/rails/rails/blob/2cef3ac45bd85e1642b1270ab4fa31629044c962/activejob/lib/active_job/queue_adapters/inline_adapter.rb#L14-L20)). 

Commit: https://github.com/joemasilotti/railsdevs.com/commit/87083be0a0329f2fa943dd837ba6b9b56e28d434

### DIFF
--- a/app/jobs/update_developer_response_rate_job.rb
+++ b/app/jobs/update_developer_response_rate_job.rb
@@ -1,8 +1,6 @@
 class UpdateDeveloperResponseRateJob < ApplicationJob
   queue_as :default
 
-  GRACE_PERIOD = Rails.application.config.developer_response_grace_period
-
   def perform(developer)
     conversations = eligable_conversations(developer)
     replied_rate = Stats::Conversation.new(conversations).replied_rate
@@ -17,8 +15,11 @@ class UpdateDeveloperResponseRateJob < ApplicationJob
   private
 
   def eligable_conversations(developer)
+    grace_period = Rails.application.config.developer_response_grace_period
+    grace_period_start = grace_period.nil? ? Time.current : grace_period.ago
+
     developer.conversations.reject do |conversation|
-      conversation.created_at > GRACE_PERIOD.ago && !conversation.developer_replied?
+      conversation.created_at > grace_period_start && !conversation.developer_replied?
     end
   end
 end
--- a/app/models/concerns/messages/notifications.rb
+++ b/app/models/concerns/messages/notifications.rb
@@ -31,9 +31,8 @@ module Messages
     end
 
     def update_developer_response_rate
-      UpdateDeveloperResponseRateJob
-        .set(wait: Rails.application.config.developer_response_grace_period)
-        .perform_later(developer)
+      wait = Rails.application.config.developer_response_grace_period
+      UpdateDeveloperResponseRateJob.set(wait:).perform_later(developer)
     end
 
     def first_message?
--- a/test/jobs/update_developer_response_rate_job_test.rb
+++ b/test/jobs/update_developer_response_rate_job_test.rb
@@ -3,8 +3,6 @@ require "test_helper"
 class UpdateDeveloperResponseRateJobTest < ActiveJob::TestCase
   include BusinessesHelper
 
-  GRACE_PERIOD = UpdateDeveloperResponseRateJob::GRACE_PERIOD
-
   setup do
     @developer = developers(:one)
     create_answered_message_for(@developer)
@@ -56,7 +54,13 @@ class UpdateDeveloperResponseRateJobTest < ActiveJob::TestCase
   def start_conversation(developer, grace_period_expired)
     business = Business.create!(business_attributes)
     conversation = Conversation.create!(developer:, business:)
-    conversation.update!(created_at: GRACE_PERIOD.ago - 1.hour) if grace_period_expired
+
+    if grace_period_expired
+      grace_period = Rails.application.config.developer_response_grace_period
+      grace_period_start = grace_period.nil? ? Time.current : grace_period.ago
+      conversation.update!(created_at: grace_period_start - 1.hour)
+    end
+
     Message.create!(conversation:, sender: business, body: "First contact")
     conversation
   end
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -85,9 +85,9 @@ Rails.application.configure do
   # Enable local development exposed via ngrok for webhook testing.
   config.hosts << /.*\.ngrok\.io/
 
-  # Set a zero value for developer response grace period in development to allow seeding without leaving
+  # Disable developer response grace period in development to allow seeding without leaving
   # unactioned jobs in the queue.
-  config.developer_response_grace_period = 0.seconds
+  config.developer_response_grace_period = nil
 
   # Disable rate limiting for local development.
   Rack::Attack.enabled = false
--- a/db/migrate/20211118171626_create_role_types.rb
+++ b/db/migrate/20211118171626_create_role_types.rb
@@ -1,4 +1,10 @@
 class CreateRoleTypes < ActiveRecord::Migration[7.0]
+  class Developer < ActiveRecord::Base
+  end
+
+  class RoleType < ActiveRecord::Base
+  end
+
   def change
     create_table :role_types do |t|
       t.belongs_to :developer, index: {unique: true}, foreign_key: true
@@ -11,7 +17,7 @@ class CreateRoleTypes < ActiveRecord::Migration[7.0]
     end
 
     Developer.find_each do |developer|
-      developer.create_role_type!
+      RoleType.create!(developer_id: developer.id)
     end
   end
 end
--- a/db/migrate/20220610180608_add_inbound_email_token_to_conversations.rb
+++ b/db/migrate/20220610180608_add_inbound_email_token_to_conversations.rb
@@ -1,4 +1,8 @@
 class AddInboundEmailTokenToConversations < ActiveRecord::Migration[7.0]
+  class Conversation < ActiveRecord::Base
+    has_secure_token :inbound_email_token
+  end
+
   def change
     add_column :conversations, :inbound_email_token, :string
     Conversation.find_each.map(&:regenerate_inbound_email_token)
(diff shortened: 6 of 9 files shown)


app/models/concerns/messages/notifications.rb
```diff
@@ -31,9 +31,8 @@ def schedule_celebration_promotion
     end
 
     def update_developer_response_rate
-      UpdateDeveloperResponseRateJob
-        .set(wait: Rails.application.config.developer_response_grace_period)
-        .perform_later(developer)
+      wait = Rails.application.config.developer_response_grace_period
+      UpdateDeveloperResponseRateJob.set(wait:).perform_later(developer)
```
joemasilotti, line 35: I'm confused by this diff - is anything changing outside of code style/formatting?

app/jobs/update_developer_response_rate_job.rb
```diff
@@ -17,8 +15,11 @@ def perform(developer)
   private
 
   def eligable_conversations(developer)
+    grace_period = Rails.application.config.developer_response_grace_period
+    grace_period_start = grace_period.nil? ? Time.current : grace_period.ago
```
joemasilotti, line 19: I'm a little nervous about this conditional because it means the only time `grace_period.ago` is called is in production.

Do you have any ideas on how we could test that codepath?

test/jobs/update_developer_response_rate_job_test.rb
```diff
@@ -56,7 +54,13 @@ def create_ignored_message_for(developer, grace_period_expired: false)
   def start_conversation(developer, grace_period_expired)
     business = Business.create!(business_attributes)
     conversation = Conversation.create!(developer:, business:)
-    conversation.update!(created_at: GRACE_PERIOD.ago - 1.hour) if grace_period_expired
+
+    if grace_period_expired
+      grace_period = Rails.application.config.developer_response_grace_period
+      grace_period_start = grace_period.nil? ? Time.current : grace_period.ago
+      conversation.update!(created_at: grace_period_start - 1.hour)
+    end
```
joemasilotti, line 62:
```suggestion
    grace_period = Rails.application.config.developer_response_grace_period || 0.seconds
    conversation.update!(created_at: grace_period.ago - 1.hour)
```

I'm fine with always updating the conversation to remove a conditional from the test suite.

app/jobs/update_developer_response_rate_job.rb
joemasilotti: Also, we might be able to simplify this a bit with:

grace_period = Rails.application.config.developer_response_grace_period || 0.seconds
conversation.created_at > grace_period.ago && !conversation.developer_replied?

joemasilotti (changes requested):
Thanks for this @jkostolansky! I appreciate the deep dive into the seeds.

I've never used "phantom" Active Record classes like that. Great idea.

I left some comments and a few code style requests. Let me know if you have any questions!

app/models/concerns/messages/notifications.rb
jkostolansky: I touched this part during the debugging process. No functional change, only code style. I've left it consistent with a [similar method above](https://github.com/joemasilotti/railsdevs.com/blob/d824ddbd084c5d076ab7b19f78a304f4babfdf67/app/models/concerns/messages/notifications.rb#L28-L36). I can revert it if you want.

app/jobs/update_developer_response_rate_job.rb
jkostolansky: > we might be able to simplify this a bit with

Nice! I've changed that.

> The only time grace_period.ago is called is in production.

The `developer_response_grace_period` is set to `nil` only in development, not during the tests. That's why I set the config to `nil` in [seeds](https://github.com/joemasilotti/railsdevs.com/blob/d824ddbd084c5d076ab7b19f78a304f4babfdf67/db/seeds.rb#L12-L13) because there are some smoke tests in Github Action that were [failing](https://github.com/joemasilotti/railsdevs.com/actions/runs/4495813460/jobs/7909850263).

Now I'm thinking... If the the grace period was set to `0.seconds` in development only because of the seeds, now it can be any value (0+ seconds). 

Should I change the grace period in development? What value should I use?

test/jobs/update_developer_response_rate_job_test.rb
jkostolansky: The condition `if grace_period_expired` can't be removed, it's part of the test suite - it determines whether the `start_conversation` helper method should create new or old message. It was there before my changes.

But I've changed the `grace_period || 0.seconds` part. 👍

joemasilotti: Why not always update an existing message though? Wouldn't that put us in the same known state?

app/jobs/update_developer_response_rate_job.rb
jkostolansky: Done!

test/jobs/update_developer_response_rate_job_test.rb
jkostolansky: I'm not sure if I understand correctly. The method `start_conversation` is used in different tests for two things - based on a method argument:
- It creates a conversation with `created_at` _before_ the beginning of the grace period
- It creates a conversation with `created_at` _after_ the beginning of the grace period

The `if grace_period_expired` condition (based on the method argument) determines which scenario should be used. If the `grace_period_expired` is `true`, the conversation is updated with `created_at` 1 hour _before_ the beginning of the grace period. If it's `false`, it's just not updated and the `created_at` is implicitly "now", so _after_ the beginning of the grace period.

Why do you want to always update the conversation? How do you want to handle the two scenarios without a condition?

joemasilotti: _OH!_ My mistake! I totally missed that this was a helper. I thought it was a test. 😆

Thanks for clarifying. This is good to merge.

joemasilotti (approved):

joemasilotti:
Thanks for your contribution, @jkostolansky! Please [email me](mailto:joe@railsdevs.com) how you'd like to receive your bounty payout. I can do Venmo, PayPal, Zelle, or GitHub Sponsorships.

PR #836, opened 2023-03-24
Refactor Hired::Form into Developers::CelebrationPackageRequest

### LINKED ISSUE #831: Refactor Hired::Form into Developers::CelebrationPackageRequest
Currently, Developer Celebration Requests live in the `Hired::Form` namespace for Users, and `Admin::Developers::Forms` for Admin.

It would be more ideal if both `Hired::Forms` and `Admin::Developers::Forms` were normalized into `Developers::CelebrationPackageRequest`for both Developer and Admin.

Following the refactor in #818, A PR accomplishing this would:

* Move `Hired::Forms` controller/routes into `Developers::CelebrationPackageRequest` for Developers
* Move `Developers::Forms` controller/routes into `Admin::Developers::CelebrationPackageRequest` for Admin
* Change the table name `hired_forms` to `developers_celebration_package_requests`
* Move the admin mailer and previews from `new_hired_form` to `developers_celebration_package_request`

MikeRayUX:
This PR refactors existing `Hired::Forms` into `Developers::CelebrationPackageRequest` and `Admin::Developers::Forms` into `Admin::Developers::CelebrationPackageRequest`

This PR closes #831 by doing the following:

- Moves `Hired::Forms` controller/routes into `Developers::CelebrationPackageRequest` for Developers
- Moves `Admin::Developers::Forms` controller/routes into `Admin::Developers::CelebrationPackageRequest` for Admin
- Changes the model & table name `Hired::Form / hired_forms` to `Developers::CelebrationPackageRequest / developers_celebration_package_requests`
- Moves the admin mailer and previews from `new_hired_form` to `developers_celebration_package_request`
- Renames `Admin::NewHiredFormNotification` to `Admin::Developers::NewCelebrationPackageRequestNotification`
- Adds a Backfill for renaming existing `Admin::NewHiredFormNotification` to `Admin::Developers::NewCelebrationPackageRequestNotification` in `lib/tasks/backfills.rake`

PR Checklist:

Post Deployment Checklist:

~Run backfills rake task for notification names `rake backfills:rename_notifications`~
- [ ] Run pending migration `rails db:migrate`


MikeRayUX:
@joemasilotti ready for review!

app/controllers/developers/celebration_package_requests_controller.rb
```diff
@@ -1,14 +1,15 @@
...
-  class FormsController < ApplicationController
+module Developers
+  class CelebrationPackageRequestsController < ApplicationController
+    before_action :test
     before_action :authenticate_user!
     before_action :require_developer!
 
     def new
-      @form = Hired::Form.new(developer: current_user.developer)
+      @form = Developers::CelebrationPackageRequest.new(developer: current_user.developer)
```
joemasilotti, line 8: I think we can remove the `Developers::` namespace since we are already in the module.

app/controllers/developers/celebration_package_requests_controller.rb
```diff
@@ -18,14 +19,17 @@ def create
...
+    def test
+    end
+
     def require_developer!
       if current_user.developer.blank?
         redirect_to new_developer_path, notice: I18n.t("errors.developer_blank")
       end
     end
 
     def form_params
```
joemasilotti, line 31: Let's rename this method to match the new domain model.

app/controllers/developers/celebration_package_requests_controller.rb
```diff
@@ -18,14 +19,17 @@ def create
 
     private
 
+    def test
+    end
+
```
joemasilotti, line 24: Please delete this code.

app/models/developers/celebration_package_request.rb
```diff
@@ -1,8 +1,8 @@
-module Hired
-  class Form < ApplicationRecord
-    include Hired::Notifications
+module Developers
+  class CelebrationPackageRequest < ApplicationRecord
+    include Developers::CelebrationPackageRequests::Notifications
```
joemasilotti, line 3:
```suggestion
    include Notifications
```

app/models/developers/celebration_package_request.rb
```diff
@@ -1,8 +1,8 @@
...
 
-    self.table_name = "hired_forms"
+    self.table_name = "developers_celebration_package_requests"
```
joemasilotti, line 5: Can we move this to the `Developers` module? See `app/models/hiring_agreements.rb:3`.

app/notifications/admin/developers/new_celebration_package_request_notification.rb
```diff
@@ -0,0 +1,22 @@
+module Admin
+  module Developers
+    class NewCelebrationPackageRequestNotification < ApplicationNotification
+      deliver_by :database
+      deliver_by :email, mailer: "AdminMailer", method: :developers_celebration_package_request
+
+      param :form
```
joemasilotti, line 7: This should probably be renamed to match the new model.

test/models/concerns/developers/celebration_package_requests/notifications_test.rb
```diff
@@ -0,0 +1,21 @@
+require "test_helper"
+
+class Developers::CelebrationPackageRequests::NotificationsTest < ActiveSupport::TestCase
```
joemasilotti, line 3: Please split these modules on different lines.

lib/tasks/backfills.rake
```diff
@@ -1,9 +1,7 @@
 desc "These tasks are meant to be run once then removed"
 namespace :backfills do
-  task specialties: :environment do
-    specialties = YAML.load_file(Rails.root.join("db", "seeds", "specialties.yml"))
-    specialties.each do |specialty|
-      Specialty.find_or_create_by!(name: specialty)
-    end
+  task rename_notifications: :environment do
+    Notification.where(type: "Admin::NewHiredFormNotification")
+      .update_all(type: Admin::Developers::NewCelebrationPackageRequestNotification.to_s)
```
joemasilotti, line 5: Ohh, great call! Thanks for finding this.

Can you add a to-do to the PR post deployment so I don't forget to run it?

joemasilotti (changes requested):
Looking good! I made a few requests for code style and naming.

My only outstanding question is if methods shouldn't include the "s", for example `def developer_celebreation_package_request`. What do you think?

CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 ### March
 
+* March 23 - Refactor hired_form into developers_celebration_package_request #836 @MikeRayUX
```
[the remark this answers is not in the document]
MikeRayUX: Improved in 570e8d829d529d6d002df1d5e28b03683df0bad8

app/controllers/developers/celebration_package_requests_controller.rb
MikeRayUX: Removed in 657afa99c63b8ea99aab168a1677fda9cc2596d2

app/controllers/developers/celebration_package_requests_controller.rb
MikeRayUX: Removed in d857138a3e4028ed86a75a07aa35b18df5a878f9

app/models/developers/celebration_package_request.rb
MikeRayUX: Looks like in order to use bare `Notifications` include we'd need to drop the `s` from the `notifications.rb` which would break from naming convention of other `../../notifications.rb`

Should we rename `developers/celebration_package_requests/notifications.rb`?

Or keep `include Developers::CelebrationPackageRequests::Notifications` here?

app/models/developers/celebration_package_request.rb
MikeRayUX: Moved in 602f8c43649a7d7dbea5a19c7a1e9cc2691cc5ad

lib/tasks/backfills.rake
MikeRayUX: Sure, added! https://github.com/joemasilotti/railsdevs.com/pull/836#issue-1638591629

app/models/developers/celebration_package_request.rb
joemasilotti: Ah, I see. Disregard my suggestion!

MikeRayUX: Was able to drop the `Developers::` though! In 1ba19b2e416ac9013a9061e9ba4b76f167d1989f

app/notifications/admin/developers/new_celebration_package_request_notification.rb
MikeRayUX: Renamed  in 9cf0ce13394fccf9b7d041e82417ba41d3d6c9b8

MikeRayUX:
> My only outstanding question is if methods shouldn't include the "s", for example `def developer_celebreation_package_request`. What do you think?

I've thought about this too. 

I'm thinking preserving pluralization keeps it consistent with `businesses_`

https://github.com/joemasilotti/railsdevs.com/blob/2b6491baf0caef7a34e68853d168cae99926bc84/app/mailers/admin_mailer.rb#L53-L72

Totally up you though! Happy to change to your liking.

MikeRayUX:
@joemasilotti ready for review!

MikeRayUX:
@joemasilotti ready for review!

joemasilotti:
After running the migration and the backfill I'm getting the following error when viewing the mailer preview for the renamed notification.

http://localhost:3000/rails/mailers/admin_mailer/developers_celebration_package_request

```
web      | AdminMailer#developers_celebration_package_request: processed outbound mail in 5.5ms
web      | Completed 500 Internal Server Error in 14ms (ActiveRecord: 2.8ms | Allocations: 7905)
web      |
web      |
web      |
web      | NameError (uninitialized constant Hired
web      |
web      |       Object.const_get(camel_cased_word)
web      |             ^^^^^^^^^^):
web      |
web      | app/mailers/admin_mailer.rb:64:in `developers_celebration_package_request'
```

MikeRayUX:
>After running the migration and the backfill I'm getting the following error when viewing the mailer preview for the renamed notification.

Hmm, taking a look right now.

MikeRayUX:
Do you get that error when initially loading the mailer preview? Or when clicking 'View request'?

Screen Shot 2023-03-25 at 2 05 21 PM

MikeRayUX:
I have a sneaking suspicion this is from the change to the form param on the `Hired` notification class. 

I just reverted that change in 11b741784a4ba89216c655c74aae1b1e5f0076d1.

Can you try again?

joemasilotti:
Unfortunately the same error. But if I define `Hired::Form` somewhere then it seems to get a bit farther.

What I think is going on is that `notifications.params` has a serialized version of a `Hired::Form` record. So even after we rename it, Noticed is trying to deserialize it and can't find the class definition.

Check out this call to the first notification of that type in the database. Note the `#params_before_type_cast` which gives us the raw jsonb value.

```
[7] pry(main)> Notification.where(type: Admin::Developers::NewCelebrationPackageRequestNotification.to_s).first.params_before_type_cast
  Notification Load (2.5ms)  SELECT "notifications".* FROM "notifications" WHERE "notifications"."type" = $1 ORDER BY "notifications"."id" ASC LIMIT $2  [["type", "Admin::Developers::NewCelebrationPackageRequestNotification"], ["LIMIT", 1]]
=> "{\"form\": {\"_aj_globalid\": \"gid://railsdevs/Hired::Form/1\"}, \"_aj_symbol_keys\": [\"form\"]}"
```

What I think we need to do is rename the `Hired::Form` part (and maybe the `form`) strings in this column to match our new modeling.

Note, to reproduce this you need to:
1. `git checkout main`
2. `bin/rails db:reset`
3. Check out this branch/PR
4. `bin/rails db:migrate`
5. `bundle exec rake backfills:rename_notifications`

MikeRayUX:
Getting same error even before backfill:

```bash
[10] pry(main)> Notification.where(type: "Admin::NewHiredFormNotification").first.params
  Notification Load (0.3ms)  SELECT "notifications".* FROM "notifications" WHERE "notifications"."type" = $1 ORDER BY "notifications"."id" ASC LIMIT $2  [["type", "Admin::NewHiredFormNotification"], ["LIMIT", 1]]
NameError: uninitialized constant Hired

      Object.const_get(camel_cased_word)
```

So it looks like we're just updating the type name and it's still very much depends on the `Hired` class existing.

I'm wondering if we can't just clone params into new records in backfill?

MikeRayUX:
I was able to access existing notification params by restoring the `Hired::Form` model and using the new table name instead of `hired_forms`:

```bash
 pry(main)> Notification.where(type: "Admin::NewHiredFormNotification").first.params
  Notification Load (0.3ms)  SELECT "notifications".* FROM "notifications" WHERE "notifications"."type" = $1 ORDER BY "notifications"."id" ASC LIMIT $2  [["type", "Admin::NewHiredFormNotification"], ["LIMIT", 1]]
  Hired::Form Load (0.7ms)  SELECT "developers_celebration_package_requests".* FROM "developers_celebration_package_requests" WHERE "developers_celebration_package_requests"."id" = $1 LIMIT $2  [["id", 1], ["LIMIT", 1]]
=> {:form=>
  #<Hired::Form:0x000000010d6a88b8
   id: 1,
   developer_id: 27,
   address: "123 Main St\nNew York, NY 10001",
   company: "Rails for All",
   position: "Rails Developer",
   start_date: Sat, 25 Mar 2023,
   employment_type: "full_time_employment",
   feedback: "Keep up the great work with RailsDevs!",
   created_at: Sun, 26 Mar 2023 04:24:14.670408000 UTC +00:00,
   updated_at: Sun, 26 Mar 2023 04:24:14.670408000 UTC +00:00>}
[8] pry(main)>
```

```ruby
# app/models/hired/form.rb
module Hired
  class Form < ApplicationRecord
    self.table_name = "developers_celebration_package_requests"

    enum employment_type: {
      freelance_or_contract: 1,
      full_time_employment: 2
    }

    belongs_to :developer

    validates :address, :company, :position, :start_date, :employment_type, presence: true
  end
end
```
### Possible Solution 1:
We re-introduce the `Hired::Form` model class. A zombie model if you will.

* The existing `Admin::NewHiredFormNotification` notifications will still be accessible
* All new `NewCelebrationPackageRequestNotification` notifications will be created with the new type: `Admin::Developers::NewCelebrationPackageRequestNotification`

### Possible Solution 2:

We don't update the names of any existing `Admin::NewHiredFormNotification` notifications since a new one will never be created again.

joemasilotti:
I just updated the migration to rename the `params` column to work with the new model name.

Want to take a pass to see if you can re-apply the method name from `#form`?

MikeRayUX:
The migration solution works well if we don't touch the form param.

Tried renaming the param from `form` to `celebration_package_request` to match the model name and it succeeded at first, but we silently lost the developer association and thus loading the mailer breaks.

FWIW: the form is tied to the param for the notification only (even if the previous model was called form)

If we look at the admin_mailer `businesses_hiring_invoice_request` uses a form param as well

https://github.com/joemasilotti/railsdevs.com/blob/c95d5c6addc537af18c13c94e697f9af2e51fe36/app/mailers/admin_mailer.rb#L53-L61

Leaving it as a `form` matches the other admin notification.

Admittedly, this is uncharted territory for me. I'm not entirely sure how to preserve the `form.developer` association with raw sql in the migration.

MikeRayUX:
>Want to take a pass to see if you can re-apply the method name from #form?

After renaming the `form` param to `celebration_package_request` inside migration and the notification/mailer, we can access the notification params fine in the console, but loading the mailer breaks because we lost the developer association.

If we can also update the developer association in the migration, we should be able to successfully rename for form param.

Something else to consider:

The `form` is tied to the param in the notification only, not the model (even if the previous model was called form)

For example, the admin_mailer `businesses_hiring_invoice_request#L57` uses a `form` param as well

So we could just leave it as `form`

https://github.com/joemasilotti/railsdevs.com/blob/c95d5c6addc537af18c13c94e697f9af2e51fe36/app/mailers/admin_mailer.rb#L53-L61

joemasilotti:
Interesting! Is it possible to clear or delete the `params` in the notification then re-assign and save it? If so, that could clear up a fair amount of this backfill.

If not, aliasing a method to use the `form` under the hood sounds like the quickest win.

MikeRayUX:
I think we have something here!

I was able to preserve the developer association by renaming the param key from `:form` to `'celebration_package_request'`, and then symbolizing the key to `:celebration_package_request` with ActiveRecord

Now previous notifications are updated to match the new class and form param. 

The mailer preview loads fine now too!

https://github.com/joemasilotti/railsdevs.com/blob/cfdbdc13e7482bffc86a020c5f1f8d067321388d/db/migrate/20230323223735_rename_hired_forms_to_developers_celebration_package_requests.rb#L20-L47

Thinking this is now ready for review!

MikeRayUX:
The down migration fails because the `Hired::Form` model no longer exists with this refactor. 

If we add `Hired::Form` back it will work:

```ruby
# app/models/hired/form.rb
module Hired
  class Form < ApplicationRecord
    self.table_name = "hired_forms"
  end
end
```

```bash
❯ rails db:migrate:down VERSION=20230323223735
== 20230323223735 RenameHiredFormsToDevelopersCelebrationPackageRequests: reverting
-- rename_table(:developers_celebration_package_requests, :hired_forms)
   -> 0.0106s
-- execute("UPDATE notifications SET params = jsonb_set(params, '{celebration_package_request,_aj_globalid}', to_jsonb(REPLACE(params->'celebration_package_request'->>'_aj_globalid', 'Developers::CelebrationPackageRequest', 'Hired::Form'))) WHERE params->'celebration_package_request'->>'_aj_globalid' LIKE '%Developers::CelebrationPackageRequest%'")
   -> 0.0012s
-- execute("UPDATE notifications SET params = jsonb_set( params, '{form}', params->'celebration_package_request') - 'celebration_package_request' WHERE params->'celebration_package_request'->>'_aj_globalid' LIKE '%Hired::Form%';")
   -> 0.0011s
== 20230323223735 RenameHiredFormsToDevelopersCelebrationPackageRequests: reverted (0.0870s)
```

Should we add back `app/models/hired/form.rb`?

joemasilotti:
Nice! Let's define the class _inside_ the migration.

https://github.com/joemasilotti/railsdevs.com/pull/829/files#diff-aa7c41fd8f4121e629b0b59cf42c07772fd25756a2c1d1539bb39aff15eead80

joemasilotti:
Weird! I couldn't get it working either.

I renamed the migration to be later than the most recently ran one. Once CI passes this should be good to merge!

joemasilotti (approved):

PR #844, opened 2023-03-27
Add UrlAttribute concern

### LINKED ISSUE #842: Fix duplicated protocol prefix in input fields
Currently, there are some form fields in the UI using an `https://` input pefix. Namely:

- Business
  - `website`
- Developer
  - `website`
  - `mastodon`
  - `scheduling_link`

However, these attributes don't have any validation, and the protocol prefix of the pasted link (`http(s)://`) is somewhat duplicated in the form. It should be clear from this screenshot:

CleanShot 2023-03-25 at 21 13 22

- [ ] Take into account and refactor the existing solution for other online presence links (Github, Twitter etc.)
- [ ] Prevent showing duplicated `http(s)://` protocol prefix in the UI
- [ ] Add validation for these URL attributes

jkostolansky:
This PR introduces `UrlAttribute` concern to remove `http(s)://`protocol and custom prefix from URL attributes. It also refactors the existing logic for social profile links (github, ...).

It doesn't add any URL validation because I think it's not necessary at this point. The `http(s)://` protocol is not required in the submitted data, so even `asdf` should be valid, because `https://asdf` is a valid URL. 

Resolves #842


joemasilotti (changes requested):
This looks great and works exactly how I'd expect it to!

But I think we need a data migration/backfill to update all the existing records. I _think_ this can be accomplished by saving every developer record. But we also need to make sure not to update `updated_at` because that is used for the "Recently active" badge.

jkostolansky:
I've added a migration, what do you think? The URL logic is duplicated in the migration, because migrations shouldn't depend on the current application code (that will change in the future).

joemasilotti:
Sorry, I didn't mean we needed an actual Rails migration. You can throw a couple lines of a script in `lib/tasks/backfills.rake` that I can run as a one-off on Heroku.

Does that make sense?

jkostolansky:
🤦‍♂️ Need to think before typing. Sure, makes sense. Changed.

joemasilotti (approved):
Excellent - thank you!

Issue #839, opened 2023-03-24
Notify Admin by tracking user's LinkedIn profile on its any experience changes.


sarvaiyanidhi:
Need to notify admin on weekly basis on any changes in user's LinkedIn profile related to its experience.

Linkedin profile will be tracked using external service https://nubela.co/proxycurl/

A PR addressing this issue should

- [ ] New table developer_linkedin_profile_details storing all information related to developer's linkedin profile like current_company_name, start_date, title, location. Note, additional fields apart from company name are stored just for information now and can be used later.
- [ ] Script running weekly fetching LinkedIn details and storing in database
- [ ] Script notifying admin on any experience changes in developer's profile every week.

joemasilotti:
Thanks for kicking this off @sarvaiyanidhi!

My only addition is that we will be doing something very similar for GitHub soon. See #779.

And if we could model the new table to handle both somehow that would be great.

sarvaiyanidhi:
Actually, I did see that and had same thought of having single table storing all external information of developer's profile but as it is still in progress so can we come back and refactor to store information later from single table.

Another option, I can create generic table for now to create linkedin information and later on we can migrate Github changes to this table once it is done.

joemasilotti:
I'm happy to go with whatever you think is best. Just wanted to call out the GitHub integration in case it changed how you would structure stuff. :)

sonyco-4u:
I am interested you reach out to me on Whatsapp +2347060787395

joemasilotti:
Closed via #845.

PR #845, opened 2023-03-28
Functionality to fetch linkedin profile details using external API

sarvaiyanidhi:
This PR addresses implementation of sending notification to Admin in case of any changes in developer's LinkedIn profile.

Solves Issue - #839 

- Have added new table developer_external_profiles which will store details of any developer external profiles like LinkedIn, Github etc
- We need to schedule below two rake tasks in Heroku scheduler to fetch LinkedIn profiles and send updated profiles list to Admin.
  - rails linkedin:fetch_profiles - This task fetches LinkedIn current experience details of all active developers who are in conversation with business currently. We are using ProxyCurl API to fetch LinkedIn details.
  - rails linkedin:weekly_digest - This task will query linkedin profiles updated in last 7 days from developer_external_profiles table and send mail to Admin with developer name, LinkedIn URL and current company details.

We also need to configure proxycurl_api_key in production credentials file.

## Pull request checklist

### Checklist for production deployment

- [ ] Configure proxycurl_api_key in production credentials file
- [ ] Configure `bundle exec rake linkedin:fetch_profiles` - weekly on every Sunday at 7:00 AM UTC task in Heroku scheduler
- [ ] Configure `bundle exec rake linkedin:weekly_digest` - weekly on every Monday at 7:00 AM UTC task in Heroku scheduler
- [ ] Run task `bundle exec rake backfills:url_attributes` to normalize LinkedIn username in database


sarvaiyanidhi:
Hi @joemasilotti Have completed integration to send notification to Admin for updated LinkedIn profiles. Please share your feedback on the implementation for any changes or improvement in the code.

I am still working on test cases for LinkedIn profile API.

app/mailers/admin_mailer.rb
```diff
@@ -76,4 +78,10 @@ def affiliates_registration
 
     mail(to: @recipient.email, subject: @notification.title)
   end
+
+  def linkedin_weekly_profiles
+    @developer_external_profiles = params[:developer_external_profiles_records]
+    subject = "#{pluralize(@developer_external_profiles.count, "Developer")} updated Linkedin profile in past 7 days"
```
joemasilotti, line 84:
```suggestion
    subject = "#{pluralize(@developer_external_profiles.count, "Developer")} updated their LinkedIn profile in the last 7 days"
```

app/models/developer.rb
```diff
@@ -29,6 +29,7 @@ class Developer < ApplicationRecord
   has_one :role_level, dependent: :destroy, autosave: true
   has_one :role_type, dependent: :destroy, autosave: true
   has_one_attached :cover_image
+  has_many :developer_external_profiles, dependent: :destroy
```
joemasilotti, line 32:
```suggestion
  has_many :external_profiles, dependent: :destroy, class_name: "Developers::ExternalProfile"
```

I think that's the correct syntax.

app/models/developer_external_profile.rb
```diff
@@ -0,0 +1,15 @@
+class DeveloperExternalProfile < ApplicationRecord
```
joemasilotti, line 1:
```suggestion
module Developers
  class ExternalProfile < ApplicationRecord
```

app/models/developer_external_profile.rb
```diff
@@ -0,0 +1,15 @@
...
+  enum :sites, %i[linkedin github]
```
joemasilotti, line 2: Let's remove `github` since we aren't using it yet. But thanks for taking into account that we will in the future!

app/models/developer_external_profiles/linkedin.rb
```diff
@@ -0,0 +1,58 @@
...
+
+module DeveloperExternalProfiles
+  class Linkedin
+    def initialize
+      @api_key = api_key
+      @endpoint = endpoint
+    end
+
+    def get_profile(url)
+      header_hash = {"Authorization" => "Bearer " + @api_key}
```
joemasilotti, line 13:
```suggestion
      header_hash = {"Authorization" => "Bearer #{@api_key}"}
```

app/models/developer_external_profiles/linkedin_profile_fetcher.rb
```diff
@@ -0,0 +1,31 @@
+module DeveloperExternalProfiles
+  class LinkedinProfileFetcher
+    def conversation_developer_profiles
+      developer_external_profiles_list = []
+
+      Conversation.includes(:developer).where.not(developers: {linkedin: nil}).map(&:developer).uniq.each do |developer|
```
joemasilotti, line 6: We should be able to query for developers directly instead of mapping and uniquing.

`Developer.joins(:conversation).where.not(linked_in: [nil, ""]).distinct`

app/views/admin_mailer/linkedin_weekly_profiles.html.erb
```diff
@@ -0,0 +1,26 @@
+<p>Here are the developer with linkedin profiles updated in past 7 days:</p>
```
joemasilotti, line 1: Can you add a mailer preview? I'm having a hard time wrapping my head around what this will look like.

db/migrate/20230327024908_create_developer_external_profiles.rb
```diff
@@ -0,0 +1,13 @@
+class CreateDeveloperExternalProfiles < ActiveRecord::Migration[7.0]
+  def change
+    create_table :developer_external_profiles do |t|
+      t.references :developer, null: false, foreign_key: true
+      t.string :site, null: false
+      t.jsonb :data
```
joemasilotti, line 6:
```suggestion
      t.jsonb :data, null: false, default: {}
```

That way we always can access the hash directly, instead of checking if it exists first.

db/schema.rb
```diff
@@ -325,10 +336,10 @@
     t.string "owner_type"
     t.bigint "owner_id"
     t.string "processor", null: false
-    t.string "processor_id"
+    t.string "processor_id", precision: nil
     t.boolean "default"
     t.jsonb "data"
-    t.datetime "deleted_at", precision: nil
+    t.datetime "deleted_at"
```
joemasilotti, line 342: Please revert these changes.

lib/tasks/linkedin.rake
```diff
@@ -0,0 +1,11 @@
+namespace :linkedin do
+  desc "Fetch and parse LinkedIn profiles"
+  task fetch_profiles: :environment do
+    DeveloperExternalProfiles::LinkedinProfileFetcher.new.conversation_developer_profiles
+  end
+
+  desc "Email linkedin updated in past 7 days"
```
joemasilotti, line 7:
```suggestion
  desc "Email LinkedIn updates from past 7 days"
```

test/fixtures/developer_external_profiles.yml
```diff
@@ -0,0 +1,6 @@
+# Read about fixtures at https://api.rubyonrails.org/classes/ActiveRecord/FixtureSet.html
```
joemasilotti, line 1: Is this file being used?

joemasilotti (changes requested):
This is looking good so far! I left a bunch of nitpicky comments.

Outside of those, I'm wondering if we want to use Faraday instead of net/http. The gem is already in the repo and it might be easier to work with the JSON responses and making HTTP requests.

I'm also a little confused as to the flow for how an error is bubbled up and then saved to the database. Can you explain the rationale behind that and how you expect it to work? Especially with how often these errors are expected - for example, if a user doesn't have work experience will that throw an error? Or is it only if the API changes or something goes wrong?

sarvaiyanidhi:
Thank you for the feedback. I will push the mentioned fixes soon.
 
> Outside of those, I'm wondering if we want to use Faraday instead of net/http. The gem is already in the repo and it might be easier to work with the JSON responses and making HTTP requests.

Yes, I will make the changes in code to use Faraday gem.

> I'm also a little confused as to the flow for how an error is bubbled up and then saved to the database. Can you explain the rationale behind that and how you expect it to work? Especially with how often these errors are expected - for example, if a user doesn't have work experience will that throw an error? Or is it only if the API changes or something goes wrong?

There can be many errors like provided linkedin username by developer is not valid so it will store in database error related to it which states that linkedin url is not valid to fetch details with error code. There can also be error if we run out of credit from ProxyCurl URL so data won't be fetched and we will get error captured for it. We also capture incase if we are not able to parse JSON to get experience details. Basically, I wanted to captured any error while fetching details from API and may be we can use in future if we want to avoid hitting those profiles again as still uses credit even it is not valid to check it.

sarvaiyanidhi:
Joe, I just had one more question while working on code

Right now I am considering only those developers who are in conversations with any profile but what if any business has not interacted on app with developer but contacted directly developer using its Github or linkeidn profile so in that case we might miss such developers details but again fetching profiles for all developers will be expensive so wanted to know your thoughts whether we should just keep flow limited to conversation developers only for now?

joemasilotti:
Gotcha, that makes sense regarding the errors. Thanks for diving in! Is there an easy way to report on the errors, either through Honeybadger, a dashboard, or a query I can quickly run in production?

And that's a great point regarding developers who haven't been contacted. Let's remove that scope for now and query everyone who has a LinkedIn profile added - we can always add it back in later.

sarvaiyanidhi:
Hi @joemasilotti Have addressed all comments and pushed revised code.

Query is also updated to fetch all developers with LinkedIn profile.

Related to showing error for failed API, if we consider Honeybadger it might add up lots of notification for exceptions in case if out of 1000 available profiles we encounter issues with 50 profiles while fetching details. 

What do you think of option of adding one more menu in Admin displaying LinkedIn profiles details with actual data or error if failed API? Another option is to display error information as well in same weekly main sent to Admin.

joemasilotti:
For the first release of this I'd like to keep it as simple as possible.

So let's go with Honeybadger notifications, only, for now. I just checked - I get 1000 notifications per month and used about 400 in March. So unless we are expecting 600+ errors the first time we run this we should be OK.

Making sure to report the right context - what happened, who the developer is, any LinkedIn API response data - and grouping the errors all the same - will make it easier to debug in the future.

sarvaiyanidhi:
Hi @joemasilotti Have done exception handling using HoneyBadger notification. Please review and let me know if any more changes are required in the implementation. Thanks

app/models/developer_external_profiles/linkedin.rb
```diff
@@ -0,0 +1,76 @@
+require "faraday"
+require "json"
+
+module DeveloperExternalProfiles
```
joemasilotti, line 4: I think we want this under `Developers::ExternalProfiles::LinkedIn`.

app/models/developer_external_profiles/linkedin_profile_fetcher.rb
```diff
@@ -0,0 +1,48 @@
+module DeveloperExternalProfiles
+  class LinkedinProfileFetcher
+    def developer_profiles
+      developer_external_profiles_list = []
```
joemasilotti, line 4: `profiles`, maybe? There's a lot of text in this method making it a bit hard to read.

app/models/developer_external_profiles/linkedin_profile_fetcher.rb
```diff
@@ -0,0 +1,48 @@
...
+    def upsert_external_profiles(profiles)
+      unless profiles.empty?
+        Developers::ExternalProfile.upsert_all(profiles, unique_by: [:developer_id, :site])
+      end
+    end
+
+    private
+
+    def developers_with_linked_in_profiles
+      Developer.where.not(linkedin: [nil, ""]).distinct
```
joemasilotti, line 35: We can drop the distinct now that we aren't joining.

app/models/developer_external_profiles/linkedin_profile_fetcher.rb
joemasilotti: And maybe rename `record` a few lines down.

app/models/developer_external_profiles/linkedin_profile_fetcher.rb
```diff
@@ -0,0 +1,48 @@
...
+      upsert_external_profiles(developer_external_profiles_list)
+    end
+
+    def external_profile(developer, response)
+      if response[:error]
+        {developer_id: developer.id, site: "linkedin", data: {}, error: response[:error]}
+      else
+        external_profile = Developers::ExternalProfile.linkedin_developer(developer)
+        external_profile.data = response[:data] unless external_profile.blank?
+        if external_profile.blank? || external_profile.data_changed?
```
joemasilotti, line 20: I'm not sure I understand this logic. Why check for blank _or_ changed? Can you explain what's going on here?

app/models/developer_external_profiles/linkedin_profile_fetcher.rb
```diff
@@ -0,0 +1,48 @@
...
+    end
+
+    def fetch_linkedin_profile(linkedin_id)
+      linkedin_url = "https://linkedin.com/in/#{linkedin_id}/"
+      get_profile(linkedin_url)
+    end
+
+    def get_profile(linkedin_url)
+      api = DeveloperExternalProfiles::Linkedin.new
+      api.get_profile(linkedin_url)
```
joemasilotti, line 45:
```suggestion
      DeveloperExternalProfiles::Linkedin.new.get_profile(linkedin_url)
```

app/models/developer_external_profiles/linkedin.rb
```diff
@@ -0,0 +1,76 @@
...
+          linkedin_api_response: response&.body
+        })
+        {error: "Exception occurred: #{e.message}"}
+      end
+    end
+
+    private
+
+    def parse_json_response(response_body)
+      response_hash = JSON.parse(response_body)
```
joemasilotti, line 53: I think Faraday can do this for us, see: https://lostisland.github.io/faraday/usage/#middleware

test/integration/developer_external_profiles/linkedin_profile_fetcher_test.rb
```diff
@@ -0,0 +1,52 @@
...
+      .with(headers: {"Authorization" => "Bearer test_api_key"})
+      .to_return(response)
+
+    response_hash = JSON.parse(response[:body])
+    @developer_external_profiles_list << @fetcher.external_profile(developer, {data: response_hash["experiences"][0]})
+    @fetcher.upsert_external_profiles(@developer_external_profiles_list)
+
+    external_profile = developer.external_profiles.find_by(site: "linkedin")
+    assert_not_nil external_profile
+    assert_equal response_hash["experiences"][0]["company"], external_profile.data["company"]
```
joemasilotti, line 30: I'm fine with reporting the actual value, "Example Company", here to make it more obvious what's going on.

joemasilotti (changes requested):
This is looking good! But I'm not sure the best way to test it.

Can you provide a quick process overview of how you are making sure this works when hitting the real API?

Also, let's remove the Honeybadger notifications. I'm fine (for now) digging into the database to see what errored and why. It looks like you are already doing a great job exposing the exception on the external profile model.

Finally, let's add some database seeds to make sure we can view the mailer preview with actual data.

sarvaiyanidhi:
Thanks for the feedback. Will push fixes shortly.

I will also post the process and steps to test the feature locally.

app/models/developer_external_profiles/linkedin_profile_fetcher.rb
sarvaiyanidhi: external_profile.blank? - This check is done to create new entry in external_profile table for the developer if there is no linkedin data associated with that developer already present in table
external_profile.data_changed? - This check is done to update existing entry of developer in external_profile table only if developer's latest company experience received from API is changed from the one already stored in database for it.

Basically, I am doing this check so that external_profile record is not updated if there are no changes as weekly_profile mail sends list of only those developers who are newly added or its experience has changed in last 7 days.

sarvaiyanidhi:
Hi @joemasilotti Have pushed all fixes as per feedback.
I have added seed data for external_profile but I am creating two new entries in developer for valid and invalid LinkedIn developer profile as I want to add both valid as well as error data in table.

Have also removed Honeybadger notification for LinkedIn from the code.

I have tested code locally using test ProxyCurl API key which we had created for fetching LinkedIn profiles.

Overview of the whole flow
- We are basically executing two tasks in scheduler.
- First task is to fetch data for all developers having LinkedIn profile. This is done by calling task rails linkedin:fetch_profiles which will internally call  Developers::ExternalProfiles::LinkedinProfileFetcher.new.developer_profiles to loop through each developer and call ProxyCurl API to fetch data for the profile and store in database
- Second task is to send developers whose LinkedIn profile is updated in last 7 days to Admin. This is done by calling task rails linkedin:weekly_digest .

In order for you to test locally, you can set test ProxyCurl API key in development credentials file. After that execute above two mentioned tasks to fetch data and email to Admin.

Let me know if I am not clear with details and need any additional information.

app/models/developers/external_profiles/linkedin_profile_fetcher.rb
```diff
@@ -0,0 +1,47 @@
...
+    end
+
+    private
+
+    def developers_with_linked_in_profiles
+      Developer.where.not(linkedin: [nil, ""])
+    end
+
+    def fetch_linkedin_profile(linkedin_id)
+      linkedin_url = "https://linkedin.com/in/#{linkedin_id}/"
```
joemasilotti, line 39: It looks like some developers have trailing and/or leading slashes in their LinkedIn field. For example, `/foo-bar` or `/baz-qux/`.

Should we handle that here or normalize data when updating your profile + a task to clean existing data up?

app/models/developers/external_profiles/linkedin_profile_fetcher.rb
```diff
@@ -0,0 +1,47 @@
...
+      get_profile(linkedin_url)
+    end
+
+    def get_profile(linkedin_url)
+      Developers::ExternalProfiles::Linkedin.new.get_profile(linkedin_url)
```
joemasilotti, line 44:
```suggestion
      Developers::ExternalProfiles::Linkedin.new.get_profile(linkedin_url)
```

app/models/developers/external_profiles/linkedin.rb
```diff
@@ -0,0 +1,54 @@
...
+        faraday.adapter Faraday.default_adapter
+        faraday.response :json
+      end
+
+      begin
+        response = conn.get("", params)
+        if response.status.to_i == 200
+          parse_json_response(response.body)
+        else
+          {error: "API Error: #{response.status} - #{response.body}"}
```
joemasilotti, line 26: I'd like to return a concrete type here instead of a hash.

```ruby
module Developers::ExternalProfiles
  class Linkedin
    class Response
      attr_reader :error, :data

      def initialize(data: nil, error: nil)
        @data = data
        @error = error
      end

      def error?
        error.present?
      end
    end
  end
end
```

Then we can make this change, which I think cleans things up a bit:

```diff
diff --git a/app/models/developers/external_profiles/linkedin_profile_fetcher.rb b/app/models/developers/external_profiles/linkedin_profile_fetcher.rb
index 29682b62..9b527370 100644
--- a/app/models/developers/external_profiles/linkedin_profile_fetcher.rb
+++ b/app/models/developers/external_profiles/linkedin_profile_fetcher.rb
@@ -12,13 +12,13 @@ module Developers::ExternalProfiles
     end

     def external_profile(developer, response)
-      if response[:error]
-        {developer_id: developer.id, site: "linkedin", data: {}, error: response[:error]}
+      if response.error?
+        {developer_id: developer.id, site: "linkedin", data: {}, error: response.error}
       else
         external_profile = Developers::ExternalProfile.linkedin_developer(developer)
-        external_profile.data = response[:data] unless external_profile.blank?
+        external_profile.data = response.data unless external_profile.blank?
         if external_profile.blank? || external_profile.data_changed?
-          {developer_id: developer.id, site: "linkedin", data: response[:data], error: nil}
+          {developer_id: developer.id, site: "linkedin", data: response.data, error: nil}
         end
       end
     end
```

app/models/developers/external_profiles/linkedin_profile_fetcher.rb
```diff
@@ -0,0 +1,47 @@
...
+        external_profile.data = response[:data] unless external_profile.blank?
+        if external_profile.blank? || external_profile.data_changed?
+          {developer_id: developer.id, site: "linkedin", data: response[:data], error: nil}
+        end
+      end
+    end
+
+    def upsert_external_profiles(profiles)
+      unless profiles.empty?
+        Developers::ExternalProfile.upsert_all(profiles, unique_by: [:developer_id, :site])
```
joemasilotti, line 28: `.upsert_all`, nice!

app/models/developers/external_profiles/linkedin.rb
```diff
@@ -0,0 +1,54 @@
...
+      params = {
+        "url" => url,
+        "fallback_to_cache" => "on-error",
+        "use_cache" => "if-present"
+      }
+
+      conn = Faraday.new(@endpoint) do |faraday|
+        faraday.headers["Authorization"] = "Bearer #{@api_key}"
+        faraday.adapter Faraday.default_adapter
+        faraday.response :json
```
joemasilotti, line 18:
```suggestion
        faraday.adapter(Faraday.default_adapter)
        faraday.response(:json)
```

app/models/developers/external_profiles/linkedin.rb
```diff
@@ -0,0 +1,54 @@
...
+      end
+
+      begin
+        response = conn.get("", params)
+        if response.status.to_i == 200
```
joemasilotti, line 23: I think there's a helper for checking of a response is OK in Faraday.

app/models/developers/external_profiles/linkedin.rb
```diff
@@ -0,0 +1,54 @@
...
+    def parse_json_response(response_body)
+      begin
+        # Attempt to access the first company name in the experiences array
+        {data: response_body["experiences"][0]}
+      rescue NoMethodError => e
+        {error: "JSON Parsing Error: #{e.message}"}
+      end
+    rescue JSON::ParserError => e
+      {error: "JSON Parsing Error: #{e.message}"}
+    end
```
joemasilotti, line 44: I think this has the same result, right?

```suggestion
    def parse_json_response(response_body)
      # Attempt to access the first company name in the experiences array
      {data: response_body["experiences"][0]}
    rescue JSON::ParserError, NoMethodError => e
      {error: "JSON Parsing Error: #{e.message}"}
    end
```

app/models/email_digests/linkedin_profiles.rb
```diff
@@ -0,0 +1,20 @@
+class EmailDigests::LinkedinProfiles
+  def send_weekly_digest
+    # return unless Time.current.monday?
```
joemasilotti, line 3: What's the rationale behind this comment?

app/models/email_digests/linkedin_profiles.rb
```diff
@@ -0,0 +1,20 @@
...
+    developer_external_profiles_records = []
```
joemasilotti, line 4:
```suggestion
    linkedin_profiles = []
```

app/models/developers/external_profiles/linkedin_profile_fetcher.rb
```diff
@@ -0,0 +1,47 @@
+module Developers::ExternalProfiles
+  class LinkedinProfileFetcher
+    def developer_profiles
+      profiles = []
+      developers_with_linked_in_profiles.each do |developer|
```
joemasilotti, line 5:
```suggestion
      developers_with_linkedin_profiles.each do |developer|
```

To keep our LinkedIn naming consistent.

app/mailers/admin_mailer.rb
```diff
@@ -78,4 +80,10 @@ def affiliates_registration
 
     mail(to: recipient.email, subject: @notification.title)
   end
+
+  def linkedin_weekly_profiles
+    @developer_external_profiles = params[:developer_external_profiles_records]
+    subject = "#{pluralize(@developer_external_profiles&.count, "developer")} updated their LinkedIn profile in the last 7 days"
```
joemasilotti, line 86: 1. Let's rename this to `linkedin_profiles`.
2. We can use [`String#pluralize`](https://apidock.com/rails/String/pluralize) and not delegate to the Rails helper.
3. Why are we optional chaining here? Shouldn't this always be present?

db/seeds/external_profiles.rb
```diff
@@ -0,0 +1,35 @@
...
+    company_linkedin_profile_url: "https://www.linkedin.com/company/freedomfund",
+    description: Faker::Lorem.paragraph(sentence_count: 5),
+    ends_at: nil,
+    location: nil,
+    logo_url: "",
+    starts_at: {
+      day: 1,
+      month: 8,
+      year: 2021
+    },
```
joemasilotti, line 25: Let's dynamically set this date so it shows up in the mailer preview.

app/views/admin_mailer/linkedin_weekly_profiles.html.erb
joemasilotti, line 1:
```suggestion
<p>Developers with LinkedIn profile updates in the last 7 days:</p>
```

And same copy update for the plain-text mailer.

test/mailers/previews/admin_mailer_preview.rb
```diff
@@ -40,4 +40,10 @@ def affiliates_registration
     notification = Notification.where(type: Admin::Affiliates::RegistrationNotification.to_s).first
     AdminMailer.with(record: notification, recipient: User.first).affiliates_registration
   end
+
+  def linkedin_weekly_profiles
+    prev_7_days = 7.days.ago.beginning_of_day..1.day.ago.end_of_day
+    developer_external_profiles = Developers::ExternalProfile.where(updated_at: prev_7_days).includes(:developer).limit(3)
+    AdminMailer.with(developer_external_profiles:).linkedin_weekly_profiles
```
joemasilotti, line 47: Let's match these to the new variable name in the mailer, including the param.

app/views/admin_mailer/linkedin_weekly_profiles.html.erb
```diff
@@ -0,0 +1,26 @@
...
+      <th scope="col" style="padding-top: 0.75rem;padding-bottom: 0.75rem; padding-left: 1.5rem;padding-right: 1.5rem; ">Developer Name</th>
+      <th scope="col" style="padding-top: 0.75rem;padding-bottom: 0.75rem; padding-left: 1.5rem;padding-right: 1.5rem; ">Linkeidn URL</th>
+      <th scope="col" style="padding-top: 0.75rem;padding-bottom: 0.75rem; padding-left: 1.5rem;padding-right: 1.5rem; ">Current Company</th>
+    </tr>
+  </thead>
+  <tbody>
+  <% @developer_external_profiles&.each do |developer_external_profile| %>
+    <tr style="background-color: #ffffff; border-bottom-width: 1px; ">
+      <th scope="row" style="padding-top: 1rem;padding-bottom: 1rem; padding-left: 1.5rem;padding-right: 1.5rem; color: #111827; font-weight: 500; white-space: nowrap; ">
+        <%= link_to developer_external_profile[:name], developer_url(developer_external_profile[:id]) %>
```
joemasilotti, line 14: When I view the preview I see their "name" as the URL, too.

I also am thinking if it might make more sense to pass this as an object (like the comment above) instead of a hash. What do you think?

config/honeybadger.yml
```diff
@@ -1 +1 @@
-api_key: <%= Rails.application.credentials.dig(:honeybadger, :api_key) %>
+api_key: "<%= Rails.application.credentials.dig(:honeybadger, :api_key) %>"
```
joemasilotti, line 1: What's the point of this change?

joemasilotti (changes requested):
This is looking _great_ @sarvaiyanidhi! Awesome work on this.

I left a bunch of nitpicky questions/comments. But the overall structure and code organization is perfect.

The big feedback is on whether to promote some of the hashes to first class objects - I provided one example. Let me know if you have any questions.

joemasilotti:
Oh, one last thing. Can you also please document the rake tasks that need to be added to production in this PR and the Production Settings markdown doc? Then I can add them to Heroku before deploying. Thanks!

sarvaiyanidhi:
> Oh, one last thing. Can you also please document the rake tasks that need to be added to production in this PR and the Production Settings markdown doc? Then I can add them to Heroku before deploying. Thanks!

Sure, will do that and also push fixes for the feedback.

config/honeybadger.yml
sarvaiyanidhi: I think it happened when I was adding additional config for grouping LinkedIn API errors as per earlier comment.. Will revert this change.

app/models/developers/external_profiles/linkedin_profile_fetcher.rb
sarvaiyanidhi: I think it will be better to handle while updating profile and running rake task to clean existing data as it not only effects this code but other places wherein we display LinkedIn profile details https://github.com/joemasilotti/railsdevs.com/blob/main/app/components/social_link_component.rb#L28

sarvaiyanidhi: Have added code to normalized LinkedIn username in existing file https://github.com/joemasilotti/railsdevs.com/blob/main/app/models/concerns/developers/has_online_profiles.rb

We already have rake task added for normalizing online public profiles which we can run again `rake backfills:url_attributes`

app/models/email_digests/linkedin_profiles.rb
sarvaiyanidhi: Removed it as it is not required

app/models/developer_external_profiles/linkedin_profile_fetcher.rb
joemasilotti: Ah! That makes sense - thanks for explaining.

app/models/developers/external_profiles/linkedin_profile_fetcher.rb
joemasilotti: Good call - thanks!

app/views/admin_mailer/linkedin_weekly_profiles.html.erb
sarvaiyanidhi: Yeah I have added name as URL pointing to page of  developer profile in application so that in case you need to check any more details or contact developer through app it will be easy to do.

Have removed Hash and passing directly object to fetch details.

app/mailers/admin_mailer.rb
sarvaiyanidhi: yes there is no need to optional chaining here. Have updated code and made other changes as well.

sarvaiyanidhi:
Hi @joemasilotti I have addressed all issues in feedback and pushed the code.

Have also updated Production Settings markdown with background jobs to be configured in Heroku as well as updated in this PR Checklist for production deployment.

Please review and let me know if it is good to merge or need anymore changes. Thanks

app/models/email_digests/linkedin_profiles.rb
```diff
@@ -0,0 +1,15 @@
+class EmailDigests::LinkedinProfiles
+  def send_digest
+    last_month = Date.current.last_month
+    last_month = last_month.beginning_of_month..last_month.end_of_month
+    linkedin_profiles = Developers::ExternalProfile
+      .where(updated_at: last_month)
```
joemasilotti, line 6: I'm confused by this line. Don't we want to include anyone who _changed their job_ in the last month? `updated_at` will be when we updated the record.

Or are you thinking that we only update once per month so _any_ update will correspond to a job change? Let me know what you're thinking here!

joemasilotti (changes requested):
OK, I think we are _really_ close! I realized I don't need to get weekly updates so migrated everything to fetch/send monthly, instead. This should also save some money since we will be using ¼ credits for API access.

The last thing we need to resolve is the question in the email digest.

app/models/email_digests/linkedin_profiles.rb
sarvaiyanidhi: We are updating record only if there is any change in its LinkedIn profile data https://github.com/joemasilotti/railsdevs.com/blob/9aaf5da476c59619c00e72b1bfae96f8e4ae0487/app/models/developers/external_profiles/linkedin_profile_fetcher.rb#L26

That is the reason I am checking updated_at as it will be record will be updated only if developer's current company data is updated.

joemasilotti: Got it, that makes sense. So the first email will be _every_ developer with a LinkedIn profile and the next one will only send changes. Thanks!

joemasilotti (approved):

PR #846, opened 2023-03-28
Imagelink

allancodenet:
## Pull request checklist

### DIFF
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -12,7 +12,6 @@
 
 ActiveRecord::Schema[7.0].define(version: 2023_03_28_223735) do
   # These are extensions that must be enabled in order to support this database
-  enable_extension "pg_stat_statements"
   enable_extension "plpgsql"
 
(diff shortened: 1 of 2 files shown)


joemasilotti:
Thanks for the PR, @allancodenet! I'm a bit confused though - why does adding a link increase accessibility? There is already an alt tag for screen readers.

allancodenet:
I see that now, It sis not a very necessary feature but It still encourages the action  to go to your personal site

db/schema.rb
```diff
@@ -328,7 +327,7 @@
     t.string "processor_id"
     t.boolean "default"
     t.jsonb "data"
-    t.datetime "deleted_at", precision: nil
```
nunommc, line 331: surely the db/schema doesn't need to change

joemasilotti:
> I see that now, It is not a very necessary feature but It still encourages the action to go to your personal site

That's true but there is already a link on the page. I'd rather not spam people with it!

Issue #834, opened 2023-03-23
Extend an (unofficial) offer


joemasilotti:
I'd love for businesses to be able to extend unofficial offers to developers right inside of RailsDevs.

To me, an unofficial offer includes the start date, compensation package, title, and a freeform comment/notes field. This should be enough for the developer to a) get excited and b) decide if they want to move forward or start negotiating.

And by keeping this information inside of RailsDevs I will be able to help developers navigate those negotiations.

A PR addressing this issue should:

- [ ] Add a button to the conversation screen for businesses to "Extend an offer"
- [ ] Present a new screen with the above fields noting that the offer is unofficial and only used to gauge interest from the developer - but also that they should follow through if accepted
- [ ] When submitted, present a different conversation element (and new notification) with the offer details
- [ ] Add two buttons - Accept and Reject - for the developer to respond to the offer
- [ ] Present the response (and new notification) in the UI

I usually pull my designs from Tailwind UI. If you can't find something there or in the existing code then feel free to throw me unstyled HTML. And I'll take it from there!

rainbowRider5:
I could give it a try, It may be a longer PR tho.

It would be my first-ever open source assignment, If you  @joemasilotti are okay with that, I can start coding it.

joemasilotti:
Excellent! All yours.

Let me know if you have any questions or get stuck. If I don't hear from you in a week I'll reach out.

joemasilotti:
This is looking great so far! A recent PR was merged in that put a similar button on this UI. If you can move yours next to the "Schedule a call" button that would be great.

rainbowRider5:
@joemasilotti I'm wondering about the new conversation element part. 

Maybe for now we would always display offer as the last element in the conversation? We could create another issue for making the offers display between messages (in chronological order) by using some custom collection / polymorphic association?

rainbowRider5:
As for now I see many similarities to messages, so I think I'd refactor it into polymorphic association at some point.

joemasilotti:
Hmm. How much of the code would we have to rip out to refactor? I’m wondering if it makes sense to do now.

rainbowRider5:
I'd need to add some model between `Conversation` and `Message` (and the `Offer` model which I created), so `Conversation` would `has_many :messages, as: :conversable`. 

Then we could move all the things that are not specific to `Message`, but also apply to the `Offer` to `Conservable` (`views/conversable`, maybe `mailers/message_mailer` and maybe some other stuff).

[...]

PR #851, opened 2023-03-31
Extend offer feature

rainbowRider5:
This PR addresses #834.

This is still work in progress, the code needs a bit of refactoring.

Todo list:

- [x] Add a button to the conversation screen for businesses to "Extend an offer"
- [x] Present a new screen with the above fields noting that the offer is unofficial and only used to gauge interest from the developer - but also that they should follow through if accepted
- [x] When submitted, present a different conversation element (and new notification) with the offer details
- [x] Add two buttons - Accept and Reject - for the developer to respond to the offer
- [x] Present the response (and new notification) in the UI
 

Code quality checks


app/models/offer.rb
```diff
@@ -0,0 +1,35 @@
...
+  has_noticed_notifications
+
+  belongs_to :conversation, touch: true
+  has_one :developer, through: :conversation
+  has_one :business, through: :conversation
+
+  enum pay_rate_time_units: {hour: 0, day: 1, year: 2}
+  enum state: {proposed: 0, accepted: 1, declined: 2}
+
+  validates :conversation, uniqueness: {conditions: -> { active }, unless: :declined?}
```
rainbowRider5, line 14: Allow only one proposed/accepted offer per conversation

app/views/offers/new.html.erb
```diff
@@ -0,0 +1,14 @@
+<%= turbo_frame_tag "modal" do %>
+  <div class="fixed z-10 inset-0 overflow-y-auto" aria-labelledby="modal-title" role="dialog" aria-modal="true">
+    <div class="flex items-end justify-center min-h-screen pt-4 px-4 pb-20 text-center sm:block sm:p-0">
+      <div class="fixed inset-0 bg-gray-500 bg-opacity-75 transition-opacity" aria-hidden="true"></div>
+
+      <!-- This element is to trick the browser into centering the modal contents. -->
+      <span class="hidden sm:inline-block sm:align-middle sm:h-screen" aria-hidden="true">&#8203;</span>
+
+      <div class="inline-block align-bottom bg-white rounded-lg px-4 pt-5 pb-4 text-left overflow-hidden shadow-xl transform transition-all sm:my-8 sm:align-middle sm:max-w-lg sm:w-full sm:p-6">
```
rainbowRider5, line 9: We should consider extracting this "modal container" into some partial/component

app/views/conversations/show.html.erb
```diff
@@ -29,17 +29,24 @@
...
+      <% end %>
+    </div>
 
-  <div class="my-3 mx-4">
-    <%= turbo_frame_tag "message_read_indicator" do %>
-      <%= render Conversations::ReadIndicatorComponent.new(current_user, conversation: @conversation) %>
-    <% end %>
-  </div>
+    <%= render(@conversation.offers.last) if @conversation.offers.any? && Feature.enabled?(:offer_extending) %>
+  </ul>
```
rainbowRider5, line 39: For now I render only last offer at the end of the conversation.

rainbowRider5:
I tried to follow the convention in which messages were coded. There may be some places where we can extract some common logic of messages and offers. I did not use hotwire to refresh the conversation view, please let me know If we need to refactor the views into more hotwire and `ViewComponent`-way.

app/controllers/offers_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+  before_action :require_active_subscription!
+
+  def new
+    @offer = Offer.new(conversation: @conversation)
+  end
+
+  def create
+    user_not_authorized unless conversation.business?(current_user)
+
+    @offer = Offer.new(**offer_params, conversation_id: params[:conversation_id])
```
rainbowRider5, line 14: I had some troubles passing these params via `ActiveView` and path helpers

app/controllers/offers_controller.rb
```diff
@@ -0,0 +1,55 @@
...
+    redirect_to conversation_path(conversation)
+  end
+
+  private
+
+  def require_active_subscription!
+    if conversation.business?(current_user) && !current_user.permissions.active_subscription?
+      redirect_to pricing_path, alert: t("errors.business_subscription_inactive")
+    end
+  end
```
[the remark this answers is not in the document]
kimpastro: If you do so, you are violating the ISP principle, "I" from SOLID, which say:
`No class should be forced to implement any interface method(s) that it does not use.`

In that case a better approach is to think on that `require_active_subscription` as a "role", for example: `ActiveSubscriptionRequirer` (Read about Duck Typing in Ruby).

That role is the behavior shared between some specific classes, so just that role must be shared.

A possible solution is to create a concern and include that concern into those controllers that have this behavior in common.

app/controllers/offers_controller.rb
kimpastro: How so?

app/models/feature.rb
kimpastro, line 1: How about this refactoring?

```ruby
class Feature
  ENABLED_FEATURES = {
    cancel_subscription: true,
    business_welcome_email: true,
    redesign: ENV.fetch("REDESIGN", false),
    badge_filter: !Rails.env.production?,
    offer_extending: !Rails.env.production?,
    developer_specialties: !Rails.env.production?
  }.freeze

  def self.enabled?(feature_name)
    feature_sym = feature_name.to_sym

    raise "Unknown feature name: #{feature_name}" unless ENABLED_FEATURES.include?(feature_sym)

    ENABLED_FEATURES[feature_sym]
  end
end
```

app/views/offer_mailer/new_offer.html.erb
```diff
@@ -0,0 +1,8 @@
+<p><%= @sender %> extended you <%= link_to "an offer", notification_url(@notification.record) %>:</p>
+
+<blockquote>
+  <p>Start date: <%= @offer.start_date %></p>
+  <p>Pay rate: <%= @offer.pay_rate_value %>/<%= @offer.pay_rate_time_unit_key %></p>
+  <p>Comment:</p>
+  <p><%= @offer.comment? ? @offer.comment : 'No comment' %></p>
```
nunommc, line 7:
```suggestion
  <p><%= @offer.comment.presence || 'No comment' %></p>
```

I believe the `#comment?` doesn't add much, and that this could be on a decorator

app/views/offer_mailer/offer_declined.html.erb
```diff
@@ -0,0 +1,8 @@
+<p><%= @sender %> declined your <%= link_to "offer", notification_url(@notification.record) %>:</p>
+
+<blockquote>
+  <p>Start date: <%= @offer.start_date %></p>
+  <p>Pay rate: <%= @offer.pay_rate_value %>/<%= @offer.pay_rate_time_unit_key %></p>
```
nunommc, line 5: To prevent mistakes, it would be easier to define a `Offer#rate` method like

```ruby
def rate
  "#{@offer.pay_rate_value}/#{@offer.pay_rate_time_unit_key}"
end
```
as this is used in a few places

app/views/offers/_form.html.erb
```diff
@@ -0,0 +1,52 @@
...
+          <%= form.label :pay_rate_value, class: "block text-sm font-medium text-gray-700" %>
+          <div class="relative mt-1 rounded-md shadow-sm">
+            <div class="pointer-events-none absolute inset-y-0 left-0 flex items-center pl-3">
+              <span class="text-gray-500 sm:text-sm">$</span>
+            </div>
+            <%= form.text_field :pay_rate_value, type: "number", value: 50, required: true, class: "pl-7 pr-20 #{field_classes(form, :pay_rate_value)}" %>
+            <div class="absolute inset-y-0 right-0 flex items-center">
+              <%= form.select :pay_rate_time_unit,
+                              Offer.pay_rate_time_units.map { |unit, value| [unit, value] },
+                              {},
```
nunommc, line 30: Simpler to use https://apidock.com/rails/v5.2.3/ActionView/Helpers/FormOptionsHelper/options_from_collection_for_select ?
```suggestion
                              options_from_collection_for_select(Offer.pay_rate_time_units. :first, :last)
```

app/views/offers/_offer.html.erb
```diff
@@ -0,0 +1,47 @@
...
+        <% case offer.state %>
+        <% when 'proposed' %>
+            <%= t(".proposed_an_offer") %>
+        <% when 'accepted' %>
+            <%= t(".accepted_the_offer") %>
+            <%= inline_svg_tag "icons/solid/check.svg", class: "h-5 w-5 text-green-600", aria_hidden: true %>
+        <% when 'declined' %>
+            <%= t(".declined_the_offer") %>
+            <%= inline_svg_tag "icons/solid/x_circle.svg", class: "h-5 w-5 text-red-600", aria_hidden: true %>
+        <% end %>
```
nunommc, line 18: too much logic in the view, this could be in a helper file?

app/controllers/offers_controller.rb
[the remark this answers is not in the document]
rainbowRider5: I do not agree. ISP principle relates to the interface's methods implementation. If we put `require_active_subscription`  in the parent class we actually implement this method through inheritance. We do not force anything.
There are no interfaces in Ruby (at least not in the language per se).

I do agree using a module is a better approach tho.

app/controllers/offers_controller.rb
rainbowRider5: I expected `conversation_id` to be included in the `offer_params` out of the box as I used the `form_with model: [conversation, offer]` helper in `offers/_form.html.erb` view.

app/models/feature.rb
rainbowRider5: Yeah, I like this approach! 👌

app/controllers/offers_controller.rb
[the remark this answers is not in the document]
rainbowRider5: We use `require_active_subscription` in three different places. Only in two, the implementation is the same. I can see your concern with moving the method into `ApplicationController` now.

It would require a little bit more changes to refactor it than I expected.
Should we leave it, and delegate another issue for checking subscription around the controllers, or should I extract these two duplicate implementations into a concern now?

app/helpers/offers_helper.rb
```diff
@@ -0,0 +1,27 @@
+module OffersHelper
+  def offer_state_notice(offer_state)
+    case offer_state
+    when "proposed" then t(".proposed_an_offer")
```
nunommc, line 4: This is now failing the i18n-tasks, could you take a look please?

test/models/offer_test.rb
```diff
@@ -57,6 +47,10 @@ class OfferTest < ActiveSupport::TestCase
     assert_equal Offer.new(pay_rate_time_unit: 2).pay_rate_time_unit_key, "year"
   end
 
+  test "#rate returns properly formatted string of pay_rate_value and pay_rate_time_unit" do
+    assert_equal Offer.new(pay_rate_value: 50, pay_rate_time_unit: 2).rate, "50.0/year"
```
nunommc, line 51: One of the ideas behind enums is so that we don't reference values by ID.

```suggestion
    assert_equal Offer.new(pay_rate_value: 50, pay_rate_time_unit: :year).rate, "50.0/year"
```

nunommc (commented):
Looks a lot better, but the build is failing. 

Could you also update the changelog as it's showing this has been added in April?

app/helpers/offers_helper.rb
rainbowRider5: Yeah, sorry for that. Fixed;

nunommc (approved):
It looks OK to me, but it may need another pair of eyes more acquainted with the project.

app/controllers/offers_controller.rb
bschrag620: I think for that to occur it has to be included in the `form` as a value. `offer_params` is requiring the value to be nested inside of `offer`, so the only way for `offer_params` to pass the `conversation_id` would be for the params to look like `offer[conversation_id]=...`. Since `conversation_id` is being supplied by the url param instead it won't be permitted by the `offer_params` method.

bschrag620: If you add the conversation as a hidden input in the form that should do the trick - but that also feels like a bit of data coupling that might not be the best option.

joemasilotti:
Closed as not planned (right now). But will keep the branch alive in case I change my mind!

PR #852, opened 2023-04-01
Refactor DeveloperQuery to use Ransack

viktorvsk:
According to https://github.com/joemasilotti/railsdevs.com/issues/805 it is possible to make Developers search more declarative using [ransack gem](https://activerecord-hackery.github.io/ransack/) which builds SQL queries based on HTML inputs names.

Current implementation is the initial proof of concept how this could look like (Phase 1). Here we can utilize ransack, avoid using imperative query builders in `DeveloperQuery`, remove unused scopes and methods. However, to not make this PR huge and keep the main idea clear, HTML is not changed here, meaning we are mapping existing HTML inputs names to inputs names consumed by ransack in `DeveloperQueryRansackParamsBuilder`.

In Phase 2 we can get rid of `DeveloperQueryRansackParamsBuilder` by adjusting HTML inputs names, however, to make it 100% ransack-way we would need to slightly change database schema role `RoleType` and `RoleLevel` models (described in comments).

As an example of Phase 2 implementation input name for `utc_offsets` was changed to `location_utc_offset_in` to illustrate how final declarative approach looks like which led to changing unit tests and other places which depend on this query attribute.

P.S. `bin/check` complains on `initializers/ransack_models_configuration.rb` but it was added like this for "less invasive" Phase 1 approach (to not make many changes in different models)

## Pull request checklist


joemasilotti:
Thanks for the PR, @viktorvsk! This is super helpful in figuring out if Ransack is the right choice for RailsDevs.

Unfortunately, I don't think it makes sense to migrate right now.

Yes, Ransack removes a lot of boilerplate code in the query. I like that a lot! But it also requires a couple of abstractions that look, to me, to be more complicated in the long run.

I'm thinking that a better approach might be a few DSL-like methods that can add filters. Something that uses a tiny bit of metaprogramming to define a few of the common filter-types on the query.

I'm going to close this for now and rethink the best way to refactor in the future. Thanks again for diving in to this!

viktorvsk:
Sure, your welcome, @joemasilotti 

In scope of this app, I think its not boiler-plate is the main problem, but the amount of decisions to be made to add a new filter (or adjust existing) in future with multiple contributors. Also taking into account the business side of the project (1k-100k devs max, hundreds of skills, tens of other options) — there probably never become a moment to switch to raw SQL because of performance

That would make me assume there is no need to tailor (i.e. invest time) the searching feature and it'd better to abstract away what possible where ransack is the only solution

But that's just me :) Maybe for this project it would be better to leave things as it is if developers are used to existing approach for example

joemasilotti:
Great points all around! I commented the following when I closed the original issue. And I think that might be the root cause of lack of direction.

> Closing this as I need to rethink what I actually want, not just "make it better".

PR #855, opened 2023-04-04
Add `developers.bumped_at`

joemasilotti:
This PR lays some groundwork for a change to the sorting algorithm.

It adds two new columns: `developers.profile_updated_at` and `developers.bumped_at`.

The former will be used to see how recent the profile was updated to know when to send the profile reminder emails. This means we can stop dancing around "touching" the profile when making changes outside of user-initiated flow.

The second is how devs will be sorted by default moving forward. Any public change to a profile will refresh the column if it was last updated more than 1 month ago.

In summary, a developer can "bump" their profile to the top of results once per month!

## Pull request checklist

## To-do

- [ ] Ensure `developers.profile_updated_at` is being used for reminder emails
- [ ] Rip out any remaining `touch: false` calls or dances around touching records
- [ ] (New PR) Start sorting developers by the new column
- [ ] Announcement email and maybe a dedicated page to how sorting works

### DIFF
--- a/app/components/developers/badges_component.html.erb
+++ b/app/components/developers/badges_component.html.erb
@@ -5,9 +5,9 @@
     </div>
   <% end %>
 
-  <% if recently_active? %>
+  <% if recently_added? %>
     <div class="mr-4 mb-2">
-      <%= render BadgeComponent.new(t(".recently_active"), color: "green") %>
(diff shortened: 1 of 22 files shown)


joemasilotti:
Closed in favor of #856.

PR #857, opened 2023-04-05
Use postgresql@14 instead of deprecated postgresql

andrewculver:
Currently when you run `brew bundle install` per the installation instructions, you see a deprecation warning for PostgreSQL's package:

```
Warning: Use postgresql@14 instead of deprecated postgresql
Warning: Use postgresql@14 instead of deprecated postgresql
```

### DIFF
--- a/Brewfile
+++ b/Brewfile
@@ -1,6 +1,6 @@
 # Postgres
 brew "libpq" # M1 macs need libpq for pg gem
-brew "postgresql"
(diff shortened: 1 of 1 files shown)


joemasilotti:
Good to know - thanks for the heads up, Andrew.

Also, I just gotta know what you're doing cloning RailsDevs! 👀

PR #860, opened 2023-04-06
Add subscription renewal email

### LINKED ISSUE #767: Automate subscription renewal email
> ~Note: This issue is blocked by #764.~ This issue is no longer blocked!

I'd like to send an email to businesses a few days before their subscription renews. This is a great time to ask if they've hired anyone and could create some goodwill by letting them know before they accidentally renew.

- [ ] Send an email to subscribed users 3 days before their subscription renews

The subject and body of the email should be as follows, ensuring HTML + plain text versions:

Subject: Your RailsDevs subscription renews in 3 days. Have you made a hire?

Hey [first name],

Joe, founder of RailsDevs here. A quick reminder that your subscription will renew on [renewal date].

Already hired someone? Congratulations! [Fill out this form](/hire/forms/new) and I’ll send you info on next steps.

If you’re still looking then hit reply or [schedule a call](https://savvycal.com/joemasilotti/RailsDevs) and I’ll help find a new batch of candidates.

[signature]

jkostolansky:
This PR:
- Adds custom mailer for the Pay gem and overrides `subscription_renewing` email.
- Configures Pay to send renewal emails for monthly subscriptions.

Resolves #767


app/mailers/pay_mailer.rb
```diff
@@ -0,0 +1,14 @@
+class PayMailer < Pay::UserMailer
+  def subscription_renewing
+    @business = params[:pay_customer].owner.business
+    @renewal_date = params[:date].to_date
+
+    mail(
```
joemasilotti, line 6: Let's copy over (or ideally, reference) the defaults from `ApplicationMailer` here.

config/locales/en.yml
```diff
@@ -712,6 +712,9 @@ en:
         revenue: Revenue
         source: Source
         title: RailsDevs revenue
+  pay_mailer:
+    subscription_renewing:
+      subject: Your RailsDevs subscription renews in 3 days. Have you made a hire?
```
joemasilotti, line 717:
```suggestion
      subject: Your RailsDevs subscription renews soon. Have you made a hire?
```

Just in case they see it a day later.

test/fixtures/files/pay/invoice.upcoming.json
```diff
@@ -0,0 +1,120 @@
+{
```
joemasilotti, line 1: I love the idea of using the fixture but I'm not too happy about 120 lines of new code.

Can we trim this down to the bare minimum needed to get the test passing?

test/pay/subscription_renewing_mail_test.rb
```diff
@@ -0,0 +1,30 @@
+require "test_helper"
+
+class SubscriptionRenewingMailTest < ActiveSupport::TestCase
+  include ActionMailer::TestHelper
+
+  test "monthly subscription should receive correct renewal email" do
+    subscription = pay_subscriptions(:full_time)
+    event = new_upcoming_invoice_event(subscription)
+
+    assert_emails 1 do
```
joemasilotti, line 10: Let's use `assert_enqueued_email_with` so we can assert the correct mail and args.

Example on `test/integration/messages_test.rb:131`.

joemasilotti (changes requested):
This looks great! I made a few nitpicky questions/comments then this is ready for merge.

app/mailers/pay_mailer.rb
jkostolansky: I can use class-level `default` statements here, but that would probably override some settings in `pay.rb` for all pay emails, not just the `subscription_renewing` one. That's the reason why I've tried to avoid that.

Here are the original classes our custom mailer inherits from: [Pay::UserMailer](https://github.com/pay-rails/pay/blob/master/app/mailers/pay/user_mailer.rb), [Pay::ApplicationMailer](https://github.com/pay-rails/pay/blob/master/app/mailers/pay/application_mailer.rb).

So I'm just making sure - if you really want me to use the class-level `default` statements instead, let me know.

config/locales/en.yml
jkostolansky: 👍 done

test/fixtures/files/pay/invoice.upcoming.json
jkostolansky: We probably can, but is it a better solution? We're providing the Stripe's webhook data to Pay. If the data is not complete (as you're suggesting) and Pay changes something, for example starts processing another line from the data, the test could fail - and not because our code isn't working.

In my opinion, it's better to just copy-paste the same file Pay uses for testing the webhook, even if it's 120-line fixture. But we pretty much don't care about the content - except two lines specified in the test. And if Stripe changes something in the webhook data, we'll just need to copy-paste the whole file.

Again, if you really want to trim this down, let me know. 🙂

test/pay/subscription_renewing_mail_test.rb
jkostolansky: Changed 👍

test/fixtures/files/pay/invoice.upcoming.json
jkostolansky: This is probably the smallest footprint:
```ruby
Stripe::Event.construct_from(
  data: {
    object: {
      subscription: subscription.processor_id,
      next_payment_attempt: renewal_timestamp,
      lines: {
        data: [
          {
            price: {
              type: "recurring",
              recurring: {interval: "month"}
            }
          }
        ]
      }
    }
  }
)
```

joemasilotti: Good point - lets keep the entire file then!

joemasilotti (approved):

joemasilotti:
Looks great, thanks @jkostolansky!

PR #861, opened 2023-04-10
Fix template path for pay mails

jkostolansky:
Fixes a bug introduced in #860 - the new `PayMailer` inherits from `Pay::UserMailer` and overrides `subscription_renewing` email. However, Rails now looks for an email template for any other Pay email in `pay_mailer` path without a fallback to the original path `pay/user_mailer`, failing with `ActionView::MissingTemplate`.

### DIFF
--- a/test/mailers/previews/pay_mailer_preview.rb
+++ b/test/mailers/previews/pay_mailer_preview.rb
@@ -1,7 +1,9 @@
 class PayMailerPreview < ActionMailer::Preview
   def subscription_renewing
-    customer = Pay::Customer.first
-    renewal_date = 3.days.from_now
-    PayMailer.with(pay_customer: customer, date: renewal_date).subscription_renewing
+    PayMailer.with(
+      pay_customer: Pay::Customer.first,
+      pay_subscription: Pay::Subscription.first,
+      date: 3.days.from_now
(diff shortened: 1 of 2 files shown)


test/mailers/previews/pay_mailer_preview.rb
```diff
@@ -1,7 +1,9 @@
 class PayMailerPreview < ActionMailer::Preview
   def subscription_renewing
-    customer = Pay::Customer.first
-    renewal_date = 3.days.from_now
-    PayMailer.with(pay_customer: customer, date: renewal_date).subscription_renewing
+    PayMailer.with(
+      pay_customer: Pay::Customer.first,
+      pay_subscription: Pay::Subscription.first,
+      date: 3.days.from_now
+    ).subscription_renewing
```
jkostolansky, line 7: I've added the `pay_subscription` param here, because [Pay is providing it](https://github.com/pay-rails/pay/blob/59728b65763340daceb98e2c230894e8ad18d01f/test/mailers/previews/pay/user_mailer_preview.rb#L29-L35) to the mailer.

joemasilotti (approved):
Excellent, thanks @jkostolansky!

PR #867, opened 2023-05-12
Dismissible modal

omegahm:
I saw that the share modal wasn't dismissible by clicking the background or <kbd>esc</kbd>.

Screenshot 2023-05-12 at 15 39 51

This PR fixes that.

## Pull request checklist

- [ ] ~I added significant changes and product updates to the [changelog](CHANGELOG.md)~

### DIFF
--- a/app/javascript/controllers/modal_controller.js
+++ b/app/javascript/controllers/modal_controller.js
@@ -4,9 +4,27 @@ export default class extends Controller {
   static targets = ["modal"]
   static values = {toggleTimeout: {type: Number, default: 1000}}
 
+  closeOnEsc(event) {
+    if (event.key === "Escape") {
+      this.close()
+    }
+  }
+
+  connect() {
+    document.addEventListener("keydown", this.closeOnEsc)
+  }
+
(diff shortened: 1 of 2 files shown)


nunommc (approved):

joemasilotti (changes requested):
This doesn't seem to work in Safari or Brave.

We also need to add the `data-action` to the other modals on the site. I know there is one on the conversation "show" page for flagging the conversation.

Finally, I think it is a better practice to bind to the keyboard event in the DOM, directly. From [the docs](https://stimulus.hotwired.dev/reference/actions#keyboardevent-filter):

> There may be cases where [KeyboardEvent](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent) Actions should only call the Controller method when certain keystrokes are used.

> You can install an event listener that responds only to the Escape key by adding .esc to the event name of the action descriptor, as in the following example.

```html
<div data-controller="modal"
     data-action="keydown.esc->modal#close" tabindex="0">
</div>
```

PR #872, opened 2023-06-23
Refactor LinkedIn Profile Fetching Code

sarvaiyanidhi:
This PR reflects the changes required in code to insert/update record on each LinkedIn profile fetched from API as well as merging fetching and email digest into single task.

## Pull request checklist

### DIFF
--- a/app/models/developers/external_profiles/linkedin_profile_fetcher.rb
+++ b/app/models/developers/external_profiles/linkedin_profile_fetcher.rb
@@ -6,13 +6,16 @@ module Developers::ExternalProfiles
 
       developers.find_each.with_index do |developer, index|
         response = fetch_linkedin_profile(developer.linkedin)
-        record = external_profile(developer, response)
-        # Developers::ExternalProfile.upsert(record, unique_by: [:developer_id, :site]) if record.present?
+        external_profile = Developers::ExternalProfile.find_or_initialize_by(developer_id: developer.id, site: "linkedin")
+        record = external_profile(developer, response, external_profile)
         if record.present?
-          external_profile = Developers::ExternalProfile.find_or_initialize_by(developer_id: record[:developer_id], site: record[:site])
           external_profile.data = record[:data]
           external_profile.error = record[:error]
+          external_profile.fetched_at = Time.now
           external_profile.save!
+        else
+          external_profile.fetched_at = Time.now
+          external_profile.save(touch: false)
         end
 
         developers_left = developers.count - index - 1
@@ -20,11 +23,10 @@ module Developers::ExternalProfiles
       end
     end
 
-    def external_profile(developer, response)
+    def external_profile(developer, response, external_profile)
       if response.error?
         {developer_id: developer.id, site: "linkedin", data: {}, error: response.error}
       else
-        external_profile = Developers::ExternalProfile.linkedin_developer(developer)
         response_data = response.data.delete_if { |key, value| key == "logo_url" } unless response.data.blank?
         external_profile.data = response_data unless external_profile.blank?
         if external_profile.blank? || external_profile.data_changed?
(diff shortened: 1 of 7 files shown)


sarvaiyanidhi:
Hi @joemasilotti I have done the discussed changes in the code but I wanted to clarify on below points before proceeding any changes in the code for it.

- I see that in rake file we have added condition to execute task only on every 21st but I think we want to run it adhoc so wanted to confirm if I should remove this condition for now or you want to still keep it scheduled for 21st
https://github.com/joemasilotti/railsdevs.com/blob/e412e51a99f75919e2e85b321def60f971f6d56c/lib/tasks/linkedin.rake#L5

- I observed that there were changes in code wherein we are emailing only profiles updated in last month but I think it should be last 30 days as we are now combining both tasks.

https://github.com/joemasilotti/railsdevs.com/blob/main/app/models/email_digests/linkedin_profiles.rb#L4

I think we should change code to fetch profiles between 30.days.ago .. Time.now 

Thanks

app/models/developers/external_profiles/linkedin_profile_fetcher.rb
```diff
@@ -1,40 +1,32 @@
...
-      upsert_external_profiles(profiles)
     end
 
     def external_profile(developer, response)
       if response.error?
         {developer_id: developer.id, site: "linkedin", data: {}, error: response.error}
       else
         external_profile = Developers::ExternalProfile.linkedin_developer(developer)
-        external_profile.data = response.data unless external_profile.blank?
+        response_data = response.data.delete_if { |key, value| key == "logo_url" } unless response.data.blank?
```
sarvaiyanidhi, line 22: Removing logo_url from response as it is adding signature in value which is different in each fetch and creating issue when we compare data to update it or not.

joemasilotti: Good call!

app/models/developers/external_profiles/linkedin_profile_fetcher.rb
```diff
@@ -1,40 +1,32 @@
...
     def developer_profiles
-      profiles = []
       developers = developers_with_linkedin_profiles
       Rails.logger.info "Fetching #{developers.count} #{"LinkedIn profile".pluralize(developers.count)}..."
 
       developers.find_each.with_index do |developer, index|
         response = fetch_linkedin_profile(developer.linkedin)
         record = external_profile(developer, response)
-        profiles << record if record.present?
+        Developers::ExternalProfile.upsert(record, unique_by: [:developer_id, :site]) if record.present?
```
joemasilotti, line 10: Using upsert for a single record doesn't feel right to me. I think we want to instead use something like `find_or_initialize_by` or `find_or_create_by`. Then we can simplify this logic to find/initialize by the developer/site and update only if necessary.

joemasilotti:
> I observed that there were changes in code wherein we are emailing only profiles updated in last month but I think it should be last 30 days as we are now combining both tasks.

We have to make a decision before moving ahead here. Is this a task that is run on the 1st of every month? Or adhoc?

If on the first then we can change 21 to 1 and keep all the "last month" type logic. If adhoc then we need a way to know when an external profile was last fetched (not necessarily _updated_). And only query for their new data if 30 days has past.

I'd love to see the adhoc workflow working but it might add more complexity. What do you think?

app/models/developers/external_profiles/linkedin_profile_fetcher.rb
```diff
@@ -1,44 +1,46 @@
...
+        external_profile = Developers::ExternalProfile.find_or_initialize_by(developer_id: developer.id, site: "linkedin")
+        record = external_profile(developer, response, external_profile)
+        if record.present?
+          external_profile.data = record[:data]
+          external_profile.error = record[:error]
+          external_profile.fetched_at = Time.now
+          external_profile.save!
+        else
+          external_profile.fetched_at = Time.now
+          external_profile.save(touch: false)
```
sarvaiyanidhi, line 18: If data is fetched for developer after days from its last fetched_date but if there is no change in its API data in that case we are just updated fetched_at attribute to current date but not updating updated_at as we don't need this developer data in email sent for profiles updated in last 30 days.

sarvaiyanidhi:
Hi @joemasilotti, as per our discussion, I have implemented the necessary code changes to enable running the LinkedIn task on an ad hoc basis. Now, when we run the task, it will only fetch developers who haven't been updated in the last 30 days. This way, we can avoid making repetitive API calls for the same profiles. Additionally, the admin will be notified only about the profiles whose data has been updated in the last 30 days.

Please review the changes and let me know if you require any further modifications.

PR #876, opened 2023-07-14
Allow custom options to be passed to foreman start via bin/dev command

wuletawwonte:
This pull request enables contributors to pass custom options to the `foreman start` command via the `bin/dev` script. With this change, contributors can specify which processes to run from the `Procfile.dev` file, rather than starting all processes by default.

## Pull request checklist

Not applicable


bin/dev
```diff
@@ -3,5 +3,5 @@
 if command -v overmind &> /dev/null; then
   overmind start -f Procfile.dev
 else
-  foreman start -f Procfile.dev
+  foreman start -f Procfile.dev "$@"
```
heliocola, line 6: +1
I use this in another repo quite often, to run everything through foreman, and `rails s` separately.

In the other repo I do this:
One terminal: `rails s`
Another terminal: `bin/dev -c worker=1,js=1,css=1,redesign=1,stripe=1`

I've trying to think how we can simplify the other terminal.
It would be more pratically to do something like `bin/dev -!c web`
Maybe foreman has that option... but I have not had time to look yet! :-D

wuletawwonte: You can do something like this for the second terminal
```bash
bin/dev -m web=0,all=1
```
This will run all processes except `web` and the order of the arguments doesn't matter. There may be other ways but this is what I use.

joemasilotti:
Thanks for this! I just pushed a commit that also applies this to the overmind command, which seems to work the same way.

Once CI passes I'll merge this in.

PR #893, opened 2024-01-22
Updated Ruby to 3.2.3

heliocola:
Update Ruby to latest 3.2.x series

- `NA` [x] My code contains tests covering the code I modified
- `NA` [x] I linted and tested the project with `bin/check`
- `NA` [x] I added significant changes and product updates to the [changelog](CHANGELOG.md)
- [x] YAY: check if Heroku already supports it. Currently docs only show Ruby 3.2.2: https://devcenter.heroku.com/articles/ruby-support#supported-runtimes


heliocola:
Update on the last `YAY` item: it looks like the Heroku doc is not out of date and that Ruby 3.2.3 will be available today!

joemasilotti:
Excellent! Reply here when it is live and I'll merge this PR.

heliocola:
@joemasilotti : I just saw that Heroku update the docs and ruby 3.2.3 is available.
They have a track record of having ruby available really soon, so even if this is one of the first apps to run that in prod, I'd say the risks are really low...
And Heroku also have a smooth way to fail buils and rollback in case of issues, so... 🚀

joemasilotti:
Sweet! Merging now.

PR #925, opened 2025-05-08
Upgrade Rails to 7.1

PatrickMcSweeny:
This upgrades the app to Rails 7.1 as part of the ultimate goal of upgrading to 8.0. Rails 7.1 and 8.0 seem to not do well without upgrading the pay gem, so this is the farthest I could upgrade Rails on its own.
## Pull request checklist


test/components/developers/card_component_test.rb
```diff
@@ -58,7 +58,7 @@ class CardComponentTest < ViewComponent::TestCase
     test "doesn't render recently added badge if developer was added more than 7 days ago" do
       @developer.update!(created_at: 2.weeks.ago)
       render_inline(CardComponent.new(developer: @developer))
-      assert_no_text I18n.t("developers.card_component.recently_added")
+      assert_no_text I18n.t("developers.badges_component.recently_added")
```
PatrickMcSweeny, line 61: For some reason this only failed when I upgraded Rails even though it appears that it should have failed before.

PatrickMcSweeny:
I've looked at the failing tests, and I am honestly stumped. Neither of them seem to have anything to do with the upgrade.

When [this one](https://github.com/joemasilotti/railsdevs.com/blob/main/test/queries/stale_developers_query_test.rb#L29) fails, it returns a result set including the stale_developer with an updated_at value does not match the updated_at value of the stale_developer.

[This one](https://github.com/joemasilotti/railsdevs.com/blob/main/test/components/developers/open_graph_meta_tags_component_test.rb#L31) causes an error that reads "ArgumentError: Cannot get a signed_id for a new record" because the avatar file is not being persisted for some reason.

PatrickMcSweeny:
I ended up having to add the "abbrev" gem after the CI failed without it, which is weird because the tests ran locally without it. In any case, it's fewer gems that had to be added in Rails 7.0.

PR #928, opened 2025-05-09
Upgrade to Rails 7.2 and Ruby to 3.4

PatrickMcSweeny:
## Pull request checklist


PatrickMcSweeny:
The two tests that are failing are the same ones I mentioned in [this comment](https://github.com/joemasilotti/railsdevs.com/pull/925#issuecomment-2863922023) in the Rails 7.1 PR.

app/models/developer.rb
```diff
@@ -14,12 +14,7 @@ class Developer < ApplicationRecord
 
   FEATURE_LENGTH = 1.week
 
-  enum search_status: {
```
PatrickMcSweeny, line 17: I was getting warnings about keyword arguments for enums being deprecated in Rails 8.

Gemfile
```diff
@@ -37,6 +37,7 @@ group :test do
   gem "webmock"
 end
 
+gem "abbrev"
```
PatrickMcSweeny, line 40: CI wouldn't run without adding this gem

PatrickMcSweeny:
I confirmed manually that the image meta tag shows up successfully when on the developer page when an avatar is present.

app/models/developer.rb
joemasilotti: What's the formatting look like if you move each case to its own line? I wonder if that would drastically reduce this PR's diff.

PatrickMcSweeny: I tried that approach and it did reduce the diff.

PatrickMcSweeny:
I did some digging into the error caused by the avatar object not persisting, and it looks like the validation is failing because the file fixture doesn't have values for byte_size or checksum. I was not able to figure out to add those values in a way that the validation would pass.

PatrickMcSweeny:
@joemasilotti What are your plans for this?

